PR #58, opened 2019-11-12
Adding rfc-001

b00f:
Moved from Tendermint:
https://github.com/tendermint/tendermint/pull/3980


rfc/rfc_001_stamping-txs.md
```diff
@@ -0,0 +1,117 @@
...
+
+* 11-09-2019: Initial draft
+
+## Context
+
+Here I am going to explain how we can decouple transactions from the block by stamping them.
+Stamping a transaction simply means putting the previous block hash in the header of a transaction.
+A *stamped transaction* unlike normal transaction has life time and it's only valid for a certain time (TTL).
+If `N` is the number of blocks after the stamp (or block hash), the *stamped transaction* after N blocks will be invalid.
+By stamping transactions we don't need to save the transactions inside a block and only keeping the transaction hash is fair enough.
```
ValarDragon: This does not make sense to me.

Every syncing node requires the transaction data in order to check validity of state. (In fact they require this from genesis) These transactions must be kept around.

Once you've synced, you don't need to keep any of a past blocks data around, unless you want to help end-users or your fellow peers sync. Stamping transactions does not affect either of these.

If you want to help end users, query for transactions, then this seems like strictly more memory over the ux-node locally maintaining a map of tx hash -> (blocknum, tx data), and forgetting the rest of the block data.

rfc/rfc_001_stamping-txs.md
```diff
@@ -0,0 +1,117 @@
...
+ |                            |
+ +----------------------------+
+```
+
+### How to validate a stamped transaction
+
+A stamped transaction easily can be verified off-chain.
+First we need to find the block with corresponding stamp (or block hash).
+If it doesn’t exist it means transaction is not valid (or probably not belong to this fork).
+If the block is found, we need to iterate over `N` block after to check if transaction hash is inside a block. Otherwise transaction is not valid.
```
ValarDragon: I suggest rewording to:

If the block is found at height S, we need to iterate over the txs in blocks `[S, S + N]` to find the transaction with this hash. Otherwise the transaction is invalid.

rfc/rfc_001_stamping-txs.md
```diff
@@ -0,0 +1,117 @@
...
+
+  Stamping transaction can reduce the block size dramatically. For keeping a stamped transaction inside a block we need to store only 32 bytes hash of transaction instead of whole transaction data. In the bitcoin, the average transaction size has increased from 450 bytes in Jan 2013 to almost 600 bytes in Oct 2015. In the below chart you can see how stamped transaction could reduce the size of bitcoin blockchian if they could had applied from the beginning:
+
+  ![Bitcoin average block size](img/btc-block-size.png)
+  ![Bitcoin blockchain size](img/btc-blockchain-size.png)
+
+* Easy light clint implementation for PoS
+
+  Implementing a light client for PoW is easy since having block header is fair enough to check validity of a transaction using merkle tree.
+  However in PoS implementing a light client is a big pain. By stamping transaction light client implementation will be easy. A light client doesn't need to store any transaction data.
```
ValarDragon: How did this change the lite client in PoS? This doesn't affect the hard of a PoS liteclient, authenticating a block header. Once you have that, the state query and/or tx query is very easy, just like in PoW.

rfc/rfc_001_stamping-txs.md
```diff
@@ -0,0 +1,117 @@
...
+
+
+* Mitigating any possible forks.
+
+  *Nothing at stake attack* and *long range attack* are most fundamental attacks in PoS blockchains. There is no solution to overcome long range attack in PoS   blockchain. Stamping transactions can mitigate this issue by binding transactions to the main fork. A stamped transaction is always loyal to the main fork. Byzantine forks aren't able to put them in their fork because the stamp is not belongs to their fork.
```
ValarDragon: I think this is the biggest advantage of stamping transactions.

b00f: Yap. Actually I wanted to find a way for solving "Long Range attack", however it also reduce the size of block, and it's beautiful. In Tendermint block structure is a bit complex. I think `block_id` can be simply the block hash. No need to have block parts. It took time for me to understand what is block part and why we need it. I think that is avoidable.

rfc/rfc_001_stamping-txs.md
b00f: > In fact they require this from genesis

Not necessary. If we can have some check points and export state of blockchain on that point, I don't think validating all transactions are necessary. However nodes can start building blockchain from the genesis. Now bitcoin is a good experience for us. People tends to connect to a pool or even download the blockchain data rather than starting from genesis.

rfc/rfc_001_stamping-txs.md
b00f: Thanks. Please let me know if you have any suggestion on rewording the text.

rfc/rfc_001_stamping-txs.md
b00f: Good point. Maybe needs to change this part.
Maybe we don't need to have a lite client anymore. A lite client is simply a client without any TX-DB.

b00f:
@ValarDragon Thanks

rfc/rfc_001_stamping-txs.md
ValarDragon: I still think this is cheating for reducing size, similar to how segwit is really just cheating by moving the goal post and not actually solving anything. Using validity proofs over the entire chain (such as Coda) is an example of solving this without cheating.

marbar3778:
Sorry, we never got around to discussing this further. 

If you would like to still propose this could you amend it to RFC-003?

PR #68, opened 2019-12-03
Add spec doc about unconditional_peer, persistent_peers_max_dial of ADR-050

dongsam:
Add doc about  `unconditional_peer_ids`, `persistent_peers_max_dial_period` to spec
This spec was proposed [ADR-050](https://github.com/tendermint/tendermint/blob/master/docs/architecture/adr-050-improved-trusted-peering.md)

PR for doc has been separated from https://github.com/tendermint/tendermint/pull/4176


marbar3778:
thank you for making changes in this repo as well 😄

PR #71, opened 2020-01-16
light client: failure detection module

melekes:
https://github.com/tendermint/tendermint/issues/4215

[Rendered](https://github.com/tendermint/spec/blob/cbe20a644a6c37ec51ca217333e434fc04df8d66/spec/consensus/light-client/detection.md)


spec/consensus/light-client.md
```diff
@@ -316,3 +316,27 @@ func Backwards(h1,h2) bool {
...
 ```
+
+### Failure Detection Module
+
+The light client must be able to detect failures in the trust assumptions.
+Ideally, it should never trust a header, which was forged by malicious
+validators.
+
+To meet the above requirement, it **must be connected to at least one honest
+full node**.
```
josef-widder: I agree. This is necessary. In case of no fork on the main chain this is also sufficient. 
If there is a fork on the main chain, this might not be sufficient anymore: a fork on the main chain means that two full nodes have decided on two different headers for the same height. Then the requirement would at least require that the lite client is connected to one honest full node on each branch of the fork.

spec/consensus/light-client.md
```diff
@@ -316,3 +316,27 @@ func Backwards(h1,h2) bool {
...
+full nodes (called witness), which belong to different companies.
+
+After the light client verifies a new header (`H`), it should cross-check `H`
+with the headers from all witnesses. Cross-checking means comparing hashes of
+the headers. If any two hashes (or more) diverge, there's a fork.
+
+And since there is no way for the light client to detect who's lying to it
+(which full node - primary or one of the witnesses), it must form an evidence
+and send it to all connected full nodes. The evidence will typically contain
+a set of diverged headers (including commits for each of them).
```
josef-widder: I guess there might be cases where the lite client may detect who is lying, e.g., if basic verification fails (e.g., the header is signed by nodes that are not in the validator set).

spec/consensus/light-client.md
```diff
@@ -316,3 +316,27 @@ func Backwards(h1,h2) bool {
...
+
+After doing so, the light client must stop its operation. The operator will be
+forced to reset the light client with a new trusted header.
```
josef-widder: This is definitely a safe solution. But if the attack was only on the lite client, and there is no fork on the main chain, a lot of information that the lite client accumulated might be lost. Should we think about a way to stay operational in such scenarios, or is resetting acceptable?

spec/consensus/light-client.md
melekes: Not if we're running bisection (which is most common use-case).

spec/consensus/light-client.md
melekes: > Should we think about a way to stay operational in such scenarios, or is resetting acceptable?

resetting does not imply deleting the existing data. sorry for confusion

spec/consensus/light-client.md
```diff
@@ -316,3 +316,71 @@ func Backwards(h1,h2) bool {
...
+  height, the light client should reject the header it came from. If it came
+  from primary, it should select a new primary from the list of witnesses and
+  try to continue.
+
+3. Other attacks (Flip-flopping Amnesia & Back to the past, Lunatic)
+
+  Since there is no way for the light client to detect who's lying to it (which
+  full node - primary or one of the witnesses), it must form an evidence and
+  submit it to all connected full nodes (witnesses and primary). The evidence
+  will typically contain a set of diverged headers (including commits for each of
```
melekes: According to https://github.com/tendermint/tendermint/issues/4182, we should split the evidence into many small InvalidHeaderEvidence, but we don't know which header is invalid, so we'll need to create evidence for both valid and invalid header.

spec/consensus/light-client.md
```diff
@@ -316,3 +316,70 @@ func Backwards(h1,h2) bool {
...
+
+### Evidence detection
+
+After the light client verifies a new header (`H`) it received from primary, it
+should cross-check `H` with the headers from all witnesses. Cross-checking
+means comparing hashes of the headers. If any two hashes (or more) diverge,
+there's a fork (on the main chain OR phantom fork targeting this light client).
+
+The light client will then need to validate the header it got from a witness
+(`H1`) and verify it's signed by at least one trusted validator. If either of
```
melekes: Should we require +1/3 (i.e. CheckSupport returning true) here? Or even if one validator from currently trusted set signed, we should check & send evidence?

spec/consensus/light-client.md
```diff
@@ -316,3 +316,70 @@ func Backwards(h1,h2) bool {
...
+these fails, it should disconnect from the offending witness.
+
+1. Equivocation
+
+  If some validator double signed, the light client should form & submit a
+  `DuplicateVoteEvidence`.
+
+  After doing so, the light client must stop its operation. The operator will be
```
melekes: Should the light client continue to work say if H1 is not signed by +1/3 (no fork on the main chain)?

spec/consensus/light-client.md
```diff
@@ -316,3 +316,70 @@ func Backwards(h1,h2) bool {
...
+otherwise, they could create more forks.
+
+In practice, this means connecting to 1 or more geographically distributed full
+nodes (called witnesses), which belong to different companies. Note this number
+does not include the primary full node (called primary), which is used for
+obtaining new headers.
+
+_Remark_: we can't guarantee all witnesses won't follow the same branch of a
+fork. to guarantee that, we'll need the light client to be connected to +1/3 of
+nodes, which is impractical.
```
melekes: http://www.nooooooooooooooo.com/

spec/consensus/light-client.md
```diff
@@ -8,7 +8,7 @@ In order to make sure that full nodes have the incentive to follow the protocol,
 
 1. The lite client needs a method to verify headers it obtains from full nodes according to trust assumptions -- this document.
 
-2. The lite client must be able to connect to one correct full node to detect and report on failures in the trust assumptions (i.e., conflicting headers) -- a future document.
+2. The lite client must be able to connect to one correct full node to detect and report on failures in the trust assumptions (i.e., conflicting headers) -- this document.
```
josef-widder: I guess it would make sense to split the spec into two documents: (1) verification/bisection and (2) detector, as the functionality is quite different. (cp. https://github.com/interchainio/tendermint-rs/blob/e2cb9aca0b95430fca2eac154edddc9588038982/docs/architecture/adr-002-lite-client.md)

spec/consensus/light-client.md
```diff
@@ -316,3 +316,70 @@ func Backwards(h1,h2) bool {
...
+full node. This is sufficient so long as there's no fork on the main chain.
+
+If there is a fork on the main chain, it means that two full nodes have decided
+on two different headers for the same height. Then the requirement would at
+least require that the lite client is **connected to one honest full node on
+each branch of the fork**.
+
+_Assumption_: no more than `2f` of the nodes are malicious (`f+1` are honest).
+otherwise, they could create more forks.
+
```
josef-widder: I don't think we discussed that. But my first guess is that Byzantine faults could collude to create one branch for each correct validator. Even if per consensus they can only branch into two: they could fork into two at height h. Then the two validator sets for height h+1 might be disjoint, except of the faulty nodes. They might create more forks on each branch, etc.

spec/consensus/light-client.md
josef-widder: Hard guarantees will be super expensive (the lite client has to talk to all full nodes). We looked a bit into probabilistic guarantees, but this may need involved analysis, which we have not done yet. I guess for now the detector should just connect to some witnesses, and replace witnesses it observes problems with them.

spec/consensus/light-client.md
```diff
@@ -316,3 +316,70 @@ func Backwards(h1,h2) bool {
...
+1. Equivocation
+
+  If some validator double signed, the light client should form & submit a
+  `DuplicateVoteEvidence`.
+
+  After doing so, the light client must stop its operation. The operator will be
+  forced to reset the light client (resetting does not imply deleting the data
+  here) with a new trusted header.
+
```
josef-widder: I guess we need to be precise about what data can be maintained and how the detector interfaces with the bisection. If the lite client trusts a header h' and has used h during bisection to get this trust, and then later it turns out h is part of a fork, then the trust in h' needs to be revoked also. If h' has already been used, I guess we might have to reset in this case.

spec/consensus/light-client.md
josef-widder: I am not sure to understand the question: does this mean H1 obviously is a attack on the lite client only? I guess if the lite client detects that it may continue and look for the correct header of that height. ... but again, I am not sure I understood the question.

spec/consensus/light-client.md
josef-widder: I guess one is enough. But I will think some more about that...

spec/consensus/light-client.md
erikgrinaker: Expensive how? If latency; requests could be dispatched in parallel and thus the typical latency would be the 33rd percentile round-trip latency, which should be on the order of 100 ms. For something as important as this, isn't that a price worth paying?

erikgrinaker: Related: in order to verify that a sufficient share of the validator set that signed h1 also signed h2, will we not need to gather the public keys of the validators anyway?

spec/consensus/light-client.md
melekes: > does this mean H1 obviously is a attack on the lite client only?

Exactly

spec/consensus/light-client.md
melekes: > in order to verify that a sufficient share of the validator set that signed h1 also signed h2, will we not need to gather the public keys of the validators anyway?

we will need them, but these can be fetched from one node (primary)

melekes: > Expensive how?

in terms of load. suppose there are 300 vals on Cosmos network. 1/3 = 100. suppose there are 10_000 mobile wallets running a light client. this results in 10_000 requests to 100 nodes every 6 seconds.

erikgrinaker: > in terms of load

Since block headers are immutable I'm thinking they can probably be cached quite heavily on the server-side, so this would basically amount to in-memory hashmap lookups?

spec/consensus/light-client.md
melekes: > Even if per consensus they can only branch into two

they can only do that if they have `2f+1` (+2/3). otherwise, they can't change validator sets (`ValidatorsHash`), they can only vote for proposals from past rounds say (flip-flopping)

spec/consensus/light-client.md
melekes: we don't need headers per se. we only need their hashes to compare them.

spec/consensus/light-client.md
melekes: > the trust in h' needs to be revoked also

ugh... this won't make retail shops happy ("here's your coffee John. enjoy! wtf, wait! looks like your transaction only was committed in a fork")

spec/consensus/light-client.md
melekes: Sorry. The above statement is incorrect. +1/3 can create as many forks as it wants (by not voting for 2/3 proposals and waiting say 10 rounds, and later constructing a fork for each round)

josef-widder: I also discussed with Igor. We will use the model checker to see whether we can generate executions of consensus with several forks.

spec/consensus/light-client/detection.md
```diff
@@ -1,3 +1,68 @@
...
+
+If there is a fork on the main chain, it means that two full nodes have decided
+on two different headers for the same height. Then the requirement would at
+least require that the lite client is **connected to one honest full node on
+each branch of the fork**.
+
+_Remark_: +1/3 of malicious validators can create an unlimited number of forks.
+
+However, if we assume that correct full node is going to halt in case of a fork
+on the main chain, then we only need to notice that fact.
```
melekes: assumption: full nodes have more connections -> more chance of noticing a fork

melekes: TODO: expose an RPC endpoint on Tendermint?

spec/consensus/light-client/detection.md
```diff
@@ -1,3 +1,68 @@
...
+full node. This is sufficient so long as there's no fork on the main chain.
+
+If there is a fork on the main chain, it means that two full nodes have decided
+on two different headers for the same height. Then the requirement would at
+least require that the lite client is **connected to one honest full node on
+each branch of the fork**.
+
+_Remark_: +1/3 of malicious validators can create an unlimited number of forks.
+
+However, if we assume that correct full node is going to halt in case of a fork
```
melekes: this does not change anything for the light client because it still needs to be connected to the "right" set of nodes (i.e. on both branches) in order to detect a fork

spec/consensus/light-client.md
melekes: note 2: current TM full node has an upper bound on peers number - 50

a simple way for us to supply witnesses to light client can be accepting Tendermint's P2P address book (address_book.json) file and using it to fill witnesses.

spec/consensus/light-client/detection.md
```diff
@@ -1,3 +1,68 @@
-# Detection
+# Failure Detection
 
-TODO
+The light client must be able to detect failures in the trust assumptions.
+Ideally, it should never trust a header, which was forged by malicious
```
ebuchman: This is true if the trust assumption holds. If it fails, we may trust such a malicious header, but then the key is to detect it, submit evidence, and wait for a human.

spec/consensus/light-client/detection.md
```diff
@@ -1,3 +1,68 @@
...
+
+## Evidence detection
+
+After the light client verifies a new header (`H`) it received from primary, it
+should cross-check `H` with the headers from all witnesses. Cross-checking
+means comparing hashes of the headers. If any two hashes (or more) diverge,
+there's a fork (on the main chain OR phantom fork targeting this light client).
+
+The light client will then need to validate the header it got from a witness
+(`H1`) and verify it's signed by at least one trusted validator. If either of
```
ebuchman:
```suggestion
(`H1`) and verify it's signed by at least one correct trusted validator. If either of
```

spec/consensus/light-client.md
ebuchman: Oh, my thinking was that the light client would just send what it saw, and the full node's would handle splitting it into many small pieces since the full nodes will know which is invalid. I think technically the clients could figure it out too with a sequential sync but might as well let the full nodes do it

spec/consensus/light-client.md
ebuchman: How could one be enough? I thought for sure we'd need +1/3 here (ie. "at least one correct") ...

spec/consensus/light-client.md
ebuchman: Not sure we should bother with the light client detecting eg individual double signs if they're not part of a successful attack.

So I think there's really only two error modes here:
1) Faulty full node: mark them as bad and stop talking to them, but otherwise continue
2) Successful +1/3 attack: submit evidence and halt, wait for human intervention

Sounds like you're describing a case where they receive a conflicting header that does not fully verify but does contain a double sign from a validator. Since the commit doesn't fully verify, this is a faulty full node. Technically there is a faulty validator in here but I guess they would just go unpunished ... we could concern ourselves with trying to capture this down the road if we really want but I don't think it's a big issue if we don't.

spec/consensus/light-client.md
ebuchman: We can't guarantee connectivity to any fraction of nodes or validators. What we're interested in is essentially connectivity to a correct validator, but since they tend to be hidden behind full nodes ("sentries") we can't guarantee that either. So it's not even a matter of performance, the structure of the network just makes it impossible :)

spec/consensus/light-client.md
ebuchman: In any case if we detect a fork the light client needs to halt and be manually re-initialized.

spec/consensus/light-client.md
melekes: > Sounds like you're describing a case where they receive a conflicting header that does not fully verify but does contain a double sign from a validator

yep

> Technically there is a faulty validator in here but I guess they would just go unpunished 

Okay, I think we can skip this for now.

spec/consensus/light-client.md
melekes: https://github.com/tendermint/spec/pull/71#discussion_r374213456

melekes:
```
Therefore we define a notion of slashable fork which is informally defined as a fork in which bonded validators misbehaved (so they can be punished). To define this notion more precisely, we assume that there exists a main chain C (sequence of blocks), and assume that there is a set of bonded validators BD(t) that consists of all bonded (slashable) processes (current and old validators) at time t. Then a slashable fork at some time t corresponds to the existence of a commit for some block B at height h (B.time < t) that is different from a block at height h of the chain C; furthermore, a commit for the block B contains at least a single valid signature from a process that is part of the set BD(t). Note that detecting a slashable fork requires access to the main chain; therefore classifying if fork is slashable can be done only at the correct full node. On the other hand a lite client can only detect a fork, and then delegate the task of determining if the fork is slashable or not to some correct full node. 
```
from fork detection doc

spec/consensus/light-client/detection.md
```diff
@@ -1,3 +1,190 @@
...
+going to halt in case of a fork on the main chain, then **we only need to
+notice that fact**.
+
+After the light client verifies a new header (`H`) it received from primary, it
+should cross-check `H` with the headers from all witnesses. Cross-checking
+means comparing hashes of the headers. If any two hashes (or more) diverge,
+there's a fork (on the main chain OR phantom fork targeting this light client).
+
+The light client will then need to validate the header it got from a witness
+(`H1`) and verify the signers account for +1/3 of the voting power.
```
cmwaters, line 39: shall we add " and verify the signers account for +1/3 of the voting power of the validator set belonging to a header from the primary at the height of the root of the fork"? Is it important that we find the root of the fork in which we compare the validator set overlap?

melekes: I'm thinking this may be the best way to go where we don't have to send traces along w/ SignedHeader's

cmwaters: Okay, I'm thinking we have a reverse bisection function where we take both the secondary and primary and using bisection try to find the last header that they had in common and then submit the headers that come directly thereafter, in which there should be at least 1/3 overlap between both headers, as the conflicting headers - this way no matter where in the chain the light client recognises the fork all light clients will submit the same type of evidence - which happens at the root

PR #84, opened 2020-03-23
RFC-001: configurable block retention

erikgrinaker:
Rendered: https://github.com/erikgrinaker/spec/blob/erik/truncated-block-history/rfc/001-block-retention.md


erikgrinaker:
> I'd vote for adding explicit checks to Cosmos SDK to prevent misconfiguration.

I've added some enforced checks and made a few other changes. Will give others a chance to review as well.

erikgrinaker:
@alexanderbez @ebuchman @brapse Do you have any opinions on this?

rfc/001-truncated-block-history.md
```diff
@@ -0,0 +1,87 @@
...
+
+## Proposal
+
+* Unbonding time (SDK): already implemented as a genesis parameter `app_state.staking.params.unbonding_time`. Must now satisfy `unbonding_time < 2 * block.retention * block.time_iota_ms` otherwise errors (factor of 2 since `block.time_iota_ms` does not take into account block execution time).
+
+* Block time (Tendermint): already implemented as a genesis parameter `consensus_params.block.time_iota_ms`.
+
+* Block retention (Tendermint): implement as a new genesis parameter `consensus_params.block.retention` (default 0, i.e. retain all), and add a local configuration option `consensus.prune_blocks` (default off) that allows node operators to either enable or disable block pruning on a per-node basis. Must be at minimum 2, since the current and previous blocks are required for progress.
+
+* Snapshot interval (SDK): implement as a new genesis parameter `app_state.snapshots.interval` (default 0, i.e. disabled). Must satisfy `interval < 2 * block.retention` otherwise errors, to allow state synced nodes to catch up via block replay (factor of 2 to account for time spent taking a snapshot).
```
alexanderbez: Why is this a genesis parameter whereas the snapshot retention isn't? Also, all child keys under `app_state` are particular to modules, so `app_state.snapshots` is a misnomer. Can this be a local configuration just like snapshot retention?

rfc/001-truncated-block-history.md
```diff
@@ -0,0 +1,87 @@
...
+
+The main configuration options involved are:
+
+* Unbonding time (SDK): the duration in which validators can be economically punished for misbehavior.
+
+* Block time (Tendermint): the minimum duration between consecutive blocks.
+
+* Block retention (Tendermint): the number of recent blocks (heights) to retain.
+
+* Snapshot interval (SDK): the interval (in heights) between taking state sync snapshots.
```
alexanderbez: The nomenclature will get confusing for operators WRT to data backend pruning strategies. I'd just add a small note here to make a clear distinguishment between what snapshots are and what pruning snapshots are.

rfc/001-truncated-block-history.md
```diff
@@ -0,0 +1,87 @@
...
+
+* Snapshot retention (SDK): the number of recent state sync snapshots to retain.
+
+## Proposal
+
+* Unbonding time (SDK): already implemented as a genesis parameter `app_state.staking.params.unbonding_time`. Must now satisfy `unbonding_time < 2 * block.retention * block.time_iota_ms` otherwise errors (factor of 2 since `block.time_iota_ms` does not take into account block execution time).
+
+* Block time (Tendermint): already implemented as a genesis parameter `consensus_params.block.time_iota_ms`.
+
+* Block retention (Tendermint): implement as a new genesis parameter `consensus_params.block.retention` (default 0, i.e. retain all), and add a local configuration option `consensus.prune_blocks` (default off) that allows node operators to either enable or disable block pruning on a per-node basis. Must be at minimum 2, since the current and previous blocks are required for progress.
```
alexanderbez: Why does retention need to be a consensus parameter?

rfc/001-truncated-block-history.md
```diff
@@ -0,0 +1,87 @@
...
+
+* Genesis parameters for block retention, block time, and unbonding time allows governance consensus to ensure availability and correctness of light client verification and short-term chain auditability, with enforced sanity checks.
+
+* Genesis parameters for snapshot interval ensures uniform snapshot schedules across nodes, providing high availability of snapshots at a given height for parallel downloads and faster state syncing, with enforced sanity checks.
+
+* Node operators can independently decide whether they want to provide complete block histories and snapshots.
+
+### Negative
+
+* Social coordination is required to run archival nodes, failure to do so may lead to permanent loss of historical blocks.
```
alexanderbez: Note I think it's possible for the chain to actually halt permanently too.

rfc/001-truncated-block-history.md
```diff
@@ -0,0 +1,87 @@
...
+
+* Social coordination is required to run snapshot nodes, failure to do so may lead to inability to run state sync, and inability to bootstrap new nodes at all if no archival nodes are online.
+
+* Genesis parameter can only be changed by creating a new chain. This means that enabling block pruning or state sync, as well as modifying the settings, would be a non-trivial operation.
```
alexanderbez: See above -- why do any of them need to be consensus/genesis parameters?

rfc/001-truncated-block-history.md
```diff
@@ -0,0 +1,87 @@
...
+
+* Chain-wide block retention sets a lower bound on disk space requirements for all nodes.
```
alexanderbez: This is negative?

rfc/001-truncated-block-history.md
```diff
@@ -0,0 +1,87 @@
...
+
+* Chain-wide snapshot interval sets a lower bound on IO throughput requirements for all nodes (slower nodes will not be able to complete a snapshot before the next starts, in which case it will be skipped).
```
alexanderbez: ditto.

rfc/001-truncated-block-history.md
erikgrinaker: Snapshot retention is something that node operators may want to vary (e.g. if someone wants to archive all state snapshots for historical or auditing purposes). Making the interval a genesis parameter ensures all nodes take snapshots at the same heights, such that there is a higher availability of snapshots peers and thus faster state syncs for larger states.

I'm fine with using a local config option instead, that was my initial plan.

rfc/001-truncated-block-history.md
erikgrinaker: I share that concern. Any ideas for alternatives?

rfc/001-truncated-block-history.md
erikgrinaker: It initial wasn't, but @melekes and @marbar3778 were concerned that if this was up to individual node operators then the chance of misconfiguration was high. For example, most node operators would probably choose a low retention on their own, which could lead to no-one running archival nodes and thus loss of historical blocks and inability to replay history. The idea was that if it is a consensus parameter then it would force operators to have a deliberate approach to it.

Personally I'm fine with either option, we just have to pick something.

rfc/001-truncated-block-history.md
erikgrinaker: How so?

rfc/001-truncated-block-history.md
erikgrinaker: If it was not a chain-wide consensus parameter then node operators would be free to choose even lower retention. In principle, it would be possible to run a node which did not retain blocks at all, only state.

rfc/001-truncated-block-history.md
erikgrinaker: I've changed snapshot interval to be a local config option.

rfc/001-truncated-block-history.md
erikgrinaker: Another argument here is that if it's a consensus parameter then the SDK has access to it and can make sure the retention time is longer than the unbonding period. I don't believe this would currently be possible with a local Tendermint config option.

alexanderbez: With the former argument, I could argue that sane defaults and proper documentation should avoid situations of low-to-zero data availability.

The latter argument makes sense. But where is that enforced and checked? In the state machine? Then wouldn't that introduce a direct dependency on Tendermint? I suppose this is OK as we kind of already do this.

I really don't have a strong opinion here 👍

rfc/001-truncated-block-history.md
alexanderbez: When all nodes restart and no nodes have blocks available to replay to the current state. Correct me if I'm wrong.

rfc/001-truncated-block-history.md
```diff
@@ -0,0 +1,86 @@
...
+
+## Proposal
+
+* Block retention (Tendermint): implement as a new consensus parameter `block.retention` (default 0, i.e. retain all), and add a local config option `consensus.prune_blocks` (default off) that allows operators to enable or disable block pruning on a per-node basis. Must be at minimum 2, since the current and previous blocks are required for progress.
+
+* Block time (Tendermint): already implemented as a consensus parameter `block.time_iota_ms`.
+
+* Unbonding time (SDK): already implemented as a genesis parameter `app_state.staking.params.unbonding_time`. Must now satisfy `unbonding_time < 2 * block.retention * block.time_iota_ms` otherwise errors (factor of 2 since `block.time_iota_ms` does not take into account block execution time).
+
+* Snapshot interval (SDK): implement as a new config option `snapshot-interval` (default 0, i.e. disabled). Must satisfy `snapshot-interval < 2 * block.retention` otherwise errors, to allow state synced nodes to catch up via block replay (factor of 2 to account for time spent taking a snapshot).
```
alexanderbez: local config, correct?

rfc/001-truncated-block-history.md
erikgrinaker: > The latter argument makes sense. But where is that enforced and checked? In the state machine? Then wouldn't that introduce a direct dependency on Tendermint?

ABCI already provides consensus parameters to the application in `InitChain`, which would include the block retention - the SDK could then verify this against the unbonding time. After chain initialization, all consensus parameter changes must be done by the ABCI application itself.

rfc/001-truncated-block-history.md
```diff
@@ -0,0 +1,86 @@
...
+
+* Snapshot retention (SDK): the number of recent state sync snapshots to retain.
+
+## Proposal
+
+* Block retention (Tendermint): implement as a new consensus parameter `block.retention` (default 0, i.e. retain all), and add a local config option `consensus.prune_blocks` (default off) that allows operators to enable or disable block pruning on a per-node basis. Must be at minimum 2, since the current and previous blocks are required for progress.
+
+* Block time (Tendermint): already implemented as a consensus parameter `block.time_iota_ms`.
+
+* Unbonding time (SDK): already implemented as a genesis parameter `app_state.staking.params.unbonding_time`. Must now satisfy `unbonding_time < 2 * block.retention * block.time_iota_ms` otherwise errors (factor of 2 since `block.time_iota_ms` does not take into account block execution time).
```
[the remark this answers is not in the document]
erikgrinaker: Noted, thanks. This check would be implemented in the SDK.

rfc/001-truncated-block-history.md
erikgrinaker: Ah, this is something I wanted to look into. If it's the case that applications are not guaranteed to persist state on `Commit` then we need some signal from the app about what _is_ persisted, so that we don't prune blocks that haven't been persisted.

We don't currently seem to have this - on startup, Tendermint queries the last app height via `Query`, and then replays from there. However, at this point any transient data would be gone by definition (because it's after a restart). I think `Query` may return transient app heights during normal operation.

Personally I think `Commit` should be guaranteed to also persist data, but we could e.g. add `PersistentHeight` or `Persisted` in `CommitResponse` or something.

rfc/001-truncated-block-history.md
marbar3778: nit: maybe the `SDK` sections can be generalized for an application. We shouldn't have things that are needed specifically in the Cosmos-SDK in the Tendermint spec.

`An application building on Tendermint that would like to take advantage of these features should have x and y`

rfc/001-truncated-block-history.md
erikgrinaker: Correct, updated.

rfc/001-truncated-block-history.md
erikgrinaker: I agree, but the problems we're discussing here are specific to the SDK. None of these would apply to e.g. the `kvstore` app, because it doesn't have a concept of unbonding time or snapshots.

erikgrinaker: I'll try to generalize it a bit and/or move it to the SDK when we figure out what to do here, but I think it's good to keep the discussion as concrete as possible until then to avoid confusion.

rfc/001-truncated-block-history.md
erikgrinaker: This could actually be an alternative to the `block.retention` consensus parameter, as @marbar3778 suggested earlier. If the app returned `RetainBlocks` in `CommitResponse`, then it could take both persistence, unbonding time, and the snapshot interval into account and return an appropriate value.

alexanderbez: > Personally I think Commit should be guaranteed to also persist data

I also agree. The latest IAVL changes break this model of thinking for performance gains :-/

Using metadata via a Commit response could be promising. We should discuss on a call really the best approach forward.

rfc/001-block-retention.md
```diff
@@ -0,0 +1,111 @@
...
+
+Since `retain_height` is dynamic, it is possible for it to refer to a height which has already been removed. For example, commit at height 100 may return `retain_height=90` while commit at height 101 may return `retain_height=80`. This is allowed, and will be ignored - it is the application's responsibility to return appropriate values.
+
+## Open Questions
+
+* Should we add a local config option for Tendermint to retain blocks beyond ABCI `retain_height`? This is needed e.g. for archive nodes. For example, an option `retain_blocks=100` would cause Tendermint to retain at minimum the 100 most recent blocks, but more if ABCI commit `retain_height` requires it. Alternatively, this could be configured in the application.
```
melekes, line 61: I'd prefer one way of doing things

rfc/001-block-retention.md
```diff
@@ -0,0 +1,111 @@
...
+
+* Should state sync backfill historical blocks, such that all nodes have an application-controlled minimum block history available? This could be done e.g. by state sync snapshots including a `backfill_height` field which specifies the height to backfill from, and could use the existing fast sync reactor with minor modifications.
```
melekes, line 63: this generally seems like a good idea. Would we be able to backfill consensus params, validator sets, etc.. OR only blocks?

erikgrinaker: Good point. I think we should backfill everything, we will need the current values for state sync anyway since some of this is necessary for consensus progress. We can use RPC for this, but I’m not sure if that’s the optimal approach. We can also extend the blockchain reactor to exchange this.

Does this need to be in the first version of state sync, or can we add it later?

melekes: > Does this need to be in the first version of state sync, or can we add it later?

Don't see a reason why we need to do this now.

rfc/001-truncated-block-history.md
jackzampolin: maybe pruning checkpoints?

rfc/001-block-retention.md
erikgrinaker: I think that makes sense. This would also give the app total knowledge about block retention for correct nodes, which might be useful.

rfc/001-truncated-block-history.md
erikgrinaker: Yeah, I think that makes sense. I’ll open a separate issue/PR.

erikgrinaker: https://github.com/tendermint/iavl/issues/229

PR #85, opened 2020-04-02
add basic description of ABCI Commit.ResponseHeight

erikgrinaker:
Documentation for block pruning, once it's merged: https://github.com/tendermint/tendermint/pull/4588.

Minimum documentation, for now - we probably shouldn't encourage using this feature too much until we release state sync.


erikgrinaker:
> should RetainHeight be a uint64, will it ever be negative?

Yes, it should, but all the other height parameters in ABCI use int64 so I kept it for consistency. At some point we should move them all to uint64.

PR #90, opened 2020-04-27
document state sync ABCI interface and P2P protocol

erikgrinaker:
The corresponding Tendermint PRs are https://github.com/tendermint/tendermint/pull/4704 and https://github.com/tendermint/tendermint/pull/4705.

We should probably add a guide that reads more like a hands-on tutorial as well, with actual commands and examples. I'll write this as part of the release package.

### DIFF
--- a/spec/abci/abci.md
+++ b/spec/abci/abci.md
@@ -20,7 +20,7 @@ shared or included in a block.
 
 The info connection is for initialization and for queries from the user.
 
-The snapshot connection is for serving and restoring state sync snapshots.
+The snapshot connection is for serving and restoring [state sync snapshots](apps.md#state-sync).
 
 Additionally, there is a `Flush` method that is called on every connection,
 and an `Echo` method that is just for debugging.
@@ -173,7 +173,7 @@ local application with `ApplySnapshotChunk`. When all chunks have been applied, 
 `AppHash` is retrieved via an `Info` query and compared to the blockchain's `AppHash` verified via 
 light client. 
 
-For more information, see separate state sync section.
+For more information, see the [state sync section](apps.md#state-sync).
 
 ## Messages
 
@@ -445,8 +445,9 @@ For more information, see separate state sync section.
     `OfferSnapshot` calls.
   - Only `AppHash` can be trusted, as it has been verified by the light client. Any other data
     can be spoofed by adversaries, so applications should employ additional verification schemes
-    to avoid denial-of-service attacks.
-  - For more information, see `Snapshot` data type or the state sync section.
+    to avoid denial-of-service attacks. The verified `AppHash` is automatically checked against
+    the restored application at the end of snapshot restoration.
+  - For more information, see the `Snapshot` data type or the [state sync section](apps.md#state-sync).
 
 ### ApplySnapshotChunk
 
@@ -458,12 +459,12 @@ For more information, see separate state sync section.
   - `Result (Result)`: The result of applying this chunk.
     - `accept`: The chunk was accepted.
     - `abort`: Abort all snapshot restoration, including any other snapshots.
-    - `retry`: Reapply this chunk, combine with `RefetchChunks` and `RejectSender` as appropriate.
+    - `retry`: Reapply this chunk, combine with `RefetchChunks` and `RejectSenders` as appropriate.
     - `retry_snapshot`: Restart this snapshot from `OfferSnapshot`, reusing chunks unless 
       instructed otherwise.
     - `reject_snapshot`: Reject this snapshot, try a different one.
   - `RefetchChunks ([]uint32)`: Refetch and reapply the given chunks, regardless of `Result`. Only  
-    the listed chunks will be refetched, and applied in sequential order.
+    the listed chunks will be refetched, and reapplied in sequential order.
   - `RejectSenders ([]string)`: Reject the given P2P senders, regardless of `Result`. Any chunks 
     already applied will not be refetched unless explicitly requested, but queued chunks from these senders will be discarded, and new chunks or other snapshots rejected.
 - **Usage**:
@@ -477,8 +478,7 @@ For more information, see separate state sync section.
     network.
   - If Tendermint is unable to retrieve the next chunk after some time (e.g. because no suitable
     peers are available), it will reject the snapshot and try a different one via `OfferSnapshot`.
-    The application should be prepared to reset and accept this or respond with an abort as
-    appropriate.
+    The application should be prepared to reset and accept it or abort as appropriate.
 
 ### 
 
--- a/spec/abci/abci.md
+++ b/spec/abci/abci.md
@@ -430,15 +430,15 @@ via light client.
 - **Response**:
   - `Result (Result)`: The result of the snapshot offer.
     - `accept`: Snapshot is accepted, start applying chunks.
-    - `abort`: Abort all snapshot restoration, including other snapshots.
+    - `abort`: Abort snapshot restoration, and don't try any other snapshots.
     - `reject`: Reject this specific snapshot, try others.
     - `reject_format`: Reject all snapshots with this `format`, try others.
     - `reject_senders`: Reject all snapshots from all senders of this snapshot, try others.
 - **Usage**:
   - `OfferSnapshot` is called when bootstrapping a node using state sync. The application may
     accept or reject snapshots as appropriate. Upon accepting, Tendermint will retrieve and
-    apply snapshot chunks via `ApplySnapshotChunk`. The application may choose to reject a
-    snapshot while restoring it, in which case it should be prepared to accept further
+    apply snapshot chunks via `ApplySnapshotChunk`. The application may also choose to reject a
+    snapshot in the chunk response, in which case it should be prepared to accept further
     `OfferSnapshot` calls.
   - Only `AppHash` can be trusted, as it has been verified by the light client. Any other data
     can be spoofed by adversaries, so applications should employ additional verification schemes
@@ -455,7 +455,7 @@ via light client.
 - **Response**:
   - `Result (Result)`: The result of applying this chunk.
     - `accept`: The chunk was accepted.
-    - `abort`: Abort all snapshot restoration, including any other snapshots.
+    - `abort`: Abort snapshot restoration, and don't try any other snapshots.
     - `retry`: Reapply this chunk, combine with `RefetchChunks` and `RejectSenders` as appropriate.
     - `retry_snapshot`: Restart this snapshot from `OfferSnapshot`, reusing chunks unless 
       instructed otherwise.
--- a/spec/abci/abci.md
+++ b/spec/abci/abci.md
@@ -15,7 +15,7 @@ ABCI methods are split across four separate ABCI _connections_:
 The consensus connection is driven by a consensus protocol and is responsible
 for block execution.
 
-The mempool Connection is for validating new transactions, before they're
+The mempool connection is for validating new transactions, before they're
 shared or included in a block.
 
 The info connection is for initialization and for queries from the user.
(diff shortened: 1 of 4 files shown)


spec/abci/abci.md
```diff
@@ -528,3 +626,21 @@ Commit are included in the header of the next block.
...
+
+- **Fields**:
+  - `Height (uint64)`: The height at which the snapshot was taken (after commit).
+  - `Format (uint32)`: An application-specific snapshot format, allowing applications to version 
+    their snapshot data format and make backwards-incompatible changes. Tendermint does not 
+    interpret this.
+  - `Chunks (uint32)`: The number of chunks in the snapshot. Must be at least 1 (even if empty).
+  - `Hash (bytes)`: An arbitrary snapshot hash. Must be equal only for identical snapshots across 
+    nodes. Tendermint does not interpret the hash, it only compares them.
+  - `Metadata (bytes)`: Arbitrary application metadata, for example chunk hashes or other
```
melekes, line 640: Is metadata size limited?

spec/abci/apps.md
```diff
@@ -451,3 +457,145 @@ If `appBlockHeight == storeBlockHeight`
...
+
+Once all chunks have been accepted, Tendermint issues an `Info` ABCI call to retrieve the
+`LastBlockAppHash`. This is compared with the trusted app hash from the chain, retrieved and 
+verified using the light client. Tendermint also checks that `LastBlockHeight` corresponds to the
+height of the snapshot.
+
+This verification ensures that an application is valid before joining the network. However, the
+snapshot restoration may take a long time to complete, so applications may want to employ additional
+verification during the restore to detect failures early. This might e.g. include incremental
+verification of each chunk against the app hash (using bundled Merkle proofs), checksums to
```
melekes, line 581: should we checksum the spooled chunks while we store them on a disk?

spec/reactors/state_sync/reactor.md
```diff
@@ -0,0 +1,77 @@
...
+Here, `Missing` is used to signify that the chunk was not found on the peer, since an empty
+chunk is a valid (although unlikely) response. 
+
+The returned chunk is given to the ABCI application via `ApplySnapshotChunk` until the snapshot
+is restored. If a chunk response is not returned within some time, it will be re-requested,
+possibly from a different peer.
+
+The ABCI application is able to request peer bans and chunk refetching as part of the ABCI protocol.
+
+If no state sync is in progress (i.e. during normal operation), any unsolicited response messages
```
melekes, line 76: should we disconnect/ban peers for sending us unsolicited messages?

spec/abci/abci.md
erikgrinaker: No, but when sent across the network the entire message is limited to 4 MB. Will add a note.

spec/abci/apps.md
erikgrinaker: I don't think so. It might make sense to use an end-to-end checksum, but we've decided to defer all verification to the application - it can implement this if it considers it important. The data is more likely to be corrupted at the source or across the network.

spec/reactors/state_sync/reactor.md
erikgrinaker: Let's see if it becomes a problem. This could be caused by race conditions as well, e.g. a delayed message reaching us after the state sync has completed.

spec/abci/abci.md
```diff
@@ -381,6 +402,83 @@ Commit are included in the header of the next block.
...
+  - `AppHash ([]byte)`: The light client-verified app hash for this height, from the blockchain.
+- **Response**:
+  - `Result (Result)`: The result of the snapshot offer.
+    - `accept`: Snapshot is accepted, start applying chunks.
+    - `abort`: Abort all snapshot restoration, including other snapshots.
+    - `reject`: Reject this specific snapshot, try others.
+    - `reject_format`: Reject all snapshots with this `format`, try others.
+    - `reject_senders`: Reject all snapshots from all senders of this snapshot, try others.
+- **Usage**:
+  - `OfferSnapshot` is called when bootstrapping a node using state sync. The application may
```
alexanderbez, line 438: How can we return `reject|abort` in the response to `OfferSnapshot` once we start accepting and receiving chunks via `ApplySnapshotChunk`?

spec/abci/abci.md
```diff
@@ -381,6 +402,83 @@ Commit are included in the header of the next block.
...
+### ApplySnapshotChunk
+
+- **Request**:
+  - `Index (uint32)`: The chunk index, starting from `0`. Tendermint applies chunks sequentially.
+  - `Chunk ([]byte)`: The binary chunk contents, as returned by `LoadSnapshotChunk`.
+  - `Sender (string)`: The P2P ID of the node who sent this chunk.
+- **Response**:
+  - `Result (Result)`: The result of applying this chunk.
+    - `accept`: The chunk was accepted.
+    - `abort`: Abort all snapshot restoration, including any other snapshots.
```
alexanderbez, line 458: Multiple snapshot restorations can occur in parallel?

spec/abci/abci.md
```diff
@@ -381,6 +402,83 @@ Commit are included in the header of the next block.
...
+    - `retry`: Reapply this chunk, combine with `RefetchChunks` and `RejectSenders` as appropriate.
+    - `retry_snapshot`: Restart this snapshot from `OfferSnapshot`, reusing chunks unless 
+      instructed otherwise.
+    - `reject_snapshot`: Reject this snapshot, try a different one.
+  - `RefetchChunks ([]uint32)`: Refetch and reapply the given chunks, regardless of `Result`. Only
```
alexanderbez, line 463: When and why would this happen? Some IO error?

spec/abci/abci.md
```diff
@@ -381,6 +402,83 @@ Commit are included in the header of the next block.
...
+- **Usage**:
+  - The application can choose to refetch chunks and/or ban P2P peers as appropriate. Tendermint
+    will not do this unless instructed by the application.
+  - The application may want to verify each chunk, e.g. by attaching chunk hashes in
+    `Snapshot.Metadata` and/or incrementally verifying contents against `AppHash`.
+  - When all chunks have been accepted, Tendermint will make an ABCI `Info` call to verify that
+    `LastBlockAppHash` and `LastBlockHeight` matches the expected values, and record the
+    `AppVersion` in the node state. It then switches to fast sync or consensus and joins the 
+    network.
+  - If Tendermint is unable to retrieve the next chunk after some time (e.g. because no suitable
```
alexanderbez, line 476: This means the application won't actually persist items to disk/commit them until the entire snapshot is applied and verified?

spec/abci/apps.md
```diff
@@ -451,3 +457,146 @@ If `appBlockHeight == storeBlockHeight`
...
+
+For details on the specific ABCI calls and types, see the [methods and types section](abci.md).
+
+### Taking Snapshots
+
+Applications that want to support state syncing must take state snapshots at regular intervals. How
+this is accomplished is entirely up to the application. A snapshot consists of some metadata and
+a set of binary chunks in an arbitrary format:
+
+* `Height (uint64)`: The height at which the snapshot is taken. It must be taken after the given
```
alexanderbez, line 482: Why before the next height is processed?

spec/abci/apps.md
```diff
@@ -451,3 +457,146 @@ If `appBlockHeight == storeBlockHeight`
...
+* `Hash ([]byte)`: An arbitrary hash of the snapshot. This is used to check whether a snapshot is
+  the same across nodes when downloading chunks.
+
+* `Metadata ([]byte)`: Arbitrary snapshot metadata, e.g. chunk hashes for verification or any other
+  necessary info.
+
+For a snapshot to be considered the same across nodes, all of these fields must be identical. When
+sent across the network, snapshot metadata messages are limited to 4 MB.
+
+When a new node is running state sync and discovering snapshots, Tendermint will query an existing
```
alexanderbez, line 501: Applications will always favor the latest snapshot, correct?

spec/abci/apps.md
```diff
@@ -451,3 +457,146 @@ If `appBlockHeight == storeBlockHeight`
...
+
+* **Asynchronous:** Taking a snapshot can be time-consuming, so it should not halt chain progress,
+  for example by running in a separate thread.
+
+* **Deterministic:** A snapshot taken at the same height in the same format should be identical
+  (at the byte level) across nodes, including all metadata. This ensures good availability of 
+  chunks, and that they fit together across nodes.
+  
+A very basic approach might be to use a datastore with MVCC transactions (such as RocksDB), start a
+transaction immediately after block commit, and spawn a new thread which is passed the transaction
```
alexanderbez, line 518: transaction handle?

spec/abci/apps.md
```diff
@@ -451,3 +457,146 @@ If `appBlockHeight == storeBlockHeight`
...
+
+Old snapshots should be removed after some time - generally only the last two snapshots are needed
+(to prevent the last one from being removed while a node is restoring it).
+
+### Bootstrapping a Node
+
+An empty node can be state synced by setting the configuration option `statesync.enabled =
+true`. The node also needs the chain genesis file for basic chain info, and configuration for
+light client verification of the restored snapshot: a set of Tendermint RPC servers, and a
+trusted header hash and corresponding height from a trusted source, via the `statesync`
```
alexanderbez, line 533: which header and height? Does it matter?

spec/abci/apps.md
```diff
@@ -451,3 +457,146 @@ If `appBlockHeight == storeBlockHeight`
...
+suitable snapshot (generally prioritized by height, format, and number of peers), and offers it
+to the application via `OfferSnapshot`. The application can choose a number of responses,
+including accepting or rejecting it, rejecting the offered format, rejecting the peer who sent
+it, and so on. Tendermint will keep discovering and offering snapshots until one is accepted or
+the application aborts.
+
+#### Snapshot Restoration
+
+Once a snapshot has been accepted via `OfferSnapshot`, Tendermint begins downloading chunks from
+any peers that have the same snapshot (i.e. that have identical metadata fields). Chunks are
```
alexanderbez, line 555: Shouldn't it just be `i.e. that have the same identical snapshot hash`?

spec/abci/apps.md
erikgrinaker: Different database libraries use different terminology for this. It is analogous to a file handle, e.g.:

```go
app.Commit()
txn := db.Begin()
go snapshot(txn)
```

Where `txn` is the transaction handle. The important part is that creating the transaction must complete before the application proceeds to make any changes to the state machine; if `db.Begin()` is called inside the goroutine there is a race condition.

spec/abci/apps.md
erikgrinaker: Only that it is some header and height that the light client considers trustworthy in relation to the snapshot height. E.g., that the validator set hasn't changed too much in the period between the height and the snapshot height that we can no longer trust the application hash. I'm a bit fuzzy on the details here, @melekes knows more.

spec/abci/apps.md
erikgrinaker: No, all fields must be the same - if node A and B present the same snapshot hash, but A says 7 chunks and B says 9, it is clearly not the same snapshot. Similarly with `Metadata`, if this contains e.g. chunk hashes and they are different even though `Hash` is the same, then they are clearly not the same snapshot.

spec/abci/apps.md
alexanderbez: Hmm, I still don't quite follow this. Before height 101 is processed by the state-machine? When would height H ever contain data from H+1? How would that even happen?

erikgrinaker: It depends on how the application implements versioning, if at all. Consider a simple counter application that does something like this:

```go
counter += 1
go saveToFile(counter)
counter += 1
```

`saveToFile` may be called after the second `counter += 1`, in which case the snapshot would contain data from the next height. In this example the application should clearly not use a goroutine at all, but it's really up to the application how to implement this.

alexanderbez: I see. Well yes, that makes sense and I guess it should be noted. However, I don't see how or why this would ever happen in a deterministic state-machine (i.e. blockchain). But it is an important note nevertheless!

I would note though, that we rephrase this a bit to be a bit more of like your response. i.e. instead of saying before the next height, state that the snapshot height cannot data from the next height.

erikgrinaker: We explicitly recommend users to take snapshots asynchronously, which complicates this. If the snapshot was taken synchronously (without a goroutine) then you're right, this would not happen - but then the chain would halt until the snapshot was completed, which can take a long time.

erikgrinaker: Made some tweaks, let me know if anything is still unclear

https://github.com/tendermint/spec/pull/90/commits/2f4fc76477951a9a53fbf64d3a3c08ac02c2cbdc

ianharrd69:
Mine

PR #93, opened 2020-05-05
restore "document state sync ABCI interface and P2P protocol (#90)""

tessr:
This PR restores @erikgrinaker's work from #90. From the original PR description:

```
The corresponding Tendermint PRs are tendermint/tendermint#4704 and tendermint/tendermint#4705.

We should probably add a guide that reads more like a hands-on tutorial as well, with actual commands and examples. I'll write this as part of the release package.
```


tessr:
My apologies for the git rodeo, everyone.

PR #94, opened 2020-05-08
reactors/pex: specify hash function

melekes:
https://github.com/tendermint/tendermint/pull/4810/files


melekes:
> Perhaps elaborate with a bit more detail?

you mean how exactly keys are calculated?

erikgrinaker:
> > Perhaps elaborate with a bit more detail?
> 
> you mean how exactly keys are calculated?

No, it's just a bit unclear from the context what the purpose of the hashing function is. The documentation says:

> Peers are put in buckets according to their IP groups.

So is the hashing function applied to the IP groups of the peer to figure out which bucket to place them in? What's an IP group?

melekes:
I will clarify, thanks!

melekes:
Done

PR #95, opened 2020-05-11
Update evidence params with MaxNum

cmwaters:
As part of [#2590](https://github.com/tendermint/tendermint/issues/2590)

Update usages of the evidence params to include the max number parameter which sets the maximum number of evidence that can be committed to a single block


spec/blockchain/state.md
```diff
@@ -113,6 +113,7 @@ type BlockParams struct {
 type EvidenceParams struct {
 	MaxAgeNumBlocks int64
 	MaxAgeDuration  time.Duration
+    MaxNum          uint32
```
melekes, line 116: formatting is off

PR #100, opened 2020-06-15
spec: link to abci server implementations

xla:
Adds links to both the Golang and Rust in-process libraries.


marbar3778:
@xla is the rust library really able to run in process?

PR #101, opened 2020-06-16
update to spec for proto changes

### LINKED ISSUE #107: spec/blockchain/encoding.md is out of date

marbar3778:
This Pr aims to update the spec in regards to the proto changes and other various updates that havent been made

closes #107


spec/abci/abci.md
```diff
@@ -238,7 +238,7 @@ via light client.
   - `ChainID (string)`: ID of the blockchain.
   - `ConsensusParams (ConsensusParams)`: Initial consensus-critical parameters.
   - `Validators ([]ValidatorUpdate)`: Initial genesis validators, sorted by voting power.
-  - `AppStateBytes ([]byte)`: Serialized initial application state. Amino-encoded JSON bytes.
+  - `AppStateBytes ([]byte)`: Serialized initial application state. Proto-encoded JSON bytes.
```
marbar3778, line 241: Not proto encoded, we should add a minimal spec for the supported json. 

We dont assume the json encoding, just `json.rawmessage`

spec/blockchain/encoding.md
```diff
@@ -314,13 +265,11 @@ Because Tendermint only uses a Simple Merkle Tree, application developers are ex
 
 ## JSON
```
marbar3778, line 266: do we want to pull this out into its own page where we can better define the spec of the json encodeer?

marbar3778:
Marking this as R4R to start getting some reviews and answer the question about json encoding

spec/reactors/mempool/messages.md
```diff
@@ -4,22 +4,14 @@
 
 There is currently only one message that Mempool broadcasts
 and receives over the p2p gossip network (via the reactor):
-`TxMessage`
 
-```go
-// TxMessage is a MempoolMessage containing a transaction.
-type TxMessage struct {
-    Tx types.Tx
+```protobuf
```
marbar3778, line 8: note to self: this is now a oneof

spec/blockchain/encoding.md
erikgrinaker: Probably. Note that the JSON encoding is not consistent: some components use the native Golang encoder, others use the Amino/Protobuf-style encoder. At some point we should just pick one and stick to it.

spec/abci/abci.md
```diff
@@ -303,7 +303,7 @@ via light client.
   - `ByzantineValidators ([]Evidence)`: List of evidence of
     validators that acted maliciously.
 - **Response**:
-  - `Tags ([]kv.Pair)`: Key-Value tags for filtering and indexing
+  - `Events ([]abci.Event)`: Key and multiple events for filtering.
```
ebuchman, line 306: Now that we have an Event type we should explicitly document it here. Also need to add some docs for the new `.index` field. Should be made clear that this is a non-deterministic subjective field that can be different on different nodes.

PR #108, opened 2020-06-19
spec: update evidence in blockchain.md

cmwaters:
to reflect the changes in tendermint repo


marbar3778:
do you think we should add conflicting headers?

cmwaters:
conflicting headers isn't an evidence that can be committed on the chain and is an intermediate step for all the other evidence so I decided to not include it (same reason why potential amnesia evidence is not there). Do you agree with this line of thinking?

marbar3778:
sounds good to me. thank you for this!!

PR #118, opened 2020-07-22
spec/abci: expand on Validator#Address

melekes:
Refs https://github.com/tendermint/tendermint/issues/3732

### DIFF
--- a/spec/abci/abci.md
+++ b/spec/abci/abci.md
@@ -532,7 +532,7 @@ via light client.
 ### Validator
 
 - **Fields**:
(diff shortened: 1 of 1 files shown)


spec/abci/abci.md
```diff
@@ -532,7 +532,7 @@ via light client.
 ### Validator
 
 - **Fields**:
-  - `Address ([]byte)`: Address of the validator (hash of the public key)
+  - `Address ([]byte)`: Address of the validator (the first 20 bytes of SHA256(public key))
```
ebuchman, line 535: This is only true for ed25519 keys, but not necessarily in general ...

PR #119, opened 2020-07-26
RFC-002: non-zero genesis

erikgrinaker:
[Rendered](https://github.com/tendermint/spec/blob/erik/rfc-nonzero-genesis/rfc/002-nonzero-genesis.md)

Initial draft to discuss non-zero genesis. Primary question to answer: should genesis include a link to the previous chain (e.g. block header), and should this link be verified and tied into the initial block of the new chain or only be advisory for users and integrators?

### DIFF
--- a/rfc/002-nonzero-genesis.md
+++ b/rfc/002-nonzero-genesis.md
@@ -5,6 +5,7 @@
 - 2020-07-26: Initial draft (@erikgrinaker)
 - 2020-07-28: Use weak chain linking, i.e. `predecessor` field (@erikgrinaker)
 - 2020-07-31: Drop chain linking (@erikgrinaker)
+- 2020-08-03: Add `State.InitialHeight` (@erikgrinaker)
 
 ## Author(s)
 
--- a/rfc/002-nonzero-genesis.md
+++ b/rfc/002-nonzero-genesis.md
@@ -1,0 +1,104 @@
+# RFC 002: Non-Zero Genesis
+
+## Changelog
+
+- 2020-07-26: Initial draft (@erikgrinaker)
+
+## Author(s)
+
+- Erik Grinaker (@erikgrinaker)
+
+## Context
+
+The recommended upgrade path for block protocol-breaking upgrades is currently to hard fork the
+chain (see e.g. [`cosmoshub-3` upgrade](https://blog.cosmos.network/cosmos-hub-3-upgrade-announcement-39c9da941aee)).
+This is done by halting all validators at a predetermined height, exporting the application
+state via application-specific tooling, and creating an entirely new chain using the exported
+application state.
+
+As far as Tendermint is concerned, the upgraded chain is a completely separate chain, with e.g.
+a new chain ID and genesis file. Notably, the new chain starts at height 1, and has none of the
+old chain's block history. This causes problems for integrators, e.g. coin exchanges and
+wallets, that assume a monotonically increasing height for a given blockchain. Users also find
+it confusing that a given height can now refer to distinct states depending on the chain
+version.
+
+An ideal solution would be to always retain block backwards compatibility in such a way that chain 
+history is never lost on upgrades. However, this may require a significant amount of engineering
+work that is not viable for the planned Stargate release (Tendermint 0.34), and may prove too
+restrictive for future development.
+
+As a first step, allowing the new chain to start from an initial height specified in the genesis
+file would at least provide monotonically increasing heights. It may also be useful to include the 
+latest block header from the previous chain, to link them.
+
+External tooling would be required to map historical heights onto e.g. archive nodes that contain 
+blocks from previous chain version. Tendermint will not include any such functionality.
+
+## Proposal
+
+There are two main approaches that can be taken:
+
+* **Initial height:** the genesis file specifies an arbitrary initial height that the chain starts 
+from.
+
+* **Chain linking:** the genesis file includes e.g. the last block header and chain ID from the
+previous chain, and possibly uses this as the basis for the initial block.
+
+ABCI applications will have to be updated to handle arbitrary initial heights.
+
+### Initial Height
+
+This approach allows a chain to start from an arbitrary height, with the least amount of changes
+to existing code and protocols. Social consensus is necessary to link it to the previous chain.
+
+* A new field `initial_height` is added to the genesis file, defaulting to `1`. It can be set to any
+non-negative integer, and `0` is considered equivalent to `1`.
+
+* A new field `InitialHeight` is added to the ABCI `RequestInitChain` message, with the same value 
+and semantics as the genesis field.
+
+If possible, no further changes will be made to any data structures. In particular, the initial
+height will not be added to the node state, and will instead be passed explicitly to any logic that 
+relies on it.
+
+### Chain Linking
+
+This approach includes an explicit link to the previous chain in the genesis file. This can take 
+several forms, e.g.:
+
+* Weak: an advisory chain ID and last block header or hash. This will not be used by Tendermint at 
+all, and is only there to guide users and integrators to the previous chain and block.
+
+* Strong: the chain ID, last block header, and last commit. This will be verified by Tendermint 
+(requiring support for the previous chain's data structures and hashing algorithms) when processing 
+the initial height on the new chain, which e.g. ensures that the same validator set is used.
+
+In these cases, the initial height can either be derived from the last block header, or combined
+with a separate `initial_height` field.
+
+Strong linking will require additional spec and engineering work that may delay Stargate, but gives
+stronger guarantees of chain continuity, which may be particularly important to IBC and light
+clients (where validator set continuity is essential to punish misbehavior on the old chain).
+
+## Status
+
+Proposed
+
+## Consequences
+
+### Positive
+
+* Heights will be unique throughout the history of a "logical" chain, across hard fork upgrades.
+
+### Negative
+
+* Upgrades still cause loss of block history.
+
+* Integrators will have to map height ranges to specific archive nodes/networks to query history.
+
+### Neutral
+
+## References
+
+- [#2543: Allow genesis file to start from non-zero height w/ prev block header](https://github.com/tendermint/tendermint/issues/2543)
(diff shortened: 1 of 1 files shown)


ValarDragon:
I don't think any of the chain linking options differ in guarantees to any node types, but I strongly prefer weak chain linking.

The four different guarantees to consider are what does it do for a synced /unsynced full node, and synced / unsynced lite clients. Here is my understanding for why it doesn't affect security for them.

A synced full nodes guarantee state integrity by replaying all of the computation themselves. Correspondingly, they should replay the code to produce the new fork's genesis file and use that. There is then no need for them to reverify the new genesis, and chain linking didn't prove anything new.

A synced Lite clients guarantee state integrity by trusting 2/3rds of validators and then just tracking consensus. Even with strong chain linking, the lite client gets no guarantee that the newly derived state root is correct. The only way to actually efficiently check the new state root would be if you had a SNARK proof. (Moreover, one without preprocessing)

An unsynced full node/lite client needs to trust some genesis point. Maybe chain linking helps with this? (a bit hard to reason about, since this trust is based of a social consensus) But if so, imo weak chain linking should suffice. If they want stronger guarantees, they should just check the header using the prior TM version and some other tooling.

However, I do think weak chain linking is a good idea for overall clarity, and debuggability.

erikgrinaker:
Thanks @ValarDragon! @cwgoes @melekes Do you share the assessment above? I.e. would ICS or light client security be improved by a strong link from a hard forked chain to the last block of the previous chain, or would it be useless without having the full block history on the new chain?

cwgoes:
I do not think strong linking vs weak linking itself matters for IBC security - as @ValarDragon mentioned, the light client can't check the state root anyways - what absolutely *does* matter for IBC security, however, is whether evidence will remain valid across an upgrade, e.g. if validators misbehave after an upgrade has occurred but within the unbonding period (double-signing block 49, if the upgrade was at block 50), they need to be held accountable after the upgrade just as if no upgrade had occurred. This will require persistence of state in the SDK (easy) and Tendermint (not sure how difficult). Is this proposal intended to address that problem (it had been discussed previously), or is that out of scope?

cmwaters:
Just briefly looking into the evidence code, I don't think it will be too difficult to do. The evidence pool requires access to both the evidence db and the state db. If both of these persist after the upgrade i.e. have access to the validator sets at heights before the upgrade then the only problem lies with having access to the previous chainID and knowing at which heights to use the old chainID and which heights to use the new one.

melekes:
What sort of benefits does strong chain-linking provide? I can't see any personally. Unless you put pre-upgrade header into `LastBlockID` (~ actually link them together), but I do not see us benefiting from that in any way.

Weak chain-linking sounds good. chainID and header's hash should be enough imho for a) external tools to prove succession b) Tendermint to verify evidence from previous chain (we'll have to put pre-upgrade chainID into the state).

:+1: for initial height.

erikgrinaker:
Thanks for your feedback everyone. It sounds like we'll be using an `initial_height` genesis setting, in addition to weak chain linking (chain ID and either last block header or last block hash, ignored by Tendermint). I'll update the RFC with a concrete proposal.

@ebuchman In https://github.com/tendermint/tendermint/issues/2543 you mentioned wanting to link the chains more strongly (e.g. "Current code expects all blocks with height > 1 to have a valid LastCommit, and we probably want to preserve this property"), do you want to weigh in here?

> what absolutely does matter for IBC security (and light client security generally), however, is whether evidence will remain valid across an upgrade [...] Is this proposal intended to address that problem (it had been discussed previously), or is that out of scope?

I see, thank you for the clarification. State migration across upgrades is out of scope for this proposal.

> The evidence pool requires access to both the evidence db and the state db. If both of these persist after the upgrade i.e. have access to the validator sets at heights before the upgrade then the only problem lies with having access to the previous chainID and knowing at which heights to use the old chainID and which heights to use the new one

Yes, ideally we should maintain backwards compatibility with old state across upgrades, in which case we may not even need to hard fork at all. However, that is a separate issue requiring more spec and implementation work - this proposal is a stopgap solution until then.

josef-widder:
Does that mean that the `epoch` in the Heights defined in [ICS007(Height)](https://github.com/cosmos/ics/tree/master/spec/ics-007-tendermint-client#height) will not be used? Or it depends on the application whether they choose to increment the epoch or start with a fresh initial height?

cwgoes:
> I'll let @cwgoes weigh in, but since networks can choose whether to restart from 0 or not during a hard fork it should probably still be supported. Since evidence is not kept across hard forks anyway, we may want to increment the epoch for hard forks even when the height is not reset.

Currently, IBC expects the epoch to be encoded into the chain ID (this is a bit of a hack to ensure that the epoch is signed over, which it wouldn't be otherwise).

rfc/002-nonzero-genesis.md
```diff
@@ -0,0 +1,102 @@
...
+
+### Negative
+
+* Upgrades still cause loss of block history.
+
+* Integrators will have to map height ranges to specific archive nodes/networks to query history.
+
+### Neutral
+
+* The predecessor block header is not verified by Tendermint at all, and must be verified by
```
melekes, line 97: How would network governors verify it? I.e., imagine Alice has genesisA.json, Bob has genesisB.json. Since we do not hash a predecessor field, both Alice and Bob can have different predecessor blocks, and there's no way for me to find what is the "right" one. Then what's the reason for putting it in genesis in the first place.

erikgrinaker: Oh, that's a good point. I somehow assumed that the entire genesis file was hashed into the first block. Does this field even have any value if it can't be trusted or agreed upon?

melekes: no. I too thought we hash entire genesis into the first block or something like that a while ago, but no 😄

alexanderbez: I recall there was a proposal to do this long ago, but it never happened it seems.

erikgrinaker: Ok, in that case I think we should just remove it. What's your take @ValarDragon?

ValarDragon: How is the Tendermint blockhash initialized? I would have assumed "LastBlockHash" for the first block in a chain is initialized to `Hash(genesis)`. Seems quite important for things like ensuring you agree on initial consensus params, and that not ensuring all nodes agree on all of the genesis would be a bug. 

For generic hashchains, you run into security proof problems under adaptively chosen inputs unless you do something like this. (Its possible its not a problem for Tendermint, but I'd be a bit surprised)

erikgrinaker: `LastBlockID` and `LastCommitHash` are set to zero values for the initial block. All genesis fields (e.g. consensus params, validator sets, and the app hash) are hashed and/or handled individually in the genesis block, seems like the only field which isn't covered in some way is `GenesisTime`.

https://github.com/tendermint/tendermint/issues/4113 added a `--genesis_hash` parameter to verify the genesis file hash on startup. We could reuse e.g. `LastBlockID.Hash` or `LastCommitHash` for the genesis hash as well, but this might get confusing, especially with non-zero genesis or predecessor chains.

ebuchman: Isn't all the data in the genesis effectively hashed in to the header already? the header of the first block i think should use the genesis time since there's otherwise no commit to draw it from, and everything else should be hashed in two (consensus params, validators, app state).

we don't take a literal hash of the json file but we should be able to verify one. though I guess that requires computing the application's merkle state from the app_state_bytes in the genessis, which might be prohibitive if you're just trying to verify that a genesis file is correct.

Not sure the best way to do it but possible the first LastBlockID should include the hash of the header or maybe we introduce a block-0 that has the genesis as the first transaction or something

ebuchman: Huh I think the AppHash may not be currently making it into the first block because abci.InitChain does not return the app hash. It probably should?

erikgrinaker: > Isn't all the data in the genesis effectively hashed in to the header already?

Yes, mostly. This appears sufficient to secure the current genesis file. However, we're considering adding a new `predecessor` field, but that field then won't be hashed into any blocks, and thus cannot be trusted.

Since it cannot be trusted it could be a potential social engineering attack vector, and I think we should err on the side of caution and drop this field until we find a way to secure the entire genesis file. The field is not actually used for anything anyway, so it's not like we _need_ it.

> maybe we introduce a block-0 that has the genesis as the first transaction or something

Would it be horrible to add a block field that's only used in the initial block?

> the AppHash may not be currently making it into the first block because abci.InitChain does not return the app hash

Seems so. That probably means that the genesis file app hash isn't verified either. I opened https://github.com/tendermint/tendermint/issues/5177 for this, we'll see if we can squeeze this in for 0.34 (the RC is already out).

erikgrinaker:
Following discussion in https://github.com/tendermint/spec/pull/119#pullrequestreview-457226355 I've removed the `predecessor` field for chain linking, since it won't be hashed into the genesis block and thus can't be trusted. We can consider adding it later, e.g. if and when we hash the entire genesis file into the genesis block (https://github.com/tendermint/tendermint/issues/5183).

rfc/002-nonzero-genesis.md
ValarDragon: So should we open a separate issue in Tendermint to anchor the blockhash with the entire genesis file, and then do weak chain linking?

erikgrinaker: Yeah, I've added https://github.com/tendermint/tendermint/issues/5183

erikgrinaker:
Prototype implementation (https://github.com/tendermint/tendermint/pull/5191) was simplified greatly by including the initial height in the node state as `State.InitialHeight`, so I've updated the RFC with this.

ebuchman:
Haven't had a chance to review yet, sorry I'm late, will get to it next week (off the next few days). Just a note some potentially related work in https://github.com/tendermint/tendermint/blob/master/docs/architecture/adr-017-chain-versions.md incase it's helpful

melekes (commented):
lgtm

PR #120, opened 2020-07-27
spec: move reactor level specs to tendermint repo

marbar3778:
The reactor specs are implementation level and should not live with the client agnostic specs.


erikgrinaker:
So, there's a few different things being covered by this. The actual reactor implementation details should not live in the spec repo, but the descriptions and specifics of the P2P protocols should be in the repo. This includes stuff like messages, channel IDs, and the protocol semantics.

marbar3778:
making this a draft pr as a lot of the specs are intertwined right now.

marbar3778:
TODO:
[ ] add channels IDs
[ ] add p2p protocol messages
[ ] add encoding information. (Amino as 0.34 is not yet released)

marbar3778:
closing this due to discussion in the last dev call

PR #122, opened 2020-07-28
spec: revert event hashing

melekes:
Refs https://github.com/tendermint/tendermint/issues/5113


erikgrinaker:
Merge of `master` went horribly wrong, for some reason. Resubmitted as #132.

Issue #121, opened 2020-07-28
Repo cleanup and polish


ebuchman:
This repo is in a bit of a mixed and under maintained state.

It originally started as a home for the "Latest Gossip on BFT"  paper about the core Tendermint consensus algorithm.

It still houses that paper, and it's Github Releases are about that paper. 

But it is also now home to the [Tendermint Block and P2P Protocol Specs](./spec) and to the spec's RFC process.

We'd like it to be the canonical place for the Tendermint specification and for governance over changes to that spec.

We should:
- Update the README to summarize the state of the repo, the state of implementations, and how to contribute
- Establish spec versioning procedure. Releases of the repo should be about the full spec, not that consensus paper. We should have versioned releases of the spec and it should correspond to the Block Protocol and P2P Protocol versions in Tendermint Core - see https://github.com/tendermint/tendermint/blob/master/version/version.go#L31. What those versions cover / mean should be better described in this repo as well.
- Update details of the RFC process and decision making for breaking changes to the protocol

cmwaters:
Closing this since we're migrating the spec back into the Tendermint repo. Also many of the bullet points here have been addressed i.g. making the rfc process clear.

PR #123, opened 2020-07-28
blockchain: rename to core

marbar3778:
This pr renames blockchain to core. Personally I don't find blockchain that descriptive and a bit confusing as the entire spec describes a blockchain protocol. 

I am willing to rename core to something else, core is the first thing the came to mind. 

ref #121


josef-widder:
Renaming blockchain.md to datastructures.md etc. makes a lot of sense! I already have some plan what could go into a blockchain.md file in the future ;-)

ebuchman:
This will break a lot of external links. Might be nice to keep the old folder around with empty files that just say "Deprecated" and point to the new location ...

marbar3778:
> This will break a lot of external links. Might be nice to keep the old folder around with empty files that just say "Deprecated" and point to the new location ...

there are a few locations in Tendermint where this is the case. I would prefer to replace the old links with permalinks instead of keeping deprecated files around. Do you know of any other locations?

josef-widder:
We also refer to it in several specification here: https://github.com/informalsystems/tendermint-rs/tree/master/docs/spec

ebuchman:
Off hand I could imagine:

- repo docs in tendermint, cosmos-sdk, ics. others?
- websites and blog posts, eg. in tendermint and cosmos medium, etc.
- informal systems specs and papers 

You're right we should be using permalinks more. Sometimes there's incentive to use a non-perma-link because you know you're still working on it  and want the link to point to the latest. 

I don't know what the right solution is 🤷‍♂️

marbar3778:
ill add empty files. When we go to refactor the docs and specs then we can worry about removing them

marbar3778:
please see #103

PR #137, opened 2020-08-10
prepare repo for reunification

tessr:
This change prepares this repo for reunification, per the strategy outlined in our [Google doc](https://docs.google.com/document/d/1GTGK4zWiD8Rpukw9o5GxyRG0w0EcEhSyMjf-WsJzoak/edit#heading=h.jyngiyogpxjo). 

In this change:

* Move the old `spec` directory into a new `go-spec` directory
* Delete unused directories, like `papers` and `research`
* Update the readme to reflect the current state of the repository


tessr:
This is currently failing the link checker because the updated README refers to directories (on master) that don't exist yet. (They will soon, though.)

Next steps for this PR include: 

1. Making sure that this doesn't break things on tendermint.com/docs (cc @marbar3778)
2. Updating CODEOWNERS to reflect the individuals responsible for each spec (cc @brapse)

marbar3778:
> Making sure that this doesn't break things on tendermint.com/docs (cc @marbar3778)

this breaks the ecosystem (cosmos & tendermint) in terms of links. Once we have a unified sped will unified-spec become spec? this will be another ecosystem braking change.

erikgrinaker:
We should also update anything that links directly to the repo to instead link to the docs site, so that we can keep incoming spec links working throughout the migration. I'll do a pass of our main repos and submit PRs.

marbar3778:
> We should also update anything that links directly to the repo to instead link to the docs site, so that we can keep incoming spec links working throughout the migration. I'll do a pass of our main repos and submit PRs.

this bleeds to tendermint-rs, ibc repos, cosmos-sdk, gaia, and most likely our user docs if they link to our docs. I'm not sure if this approach should be taken, is there a way to have less breakage? We just went through this issue with renaming blockchain to core.

erikgrinaker:
>this bleeds to tendermint-rs, ibc repos, cosmos-sdk, gaia, and most likely our user docs if they link to our docs.

As long as we make sure the docs.tendermint.com links don't break, I think we should be fine changing the repo structure? Not sure if we need to guarantee link stability for the source code, as long as the published docs are stable.

tessr:
I think we can't make guarantees to the ecosystem about not changing directory structure. If people need persistent links, they should link to a commit rather than to master. And if people are linking to master because they want the most up-to-date information on the spec... well, I think it's OK that it shows things are a little broken! The latest situation _is_ a little bit broken, and if people need to do a little legwork to figure out exactly which spec they should be referencing, I think that's OK for now. Better for things to be explicitly messy than incorrect-but-misleading. 

> But I'd be fine with just keeping the Go spec as spec/ until unified-spec is complete as well, or possibly as a symlink to go-spec.

Personally I don't feel super strongly about the names, but I do think it's important that we don't treat the current Go spec as the "default." The new unified spec should start as a blank canvas, and we should only migrate components into it once members from both the Informal team and the Interchain team have ratified them.

marbar3778:
> But I'd be fine with just keeping the Go spec as `spec/` until `unified-spec` is complete as well, or possibly as a symlink to `go-spec`.

can you sym-link github links?

> As long as we make sure the docs.tendermint.com links don't break, I think we should be fine changing the repo structure? Not sure if we need to guarantee link stability for the source code, as long as the published docs are stable.

https://github.com/tendermint/spec/pull/123#issuecomment-665170771. we just had this conversation and I was for allowing breakage but we decided to go with stability. I was persuaded to stability especially since this is an open-source project that gets used heavily and gets linked to heavily in papers.

erikgrinaker:
>can you sym-link github links?

Apparently not. ☹️

If we have a ton of incoming links to the spec repo itself then I don't think the marginal benefit of renaming `spec` to `go-spec` is worth it. Let us be frustrated over it not being called `go-spec` instead of our users being frustrated that everything is broken. We can clean this up once and for all when the final unified spec is done, including explicitly recommending people to link to a published spec instead of the source repo lest their links will break. And let's make sure it really is final before doing so.

tessr:
> Let us be frustrated over it not being called go-spec instead of our users being frustrated that everything is broken. We can clean this up once and for all when the final unified spec is done, including explicitly recommending people to link to a published spec instead of the source repo lest their links will break.

This sounds reasonable to me, but it raises another concern. In a world where there are `go-spec` and `rust-spec` directories acting as holding zones for non-unified specs, it's easy to mark components as ratified/unified once they move over into the unified directory. e.g., if you are working on the light client evidence handling (LCEH) spec, you can consider it ratified once it's living in the `unified-spec` directory and has been removed from the `go-spec` and `rust-spec` directories.

But in a world where the current Go specs live under `spec`, what happens once we unify that spec?

(I'll keep using LCEH as an example, because Josef, Zarko, Anton, and Callum are all working on getting that spec into a place where it will be unified for both Go and Rust; and I expect it will be our first unified piece of the spec.) 

Will we keep the old LCEH specs in https://github.com/tendermint/spec/tree/master/spec/consensus/light-client? (It's potentially confusing, especially if people are still linking to it.) Will we update them to match the LCEH in `unified-spec/`? (Not ideal, since it means we still have two canonical specs for LCEH.) Will we delete the old LCEH stuff from `/spec`? (That breaks links anyways, and it means that as we get closer and closer to having everything unified, we'll have more and more broken links.) 

One maybe wacky idea is to have `spec` be the unified spec directory, so that links are _restored_ as we repopulate that directory. Alternately, we  could also update the files in the spec repo to have little manual deprecated notices with links (like @ebuchman's suggestion here: https://github.com/tendermint/spec/pull/123#issuecomment-665170771). 

I feel a little bit like I'm bikeshedding here, but I think we want to be precise in this case. A lot of the confusion over (for example) the state of the LCEH spec came from having specs in too many different places, so it's worth being thoughtful about how we unify.

erikgrinaker:
I would consider the new, unified spec a work in progress until we finalize it and give it a version number. The spec is not ratified until it has a version. At that point we can publish it and remove the old specs.

While we are working on the new unified spec, we will also have to keep the current spec updated with changes in the Go implementation - unless we want to freeze all protocol changes until the new spec is ready, which I'm guessing is not viable.

If we want to "move" parts from the Go/Rust specs into the unified spec while we are working on them, I suggest we do this in directories separate from the currently published spec. People still rely on the current spec.

ebuchman:
As a general comment can we leave the old directory structure in place under the `spec` dir and start fresh with copies into new dirs so it's easier to avoid breaking links?

ebuchman:
Also what are the divergences between the specs in tendermint-rs that justify an independent go-spec and rust-spec? Can't we just bring it all together right away?

marbar3778:
> Also what are the divergences between the specs in tendermint-rs that justify an independent go-spec and rust-spec? Can't we just bring it all together right away?

I believe the only divergence is around the light client. Does Tendermint-rs have any other specs?

We can leave the current structure and move the current light client spec into its own folder and then add informals spec in sub dir? Or both go into a sub dir?

tessr:
Ah, interesting, the discrepancies might not be as dramatic as I thought. Are the rust specs just for the light client and fast sync? https://github.com/informalsystems/tendermint-rs/tree/master/docs/spec

erikgrinaker:
> My original PR matches the decision we made in the Dev Call last week (and the doc I wrote afterwards to make sure everyone was on the same page). I'm curious how that conversation and that doc could lead to a different conclusion than the one we seem to be getting to here. Did we do something wrong in that call, or were we missing information?

I think we all agreed (and still agree) on the overall approach of starting with the Go and Rust specs and moving pieces into a unified spec as they are reviewed and updated. However, let's not break the current spec until the new spec is ready, e.g. by making a working copy. I think this was commented on in the doc as well.

tessr:
Well, this comment (from @ebuchman) made me rethink that, as well: 

> Also what are the divergences between the specs in tendermint-rs that justify an independent go-spec and rust-spec? Can't we just bring it all together right away? 

Which I think is a reasonable question! When I went and looked in tendermint-rs, the specs were a lot lighter than I expected after Thursday's conversation. 

Anyways, this is all a long way of saying that I sort of feel like we're back at square one again with this process. @brapse, as someone who was advocating for this on the Informal side, what are you thinking at this point?

tessr:
Closing this and will open a new PR when we have a new plan in place.

Issue #116, opened 2020-07-12
DuplicateVoteEvidence should be valid even if Vote.ValidatorIndex does not match


ebuchman:
According to the [spec] and the [Go impl], both Votes in a DuplicateVoteEvidence must also have the same Vote.ValidatorIndex (position of the validator in the validator set). However this would only be true if the validator sets for the respective blocks being signed were the same, or had the same order, which might be the case for the full node, but not necessarily for the light client.  

So I think we need to remove the requirement that the ValidatorIndex matches

[spec]: https://github.com/tendermint/spec/blob/master/spec/blockchain/blockchain.md#evidence-1
[Go impl]: https://github.com/tendermint/tendermint/blob/master/types/evidence.go#L287

cmwaters:
I tend to agree. 

I think the only characteristic that using the index brings for the full node, is that if there is a fork then in subsequent blocks where the validator sets diverge, the node won't consider DuplicateVoteEvidence if a node in both sets signs for two blocks in either of the forks. Whereas omitting the check would mean that we would still produce evidence.

As it stands for the light client, because it will most likely be that the conflicting header isn't the header with the common ancestor, an attempt to detect a full fork will most likely fail detection because the validator sets could at the height of the conflicting headers be different. 

I guess looking from the other angle, I can't see a situation where removing this check will allow for "honest" behaving nodes to be framed with DuplicateVoteEvidence

Last thing I wanted to note is that AmnesiaEvidence also has the same check

ebuchman:
Should we fix this for v0.34 tendermint?

cmwaters:
I will open an issue in the go tendermint repo. It should be just a few line change to get it done so I don't see why not

marbar3778:
could you add validation rules to the evidence section in data_structures? This is missing for all evidence.

PR #139, opened 2020-08-11
Add evidence verification to data structures

cmwaters:
Closes: #116

I have added a synopsis of the validation code for evidence in the go implementation with the corresponding data structures. This also implies the removal of checking that the validator indexes match in `DuplicateVoteEvidence`. Would be nice to get this cross-checked.

### DIFF
--- a/spec/core/data_structures.md
+++ b/spec/core/data_structures.md
@@ -221,7 +221,9 @@ type Evidence interface {
 
 All evidence can be encoded and decoded to and from Protobuf with the `EvidenceToProto()` 
 and `EvidenceFromProto()` functions. The [Fork Accountability](../consensus/light-client/accountability.md) 
-document provides a good overview for the types of evidence and how they occur.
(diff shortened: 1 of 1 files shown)


spec/core/data_structures.md
```diff
@@ -236,6 +238,20 @@ type DuplicateVoteEvidence struct {
...
+
+- Validator Address, Height, Round and Type of vote must be the same for both votes
+
+- BlockID must be different for both votes (BlockID can be for a nil block)
+
+- Validator must have been in the validator set at that height
+
+- Vote signature must be valid (using the chainID)
+
+- Time must be equal to the block time
```
ebuchman, line 251: Isn't the Timestamp in the votes the validator's local clock? Not sure there's any validation we can/should do on the Timestamp (and I don't see any in the current code)

ebuchman: Nvm was looking at outdated version.

PR #144, opened 2020-08-20
RFC: adopt zip 215

marbar3778:
Short RFC to adopt zip215 for Tendermint use cases. 

Read ZIP215 here: https://zips.z.cash/zip-0215


melekes (approved):
👍 

1) what library tendermint-rs is using right now?
2) was https://github.com/hdevalence/ed25519consensus verified by someone outside of Zcash? Do we want to perform any additional testing before adopting it?

marbar3778:
>     1. what library tendermint-rs is using right now?
ed22519-dalek
>     2. was [hdevalence/ed25519consensus](https://github.com/hdevalence/ed25519consensus) verified by someone outside of Zcash? Do we want to perform any additional testing before adopting it?

we can perform additional testing. Filippo did some work on the library as well. The part of the library that we will use is only the verify function, everything else is still under the go std library.

erikgrinaker:
Makes sense to adopt some standard, don't have an informed opinion as to which one. Will this have to be a breaking change, or can we support both the old and new schemes in a transition period?

marbar3778:
> Makes sense to adopt some standard, don't have an informed opinion as to which one. Will this have to be a breaking change, or can we support both the old and new schemes in a transition period?

It is not breaking but it is best to have it be handled on a new chain instead of an upgrade of version

zmanian:
> 
> 
> 1. what library tendermint-rs is using right now?
> 2. was https://github.com/hdevalence/ed25519consensus verified by someone outside of Zcash? Do we want to perform any additional testing before adopting it?

Fillipo who is the maintainer for `x/crypto` in the Go standard library also looked at it.

rfc/003-ed25519-verification.md
```diff
@@ -0,0 +1,41 @@
...
+
+### Positive
+
+- Batch verification
+- Signature verification across implementations
+
+### Negative
+
+### Neutral
+
```
ebuchman, line 38: Should reflect that the library does not come with the eg. the std go crypto libs, even though Filipp'os reviewed it, it's in some sense a more external dependency. And if we only use it for verification (ie. signing is done using the default libs) then we have additional imports and to remember to use the right one for verification. Worth mentioning these consequences

 we'd be using some standard lib for most operations (signing) and only the zip215 lib for the verification.

marbar3778: Added. I put it in a sub section in negative for Tendermint-go.

tessr:
I'm generally in favor of adopting ZIP 215. I'd love to see this RFC address a few more things before we merge it:

1. Concerns around breaking-ness. Although technically not a breaking change, it is a change that needs to be coordinated so that all validators on a network can verify signatures consistently with each other.* 
2. Anything we know about proposed plans regarding the future of this new library. 
3. If we want to do any security review on this and if this impacts other requested work, like signature aggregation or batch verification. 

----

*To expand on the "breaking-ness" of this change: 

1. Old signatures (e.g. signatures created with the non-ZIP-215 ed5519 library) can be validated by new verification functions (e.g. the verification function from the ZIP-215 library). 
2. New signatures can also be validated with old verification functions; _however_, signatures can be maliciously crafted to validate with the new verification functions but not the old verification functions. This means that someone could deliberately create a malicious signature which would split (and halt) the network. 
3. Therefore, it’s not “breaking” to adopt ZIP 215, but it is important to coordinate upgrades all together. I raised this over in the Zcash Discord, and Henry agreed that it would be good to treat this change as a breaking one.

robert-zaremba:
In [SDK](https://github.com/cosmos/cosmos-sdk/issues/7315), I was mentioning [BIP-340](https://github.com/bitcoin/bips/blob/master/bip-0340.mediawiki).

Are you going to use `ed25519`? @marbar3778 mentioned that you are working on `sr25519` (W3F researched it for Polkadot), is it different than ZIP-215? Seams like `sr25519` could be a safe bet.

marbar3778:
> Are you going to use `ed25519` with ZIP-215? @marbar3778 mentioned that you are working on `sr25519` (W3F researched it for Polkadot), is it different than ZIP-215? Seams like `sr25519` could be a safe bet.

Sorry for the confusion. I meant that it is merely supported in the crypto package. We do not utilize the key type in Tendermint.

rfc/003-ed25519-verification.md
```diff
@@ -0,0 +1,41 @@
...
+## Status
+
+Proposed
+
+## Consequences
+
+### Positive
+
+- Batch verification
```
robert-zaremba, line 32: hdevalence/ed25519consensus doesn't support batch verification: [godoc](https://pkg.go.dev/github.com/hdevalence/ed25519consensus)
This library is a good example with ed25519 batch verification: https://github.com/oasisprotocol/ed25519

marbar3778: Will be more explicit, this will come at a later date. 

As for oasis protocol library, we discussed and would only use for our fast sync process and not for consensus. The verification function in `hdevalence/ed25519consensus` handles sig verification in more detail, thus we would be able to utilize it for consensus verification. Since this is not supported at this time I decided to leave it out of this conversation. Batch verification is an implementation level decision as well, no need to have the conversation in the spec repo.

robert-zaremba: So let's have a section  _further ideas_ and let's put it there.

hdevalence:
Hi all, I wrote up a more detailed post on the background behind the ZIP215 rules: https://hdevalence.ca/blog/2020-10-04-its-25519am

rfc/003-ed25519-verification.md
```diff
@@ -0,0 +1,60 @@
...
+- Marko (@marbar3778)
+
+## Context
+
+Ed25519 keys are the only supported key types for Tendermint validators currently. Tendermint-Go wraps the ed25519 key implementation from the go standard library. As more clients are implemented to communicate with the canonical Tendermint implementation (Tendermint-Go) different implementations of ed25519 will be used. Due to [RFC 8032](https://www.rfc-editor.org/rfc/rfc8032.html) not guaranteeing implementation compatibility, Tendermint clients must to come to an agreement of how to guarantee implementation compatibility. [Zcash](https://z.cash/) has multiple implementations of their client and have identified this as a problem as well. The team at Zcash has made a proposal to address this issue, [Zcash improvement proposal 215](https://zips.z.cash/zip-0215).
+
+## Proposal
+
+- Tendermint-Go would adopt [hdevalence/ed25519consensus](https://github.com/hdevalence/ed25519consensus).
+    - This library is implemented as an extension of the go standard library.
```
hdevalence, line 18: It might be good to be slightly more specific here and mention that the `ed25519consensus` library only provides verification, as `ed25519consensus.Verify`.  So presumably Tendermint-Go would use `crypto/ed25519` for key generation and signing, but use the more precisely specified `ed25519consensus.Verify` for verification.

rfc/003-ed25519-verification.md
```diff
@@ -0,0 +1,60 @@
...
+    -
+- Tendermint-rs would adopt [ed25519-zebra](https://github.com/ZcashFoundation/ed25519-zebra)
+    - related [issue](https://github.com/informalsystems/tendermint-rs/issues/355)
+
+As signature verification is one of the major bottlenecks of Tendermint-go, if ZIP 215 is adopted batch verification of signatures will be safe in consensus critical areas.
```
hdevalence, line 23: TIOLI: it might be clearer to restructure this in Situation-Complication-Resolution form: signature verification is a bottleneck (situation), batch verification can't be used unless it has the same consensus rules (complication), zip215 makes batch and single verification agree (resolution)

rfc/003-ed25519-verification.md
```diff
@@ -0,0 +1,60 @@
...
+
+The change has been recommended to be done in a major release. Although it is not a breaking change if this change were to be rolled out in a minor release a node could construct signatures which could only be verified by nodes which have received updates to the verification function resulting in a possible network halt.
```
hdevalence, line 25: I think this description is not quite right: in terms of consensus compatibility, this *is* a breaking change, even though in practice it's likely to have no observable effect, because (1) it's backwards compatible with all existing signatures and (2) it's incompatible only on specially-crafted cases that probably won't be part of the chain either before or after the upgrade.

However, because (as you mention here) a node *could* submit those specially-crafted cases, it does still have to be treated as a breaking change.

hdevalence:
I left some review comments -- I realize that this is somewhat of a "drive-by review" since I don't have a lot of context about Tendermint, so my apologies if I missed something because of that, and I hope the comments are helpful anyways.

rfc/003-ed25519-verification.md
hdevalence: FWIW, I think it makes sense to classify the *ability* to perform batch verification as a "now" positive rather than in a "further ideas" section, because it's something that's enabled immediately by the proposed change, whereas "further ideas" seems like a place for changes that are not proposed now but could be made in the future.   In the case of batch verification, if ZIP215 were adopted, using batch verification in Tendermint-Go wouldn't require any further changes to the spec, it would be purely an implementation optimization.

(Of course, this is just my opinion, and I don't have full context on Tendermint or its processes, so please feel free to discount it appropriately)

rfc/003-ed25519-verification.md
```diff
@@ -0,0 +1,60 @@
...
+- Enable safe batch verification
+    - This has not yet been implemented.
+
+### Negative
+
+#### Tendermint-Go
+
+- Third_party dependency
+    - library has not gone through a security review.
+    - unclear maintenance schedule
```
hdevalence, line 52: In re: this line and the comment made by @tessr here: https://github.com/tendermint/spec/pull/144#issuecomment-690967591 ,

> Anything we know about proposed plans regarding the future of this new library (e.g., is Filippo and/or Google going to take it over? I couldn't quite track the conversation here).

 as far as I know there are no current plans to upstream the library to Filippo or Google, so definitely don't count on that.

The maintenance aspect of that library is definitely a weak point. I'm not a Go expert and I'm only casually familiar with its ecosystem (e.g., I had no idea how to publish a package and I'm still not sure I did it right), so ideally someone who has those skills could help with that (or take it over).

That said, once the library is complete I think that it should require very little maintenance, because it has a tiny API surface, and it solves a problem with a fixed specification (ZIP215).

What has to happen for it to be complete?  Basically there are two changes. First, it should be changed to use https://pkg.go.dev/filippo.io/edwards25519 internally, now that that library exists.  Second, I'd like it to support batch verification with an init-update-finalize style API like `ed25519-zebra`.  Currently this isn't scoped out on the repo's roadmap because I've been pretty busy, and I'm not sure when I would have time to make those changes, but I'd be happy to mentor someone else who'd like to have a go at making them.

rfc/003-ed25519-verification.md
```diff
@@ -0,0 +1,56 @@
...
+    - This library is implements `ed25519.Verify()` in accordance to zip-215. Tendermint-go will continue to use `crypto/ed25519` for signing and key generation.
+
+- Tendermint-rs would adopt [ed25519-zebra](https://github.com/ZcashFoundation/ed25519-zebra)
+    - related [issue](https://github.com/informalsystems/tendermint-rs/issues/355)
+
+Signature verification is one of the major bottlenecks of Tendermint-go, batch verification can not be used unless it has the same consensus rules, ZIP 215 makes verification safe in consensus critical areas.
+
+This change constitutes a breaking changes, therefore must be done in a major release. No changes to validator keys or operations will be needed for this change to be enabled.
+
+This change has no impact on signature aggregation. To enable this signature aggregation Tendermint will have to use a different curve i.e [BLS](https://en.wikipedia.org/wiki/Boneh%E2%80%93Lynn%E2%80%93Shacham). Secondly, this change will enable safe batch verification for the Tendermint-Go client. Batch verification for the rust client is already supported in the library being used.
```
robert-zaremba, line 27: `curve25519` can also be used for key aggregation (eg Schnorr)

robert-zaremba: So, no need to use a different curve.

marbar3778: true, will amend this paragraph.

rfc/003-ed25519-verification.md
```diff
@@ -0,0 +1,56 @@
...
+    - This library is implements `ed25519.Verify()` in accordance to zip-215. Tendermint-go will continue to use `crypto/ed25519` for signing and key generation.
+
+- Tendermint-rs would adopt [ed25519-zebra](https://github.com/ZcashFoundation/ed25519-zebra)
+    - related [issue](https://github.com/informalsystems/tendermint-rs/issues/355)
+
+Signature verification is one of the major bottlenecks of Tendermint-go, batch verification can not be used unless it has the same consensus rules, ZIP 215 makes verification safe in consensus critical areas.
+
+This change constitutes a breaking changes, therefore must be done in a major release. No changes to validator keys or operations will be needed for this change to be enabled.
+
+This change has no impact on signature aggregation. To enable this signature aggregation Tendermint will have to use different keys (sr25519, BLS). Secondly, this change will enable safe batch verification for the Tendermint-Go client. Batch verification for the rust client is already supported in the library being used.
```
robert-zaremba, line 27:
```suggestion
This change has no impact on signature aggregation. To enable this signature aggregation Tendermint will have to use different signature schema (Schnorr, BLS, ...). Secondly, this change will enable safe batch verification for the Tendermint-Go client. Batch verification for the rust client is already supported in the library being used.
```
In case of ed25519 and Schnorr keys can stay the same.

PR #155, opened 2020-09-04
spec/reactors/mempool: batch txs per peer

melekes:
Refs https://github.com/tendermint/tendermint/issues/625

### DIFF
--- a/spec/reactors/mempool/reactor.md
+++ b/spec/reactors/mempool/reactor.md
@@ -13,6 +13,11 @@ for details.
 Sending incorrectly encoded data or data exceeding `maxMsgSize` will result
 in stopping the peer.
 
+`maxMsgSize` equals `MaxTxBytes` (1MB) * `TxBatchSize` (100) + 4 (proto
+overhead). `MaxTxBytes` and `TxBatchSize` are mempool config parameters ->
+defined locally. The reactor sends transactions to the connected peers in
+batches. The maximum size of one batch is `TxBatchSize`.
+
 The mempool will not send a tx back to any peer which it received it from.
 
 The reactor assigns an `uint16` number for each peer and maintains a map from
--- a/spec/reactors/mempool/reactor.md
+++ b/spec/reactors/mempool/reactor.md
@@ -13,10 +13,10 @@ for details.
 Sending incorrectly encoded data or data exceeding `maxMsgSize` will result
 in stopping the peer.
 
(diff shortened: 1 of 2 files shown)


marbar3778 (approved):
LGTM. Do we want one approval from IG and one from Informal?

spec/reactors/mempool/reactor.md
```diff
@@ -13,6 +13,11 @@ for details.
 Sending incorrectly encoded data or data exceeding `maxMsgSize` will result
 in stopping the peer.
 
+`maxMsgSize` equals `MaxTxBytes` (1MB) * `TxBatchSize` (100) + 4 (proto
```
ValarDragon, line 16: Can we set maxMsgSize as a config parameter independently from MaxTxBytes and TxBatchSize?

e.g. suppose I expect all txs to be 1KB, want a batch size of 1000, and want to allow occasional txs at the maxTxBytes (1MB).

In this setting, it seems reasonable to me to habe a maxMsgSize of around 1 to 2 MB, whereas this parameterization would net me 1GB.

spec/reactors/mempool/reactor.md
```diff
@@ -13,6 +13,11 @@ for details.
...
+overhead). `MaxTxBytes` and `TxBatchSize` are mempool config parameters ->
+defined locally. The reactor sends transactions to the connected peers in
+batches. The maximum size of one batch is `TxBatchSize`.
```
ValarDragon, line 19: Do peers get told your MaxTxBytes and TxBatchSize in the handshake?

ValarDragon (changes requested):
Glad that batch mempool txs are getting added!

melekes:
> LGTM. Do we want one approval from IG and one from Informal?

sure

spec/reactors/mempool/reactor.md
melekes: sorry, I just updated the spec to align with latest implementation. Now, there's a single parameter for maxMsgSize - MaxBatchSize (10MB).

spec/reactors/mempool/reactor.md
melekes: Not yet -> https://github.com/tendermint/tendermint/issues/5332

alexanderbez (approved):

brapse (approved):
👍

PR #157, opened 2020-09-09
Fastsync spec from tendermint-rs

josef-widder:
This contains the fastsync spec from tendermint-rs. As it was done for the Go implementation, I think we can directly move it to spec/spec

https://github.com/informalsystems/tendermint-rs/issues/561


tessr (approved):
Thank you for doing this! You'll delete it from tendermint-rs, too, right?

spec/fastsync/fastsync.md
```diff
@@ -0,0 +1,1216 @@
...
+It is the same as the sequential version
+[**[FS-SEQ-SAFE-STATE]**](#FS-SEQ-SAFE-STATE).
+
+#### **[FS-DIST-NONABORT]**
+
+If there is one correct process in *peerIDs* [FS-SOME-CORR-PEER],
+*Fastsync* never terminates with failure. (Together with [FS-DIST-LIVE]
+ that means it will terminate successfully.)
+
+# Part IV - Fastsync protocol
```
marbar3778, line 386: This sections is specific to the go implementation or will the rust implementation be following the same spec?

marbar3778 (commented):
How does this differ from block_sync from the spec/reactor section? This feels like it should be moved there?

josef-widder:
We did the fastsync spec for Golang implementations. Therefore the files should go into specs

josef-widder:
> Thank you for doing this! You'll delete it from tendermint-rs, too, right?

The plan is to replace the files in tendermint-rs with files that just contain links to the new location, because we linked to them in reports and research papers

marbar3778:
> We did the fastsync spec for Golang implementations. Therefore the files should go into specs

The spec repo's aim is to be implementation agnostic.
Are we shifting from this and housing implementation specs in the spec repo? I don't have a preference for which path is taken, but we should document it.

josef-widder:
> How does this differ from the `block_sync` section in spec/reactor? This feels like it should be moved there?

These specs looks specific to v0 and v1. The fastsync spec should address all of those. Whether it should go there I do not know. I am not sure I understand the structure of the directories, and what is considered a "reactor" and what not.

marbar3778 (approved):
Thank you for migrating these specs. Can't wait to dive into them.

PR #158, opened 2020-09-09
Current versions of light client specs from tendermint-rs

josef-widder:
https://github.com/informalsystems/tendermint-rs/issues/561


tessr (approved):
Fantastic, thanks for doing this! Also, I think CI is failing because the link checker running in CI got rate-limited, not for any real reason.

PR #160, opened 2020-09-10
Update README.md

josef-widder:
added link to rust-spec


josef-widder:
Again the same CI problem with http errors 429. We should fix that. I don't know who can do that...

marbar3778 (approved):
LGTM. 

I'll look into ways of fixing this

tessr:
The problem with the CI tool is that Github is rate-limiting it, I think. So the first _n_ checks will pass, and then Github will say, "No more!" I think the solutions are that we either modify our CI job to check fewer links (maybe only in files that were touched?) or somehow authenticate the requests so that GH doesn't rate-limit them.

Issue #105, opened 2020-06-18
evidence: spec in blockchain out of date


marbar3778:
The Blockchain spec says we only have one type of evidence currently. 

This is not true anymore. We should update to reflect the new types of evidence.

https://docs.tendermint.com/master/spec/blockchain/blockchain.html#evidence-2 

assigning callum & anton as they have the most domain knowledge here.

Issue #104, opened 2020-06-17
Write specification of on chain fork detection protocol


melekes:
We need to write precise specification of execution of fork detection protocol on chain. This includes interaction with a lite client (once lite client detects a fork), evidence creation, gossiping of evidence, committing evidence as part of block, and then evidence processing as part of block execution. The specification should define precisely various terms (fork, slashable fork, evidence, height vote set as part of challenge response, etc) and also contain diagram of interactions between various parties.

_(the issue was moved from https://github.com/tendermint/tendermint/issues/4217)_

josef-widder:
Here is the doc that gives an overview over several documents and issues: https://docs.google.com/document/d/1LMVUrdZ6r2G_gD9Yc8nr4RC-SEkgYYPR3DAKSdFdY7M/edit?ts=5eff62aa#heading=h.qlzor1kzfmon

PR #164, opened 2020-09-18
spec: Light client attack detector

josef-widder:
This English spec ([rendered](https://github.com/tendermint/spec/blob/dcc3b974c63470fac1424f69bb7831fe114e99d6/rust-spec/lightclient/detection/detection.md)) implements the ideas from the [evidence handling document](https://github.com/tendermint/spec/blob/master/rust-spec/lightclient/attacks/evidence-handling.md) on the light node side.

It addresses: #104 #105 https://github.com/informalsystems/tendermint-rs/issues/494


rust-spec/lightclient/detection/detection.md
```diff
@@ -1,50 +1,68 @@
...
+by generating (signing) a block that deviates from the block of the
+same height on the blockchain. In order to do so, some of these faulty
+full nodes must have been validators before and violate
+[[TMBC-FM-2THIRDS]](TMBC-FM-2THIRDS-link), as otherwise, if
+[[TMBC-FM-2THIRDS]](TMBC-FM-2THIRDS-link) would hold,
+[verification](verification) would satisfy
+[[LCV-SEQ-SAFE.1]](LCV-SEQ-SAFE-link).
+
+An attack detector (or detector for short) is a mechanism that is used
+by the light client [supervisor](supervisor) after
```
milosevic, line 38: Supervisor link is broken.

rust-spec/lightclient/detection/detection.md
```diff
@@ -1,50 +1,68 @@
...
+[verification](verification) of a new light block
+with the primary, to cross-check the newly learned light block with
+other peers (secondaries).  It expects as input a light block with some
+height *root* (that serves as a root of trust), and a verification
+trace (a sequence of lightblocks) that the primary provided.
+
+In case the detector observes a light client attack, it computes
+evidence data that can be used by Tendermint full nodes to isolate a
+set of faulty full nodes that are still within the unbonding period
+(more than 1/3 of the validator set of some block on the chain),
```
milosevic, line 48:
```suggestion
(more than 1/3 of the voting power of the validator set at some block of the chain),
```

rust-spec/lightclient/detection/detection.md
```diff
@@ -1,50 +1,68 @@
...
-eventually delivered) and timely communicate with a correct full node. If
-this assumption is violated, the light client can be fooled to trust a
-header that was not generated by Tendermint consensus.
-
-This specification, the fork detector, is a "second line of defense",
-in case the 1/3 assumption is violated. Its goal is to detect fork (conflicting headers) and collect
+[[TMBC-FM-2THIRDS]](TMBC-FM-2THIRDS-link).  It is safe under this
+assumption, and live if it can reliably (that is, no message loss, no
+duplication, and eventually delivered) and timely communicate with a
+correct full node. If this assumption is violated, the light client
```
milosevic, line 59:
```suggestion
correct full node. If [[TMBC-FM-2THIRDS]](TMBC-FM-2THIRDS-link) assumption is violated, the light client
```

rust-spec/lightclient/detection/detection.md
```diff
@@ -54,39 +72,55 @@ project.
...
-> **TODO** be more precise. +1/3 of b.NextV signed c. For now
-> the following will do:
+- *t - trustingPeriod < b.Header.Time < t*
+- the voting power in *b.NextValidators* of nodes in *c.Commit*
+  is more than 1/3 of *TotalVotingPower(b.Header.NextValidators)*
 
-Let *b* and *c* be two light blocks. We define **supports(b,c,t)**
-    iff `ValidAndVerified(b, c)` at time *t*
+> That is, if the [Tendermint failure model](TMBC-FM-2THIRDS-link)
+> holds, then *c* has been signed by at least on correct full node, cf.
```
milosevic, line 113:
```suggestion
> holds, then *c* has been signed by at least one correct full node, cf.
```

milosevic, line 113: Do we need to add an additional assumptions that b comes from Tendermint chain?

rust-spec/lightclient/detection/detection.md
```diff
@@ -105,80 +139,194 @@ time t there exists an *h* and a sequence *a(1)*, ... *a(h)* s.t.
 - *a(h) = c* and
 - *supports( a(i), a(i+1), t)*, for all i, *1 <= i < h*.
 
-> **TODO** In the above we might use a sequence of times t(i). Not sure.
-> **TODO:** I believe the following definition
-> corresponds to **Slashable fork** in
-> [forks][tendermintfork]. Please confirm!  
-> Observe that sign-skip-match  is even defined if there is a fork on
-> the chain.
+We call such a sequence *a(1)*, ... *a(h)* a **verification trace**.
```
milosevic, line 142: Why skipping condition is called skip-root?

rust-spec/lightclient/detection/detection.md
```diff
@@ -105,80 +139,194 @@ time t there exists an *h* and a sequence *a(1)*, ... *a(h)* s.t.
...
-implies *b = c*.
+implies *b.Header = c.Header*.
 
-#### **[TMBC-SIGN-FORK.1]**
+> Observe that *sign-skip-match* is defined via an implication. If it
+> evaluates to false this means that the left-hand-side of the
+> implication evaluates to true, and the right-hand-side evaluates to
+> false. In particular, there are two **different** headers *b* and
+> *c* that both can be verified from a common block *a* from the
+> chain. Thus, thee following describes an attack.
```
milosevic, line 169:
```suggestion
> chain. Thus, the following describes an attack.
```

rust-spec/lightclient/detection/detection.md
```diff
@@ -54,39 +72,55 @@ project.
...
 Let *Genesis* be the agreed-upon initial block (file).
 
-#### **[TMBC-FUNC.1]**
+#### **[TMBC-FUNC-SIGN.1]**
 
-> **TODO** be more precise. +2/3 of b.NextV = c.Val signed c. For now
-> the following will do:
+Let *b* and *c* be two light blocks with *b.Header.Height + 1 =
+c.Header.Height*. We define the predicate **signs(b,c)** to hold
+iff *c.Header.LastCommit* is in *PossibleCommit(b)*.
```
cmwaters, line 100: Is this the same as the two conditions b.NextValidatorsHash = c.ValidatorsHash && 2/3 of these Validators signed c?

rust-spec/lightclient/detection/detection.md
```diff
@@ -105,80 +139,194 @@ time t there exists an *h* and a sequence *a(1)*, ... *a(h)* s.t.
...
+- from attack we have E2(h-1), and as there is no evidence for
+  i = h - 1 we get **NOT E1(h-1)**. Contradiction.
+QED.
+
+#### **[TMBC-LC-EVIDENCE-DATA.1]**
+
+To prove the attack to p1, because of Point E1, it is sufficient to
+submit
+
+- v(i).Height (rather than v(i)).
```
milosevic, line 269: Callum made an observation that in case of Equivocation and Amnesia conflicting blocks are for the same height and they have the same validator set. So we don't need root height, i.e., we could set root height to v(i+1).Height. But maybe this can be seen as an optimisation (I think this is how it is being implemented in Go version).

rust-spec/lightclient/detection/detection.md
```diff
@@ -187,57 +335,68 @@ Let *b* be a light block and *t* a time. We define *bogus(b,t)* iff
...
 
-#### **[LCD-IP-PEERSET.1]**
+- `root-of-trust` is a light block that is trusted (that is,
+except upon initialization, the primary and the secondaries
+agreed on in the past), and
+- `verifiedLS` is a lightstore that contains a verification trace that
+  starts from a lightblock that can be verified with the
+  `root-of-trust` in one step and ends with a lightblock of the height
+  requested by the user
+- `Evidences` is a list of evidence for misbehavior
```
milosevic, line 386:
```suggestion
- `Evidences` is a list of evidences for misbehavior
```

rust-spec/lightclient/detection/detection.md
```diff
@@ -276,421 +443,356 @@ This implies that we need a timeout of at least *2 Delta* for remote
...
+                Evidences.Add(EvidenceForPrimary);
+            }
+            // At this point we do not care about the other error
+            // codes. We already have generated evidence for an
+            // attack and need to stop the lightclient. It does not
+            // help to call replace_primary. Also we will use the
+            // same primary to check with other secondaries in
+            // later iterations of the loop
+         }
+         else {
```
milosevic, line 657: I don't get what this branch is doing.

rust-spec/lightclient/detection/detection.md
```diff
@@ -276,421 +443,356 @@ This implies that we need a timeout of at least *2 Delta* for remote
...
+             Replace_Secondary(root);
+         }
+    }
+    return Evidences; 
+}
+```
 
 - Expected precondition
-    - *FullNodes* is nonempty
+    - root and primary trace satisfy are a verification trace
```
milosevic, line 669:
```suggestion
    - root and primary trace are a verification trace
```

rust-spec/lightclient/detection/detection.md
josef-widder: No. I also will use that to define a verification trace which does not necessarily mean that the b is from the chain. This will come later when `supports` is used in other definitions.

rust-spec/lightclient/detection/detection.md
josef-widder: For historic reasons ;-). 
I guess `skip-trace` would be a better name. I will update that.

rust-spec/lightclient/detection/detection.md
josef-widder: Yes.  I will add a comment to make this clear.

rust-spec/lightclient/detection/detection.md
josef-widder: I don't see what the "optimization" here would be. From a semantic viewpoint, currently the evidence says:  
"Some of the **next validators** in v(i).Height messed up v(i+1)"

If we would change that the semantics it would say:
"Some of the next validators in v(i).Height messed up v(i+1), unless v(i).Height = v(i+1).Height in which case the **validators** in v(i).Height messed up v(i+1)"

My feeling is that it is more natural to treat the special case "v(i).Height + 1= p v(i+1).Height" on the receiving side of the evidence rather on the sending side, and check this condition there.

Also, we do not check at the light client what kind of attack we have  (we decided to not add this information in the evidence). So it is not clear how to figure out that case.

@cmwaters : do I miss something here?

rust-spec/lightclient/detection/detection.md
josef-widder: Sorry. I changed the control flow last second. I will fix that right now. We do nothing here. This is the case where the secondary agrees with the primary.

rust-spec/lightclient/detection/detection.md
```diff
@@ -276,421 +443,356 @@ This implies that we need a timeout of at least *2 Delta* for remote
...
+            // help to call replace_primary. Also we will use the
+            // same primary to check with other secondaries in
+            // later iterations of the loop
+         }
+         else {
+             // first the secondary reported a faulty block but
+             // then reported a matching block for the same height
+             // first block was most likely bogus
+             Replace_Secondary(root);
+         }
```
josef-widder, line 662:
```suggestion
         // In the case where the secondary reports NoEvidence
         // we do nothing
```

rust-spec/lightclient/detection/detection.md
josef-widder: We don't have the files on master yet. I will fix that then.

rust-spec/lightclient/detection/detection.md
josef-widder: fixed

rust-spec/lightclient/detection/detection.md
josef-widder: added

rust-spec/lightclient/detection/detection.md
```diff
@@ -96,148 +134,273 @@ there exist light blocks a(i) s.t.
...
-> F4 and F5 (phantom validators, lunatic) cannot lead to a fork on the
-> chain but to a light client
-> fork if *t+1 < f < 2t+1*.  
-> F4 and F5 can also lead to bogus blocks
-
 ### Informal Problem statement
 
-> We put tags to informal problem statements as there is no sequential
-> specification.
+There is no sequential specification: the detector only makes sense
```
cezarad, line 344: There is a sequential specification using a global state, that contains a set of light blocks. You can use sing and support to define input output relations, i.e., a list, resp. a tree (see our discussion):).

rust-spec/lightclient/detection/detection.md
```diff
@@ -276,421 +447,352 @@ This implies that we need a timeout of at least *2 Delta* for remote
...
+    no evidence was found before that happened
 
-### From the verifier
+----
 
 ```go
-func VerifyToTarget(primary PeerID, lightStore LightStore,
-                    targetHeight Height) (LightStore, Result)
+func CreateEvidenceForPeer(peer PeerID, root LightBlock, trace LightStore)
+                          (Evidence, LightStore, result) {
```
milosevic, line 682: I think we are missing one return value as it returns 4 values.

rust-spec/lightclient/detection/detection.md
```diff
@@ -276,421 +447,352 @@ This implies that we need a timeout of at least *2 Delta* for remote
...
+                ev.Evidence.ConflictingBlock := trace[i];
+                ev.Evidence.CommonHeight := common.Height;
+                ev.Peer := peer
+                return (ev, common, auxLS, FoundEvidence)
+            }
+            else {
+                // the peer agrees with the trace, we move common forward
+                // we could delete auxLS as it will be overwritten in
+                // the next iteration
+                common := trace[i].Header
```
milosevic, line 709:
```suggestion
                common := trace[i]
```

milosevic: I think we are passing LightBlock to VerifyToTarge, not header.

rust-spec/lightclient/detection/detection.md
```diff
@@ -276,421 +447,352 @@ This implies that we need a timeout of at least *2 Delta* for remote
...
+    }
+    return (nil, nil, nil, NoEvidence)
+}
 ```
+- Expected precondition
+    - root and trace are a verification trace
+- Expected postcondition
+    - finds evidence where trace and peer diverge
+- Error condition
+    - `ErrorTrustExpired`: fails if root expires (outside trusting
```
milosevic, line 721: Should we return error obtain from VerifyToTarget so this is true?

milosevic (approved):
Great work!LGTM.

rust-spec/lightclient/detection/detection.md
josef-widder: I added a clarification

rust-spec/lightclient/detection/detection.md
josef-widder: yes, I added the LightBlock at the second position

PR #165, opened 2020-09-22
evidence: update data structures

cmwaters:
Updated the data structures and other areas of the spec to reflect the go implementation.

Question: what do we want to do about the evidence reactor part in the spec. At the moment, it is almost blank

### DIFF
--- a/spec/core/data_structures.md
+++ b/spec/core/data_structures.md
@@ -4,14 +4,14 @@ Here we describe the data structures in the Tendermint blockchain and the rules 
 
 The Tendermint blockchains consists of a short list of basic data types:
 
-- `Block`
-- `Header`
-- `Version`
-- `BlockID`
-- `Time`
-- `Data` (for transactions)
-- `Commit` and `Vote`
-- `EvidenceData` and `Evidence`
+-   `Block`
+-   `Header`
+-   `Version`
+-   `BlockID`
+-   `Time`
+-   `Data` (for transactions)
+-   `Commit` and `Vote`
+-   `EvidenceData` and `Evidence`
 
 ## Block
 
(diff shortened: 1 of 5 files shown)


marbar3778:
could you fix the linter please

marbar3778 (approved):
lgtm

spec/core/data_structures.md
```diff
@@ -231,24 +226,46 @@ in the same round of the same height. Votes are lexicographically sorted on `Blo
...
+Valid Light Client Attack Evidence encompasses three types of attack and must adhere to the following rules
+
+- If the header of the light block is invalid, thus indicating a lunatic attack, the node must check that
+    they can use `verifySkipping` from their header at the common height to the conflicting header
+
+- If the header is valid, then the validator sets are the same and this is either a form of equivocation
+    or amnesia. We therefore check that 2/3 of the validator set also signed the conflicting header
+
+- The trusted header of the node at the same height as the conflicting header must have a different hash to
+    the conflicting header.
```
josef-widder, line 266:
```suggestion
Valid Light Client Attack Evidence encompasses three types of attack and must adhere to the following rules

- The trusted header of the node at the same height as the conflicting header must have a different hash to the conflicting header.

- CommonHeight < ConflictingBlock.Header.Height 
    
- If CommonHeight + 1 = ConflictingBlock.Header.Height we do sequential verification (we check that 2/3 of the validator set also signed the conflicting header)

- If CommonHeight + 1 < ConflictingBlock.Header.Height we check `verifySkipping` 
```

I am not sure wether validity is the point around we should do the case distinction. I guess nodes may also do amnesia for invalid blocks?

josef-widder:
Ups. I wrote my comment while it was merged...

spec/core/data_structures.md
cmwaters: The actual implementation does in fact use the common height as the case distinction

PR #159, opened 2020-09-09
Light client supervisor

josef-widder:
This is the first draft for the lightclient supervisor. It addresses (partly)

the new lightstore semantics https://github.com/informalsystems/tendermint-rs/issues/499,
the proposed changes to the parameter handling from https://github.com/informalsystems/tendermint-rs/pull/509,
initialization from Genesis #131
In the spec, we call a function AttackDetector which will eventually implement the latest ideas to compute evidence to address https://github.com/informalsystems/tendermint-rs/issues/461 (although proof of fork will only survive for the IBC use case).

There are still many TODOs. I hope it is understandable nonetheless. I would prefer to start to work on the AttackDetector Spec now, before polishing this one.

PR #171, opened 2020-09-25
spec: update light client verification to match supervisor

josef-widder:
Due to several discussion, I did some small changes to the verification spec:

- `VerifyToTarget` and `Backwards` are called with a single lightblock as root of trust in contrast to passing the complete lightstore. This is to make it consistent with the proposed changes to the light client supervisor:
https://github.com/tendermint/spec/pull/159. 
https://github.com/informalsystems/tendermint-rs/pull/509

- Lightblocks in a lightstore only have a state internal to verification (StateVerified, StateTrusted, etc. are not visible externally). The lightstore at the supervisor level should only contain verified (and cross-checked) lightblocks. So we explicitly shoud dispose lightblocks that were not verified during verification (because bisection jumped over them). 
https://github.com/informalsystems/tendermint-rs/issues/499
  
- During verification, we record for each lightblock which other lightblock can be used to verify it in one step. This is needed to generate verification traces that are needed for IBC.  
https://github.com/informalsystems/tendermint-rs/issues/499. 
https://github.com/informalsystems/tendermint-rs/issues/497

As these changes do not change the verification itself (but do some clean-up, and record some auxiliary information for later use), I don't think we need to change the TLA+ specs.


rust-spec/lightclient/verification/verification_002_draft.md
```diff
@@ -0,0 +1,1045 @@
...
+- Expected precondition:
+    - *untrusted* is valid, that is, satisfies the soundness [checks][block]
+    - *untrusted* is **well-formed**, that is,
+        - *untrusted.Header.Time < now + clockDrift*
+        - *untrusted.Validators = hash(untrusted.Header.Validators)*
+        - *untrusted.NextValidators = hash(untrusted.Header.NextValidators)*
+    - *trusted.Header.Time > now - trustingPeriod*
+    - *trusted.Commit* is a commit for the header
+     *trusted.Header*, i.e., it contains
+     the correct hash of the header, and +2/3 of signatures
```
josef-widder, line 840:
```suggestion
```

rust-spec/lightclient/verification/verification_002_draft.md
```diff
@@ -0,0 +1,1042 @@
...
+     of the blockchain of height *targetHeight*
+     (that is, the LightBlock has been added to *lightStore*) **[LCV-POST-LS.1]**
+- Error conditions
+    - if the precondition is violated
+    - if `ValidAndVerified` or `FetchLightBlock` report an error
+    - if [**[LCV-INV-TP.1]**](#LCV-INV-TP.1) is violated
+  
+### Details of the Functions
+
+#### **[LCV-FUNC-VALID.1]**
```
josef-widder, line 825:
```suggestion
#### **[LCV-FUNC-VALID.2]**
```

rust-spec/lightclient/verification/verification_002_draft.md
```diff
@@ -0,0 +1,1061 @@
...
+      the LightStore does not contain such an *lb*.
+---
+
+#### **[LCV-FUNC-LOWEST.2]**
+
+```go
+func (ls LightStore) Lowest() (LightBlock)
+```
+- Expected postcondition
+    - returns the lowest light block within trusting period
```
josef-widder, line 969:
```suggestion
    - returns the lowest trusted light block within trusting period
```

cezarad (approved):

PR #173, opened 2020-09-29
TLA+ light client verification specs from MBT revision

josef-widder:
I did not rename (and thus put version numbers) on the old TLA+ specs, as this requires to update all the MC* files. Not sure we want to do this now in that we. @konnov, what do you think?


konnov (approved):
The updates look good to me. I think we can use the new version in the MC.*, once we model checked the new version. Then its version will also increment. Let's merge as it is and create an issue on model checking.

josef-widder:
I will create an issue

PR #176, opened 2020-10-01
Core: move validation & data structures together

### LINKED ISSUE #128: Combine data structure descriptions and validation rules in core/data_structures.md
Currently we have a section for all the data structures and a section for all validation rules.

Instead, we should probably have the validation rules right next to the data structures.

### LINKED ISSUE #182: evidence: missing lightblock type
`lightclientAttack` evidence references `lightBlock`: https://docs.tendermint.com/master/spec/core/data_structures.html#lightclientattackevidence but is not defined in the spec. 

## Actionable

Add `lightblock` to core data structure document

marbar3778:
- Move validation of data structures next to the data structure
- changes data structures to be agnostic, the proto files will be moved to the spec repo shortly and referenced as well. 

closes #182 
closes #128 

[Rendered](https://github.com/tendermint/spec/blob/edaa726b059c65a1f27094376c0fe55667def700/spec/core/data_structures.md)


spec/core/data_structures.md
```diff
@@ -4,195 +4,243 @@ Here we describe the data structures in the Tendermint blockchain and the rules
...
+|------------|--------------------------------|--------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------|--------------------------------------------------------------------------------------------------|
+| Header     | [Header](#header)              | Header corresponding to the block. This field contains information used throughout consensus and other areas of the protocol. To find out what it contains, visit [header] (#header) | Must adhere to the validation rules of [header](#header)                                         |
+| Data       | [Data](#data)                  | Data contains a list of transactions. The contents of the transaction is unknown to Tendermint.                                                                                      | This field can be empty or populated, but the data is unknown to Tendermint                      |
```
cmwaters, line 26: You sort of mention that data is unknown to Tendermint twice. On another note we could add that validation of txs is reliant on the application i.e `CheckTx()`

spec/core/data_structures.md
```diff
@@ -4,195 +4,243 @@ Here we describe the data structures in the Tendermint blockchain and the rules
...
+| Evidence   | [EvidenceData](#evidence_data) | Evidence contains a list of infractions committed by validators.                                                                                                                     | Can be empty, but when populated the validations rules from [evidenceData](#evidence_data) apply |
+| LastCommit | [Commit](#commit)              | `LastCommit` includes one vote for every current validator.  All votes must either be for the previous block, nil or absent. All votes must have a valid signature from the corresponding validator.  The sum total of the voting power of the validators that voted must be greater than 2/3 of the total voting power of the complete validator set. The number of votes in a commit is limited to 10000 (see `types.MaxVotesCount`).                                                                                             | Can be empty for the initial height and must adhere to the validation rules of [commit](#commit).  |
```
cmwaters, line 28: Doesn't a commit have to be empty for the initial height? Consider changing "Can" to "Must"

spec/core/data_structures.md
```diff
@@ -4,195 +4,243 @@ Here we describe the data structures in the Tendermint blockchain and the rules
...
+| Time              | [Time](#time)       |                   The timestamp is equal to the weighted median of honest validators. Read more on time in the [BFT-time section](../consensus/bft-time.md). Note: the timestamp of a vote must be greater by at least one millisecond than that of the block being voted on.                                                                                                                                                                                                                                                                                                                                                                    | Time must be >= previous header timestamp + consensus parameters TimeIotaMs.  The timestamp of the first block must be equal to the genesis time (since there's no votes to compute the median).  |
```
cmwaters, line 81:
```
The timestamp is equal to the weighted median of honest validators
```
should be
```
The timestamp is equal to the weighted median of vote timestamps present in the last commit
```

spec/core/data_structures.md
```diff
@@ -4,195 +4,243 @@ Here we describe the data structures in the Tendermint blockchain and the rules
...
 
 `CommitSig` represents a signature of a validator, who has voted either for nil,
 a particular `BlockID` or was absent. It's a part of the `Commit` and can be used
 to reconstruct the vote set given the validator set.
 
-```go
-type BlockIDFlag byte
+| Name             | Type                        | Description                                                                                                  | Validation                                                        |
+|------------------|-----------------------------|---------------------------------------------------------------------------------------------------------------|-------------------------------------------------------------------|
+| BlockIDFlag      | [BlockIDFlag](#blockidflag) | Represents the way the validator voted, if the validator missed the voting period an absent flag will be set. | Must be one of the fields in the [BlockIDFlag](#blockidflag) enum |
```
cmwaters, line 150: Suggestion:
```
Represents the way the validator voted, if the validator missed the voting period an absent flag will be set.
```
To
```
Represents the validators participation in consensus: Either voted for the block that received the majority, voted for another block, voted nil or did not vote.
```

spec/core/data_structures.md
```diff
@@ -4,195 +4,243 @@ Here we describe the data structures in the Tendermint blockchain and the rules
...
- BlockIDFlag      BlockIDFlag
- ValidatorAddress Address
- Timestamp        time.Time
- Signature        []byte
+BlockIDFlag represents which BlockID the [signature](#commitsig) is for.
+
+```go
+enum BlockIDFlag {
+  BLOCK_ID_FLAG_UNKNOWN = 0;
+  BLOCK_ID_FLAG_ABSENT  = 1;
```
cmwaters, line 165:
```suggestion
  BLOCK_ID_FLAG_ABSENT  = 1; // signatures for other blocks are also considered absent
```

spec/core/data_structures.md
```diff
@@ -4,195 +4,243 @@ Here we describe the data structures in the Tendermint blockchain and the rules
...
-
 ## Vote
 
 A vote is a signed message from a validator for a particular block.
-The vote includes information about the validator signing it.
+The vote includes information about the validator signing it. When stored in the blockchain or propagated over the network, votes are encoded in Protobuf.
+
+| Name             | Type                            | Description                                                                                 | Validation                                                                                           |
+|------------------|---------------------------------|---------------------------------------------------------------------------------------------|------------------------------------------------------------------------------------------------------|
+| Type             | [SignedMsgType](#signedmsgtype) | [SignedMsgType](#signedmsgtype)                                                             | A Vote is valid if its corresponding fields are included in the enum [signedMsgType](#signedmsgtype) |
```
cmwaters, line 178: I would suggest changing the description to:

```
Either prevote or precommit
```
and removing the bottom part ... "There are two types of votes"

cmwaters: Also do we want to add the data struct of the Proposal

cmwaters (commented):
Still not quite finished looking though it all but this is excellent work :+1:

spec/core/data_structures.md
marbar3778:
```suggestion
| BlockIDFlag      | [BlockIDFlag](#blockidflag) | Represents the validators participation in consensus: Either voted for the block that received the majority, voted for another block, voted nil or did not vote | Must be one of the fields in the [BlockIDFlag](#blockidflag) enum |
```

spec/core/data_structures.md
```diff
@@ -278,297 +312,49 @@ Valid Light Client Attack Evidence encompasses three types of attack and must ad
...
-
-  val := state.LastValidators[i]
-  vote.Verify(block.ChainID, val.PubKey) == true
-
-  talliedVotingPower += val.VotingPower
-}
-
-talliedVotingPower > (2/3)*TotalVotingPower(state.LastValidators)
-```
+## LightBlock
```
cmwaters, line 315: Maybe a short note on what it is and why we have it i.e. LightBlock is the core data structure used by light clients.

spec/core/data_structures.md
```diff
@@ -278,297 +312,49 @@ Valid Light Client Attack Evidence encompasses three types of attack and must ad
...
-```go
-state(initialHeight) = InitialState
-state(h+1) <- Execute(state(h), ABCIApp, block(h))
-```
+## Address
 
-where `InitialState` includes the initial consensus parameters and validator set,
-and `ABCIApp` is an ABCI application that can return results and changes to the validator
-set (TODO). Execute is defined as:
+Address is a type alias of a slice of bytes. The address is calculated by hashing the bytes using sha256 and truncating it to only use the first 20 bytes of the slice.
```
cmwaters, line 349: >The address is calculated by hashing the bytes

Hashing the bytes of what? The pub key?

spec/core/data_structures.md
marbar3778: This is true, but the block validation does not rely on checkTx. I can add a minor reference

spec/core/data_structures.md
marbar3778: Yes, there are few things missing. I didn't want bloat this PR, is it okay if I do a followup with more changes?

melekes (approved):
great work 🔥

spec/core/data_structures.md
```diff
@@ -4,195 +4,239 @@ Here we describe the data structures in the Tendermint blockchain and the rules
...
+| LastResultHash    | slice of bytes (`[]byte`)     | `LastResultsHash` is the root hash of a Merkle tree built from `ResponseDeliverTx` responses (`Log`,`Info`, `Codespace` and `Events` fields are ignored). The first block has `block.Header.ResultsHash == MerkleRoot(nil)`, i.e. the hash of an empty input, for RFC-6962 conformance.                                                                                               | Must  be of length 32                                                                                                                                                                                                  |
```
josef-widder, line 89: compared to the other fields, the description reads much more technical and detailed, and less insightful at the same time.

spec/core/data_structures.md
```diff
@@ -4,195 +4,239 @@ Here we describe the data structures in the Tendermint blockchain and the rules
...
+
+| Name        | Type                        | Description                                                                                                                                                      | Validation                                                         |
+|-------------|-----------------------------|------------------------------------------------------------------------------------------------------------------------------------------------------------------|--------------------------------------------------------------------|
+| Hash        | slice of bytes (`[]byte`)   | MerkleRoot of all the fields in the header (ie. `MerkleRoot(header)`.                                                                                            | hash must be of length 32                                          |
+| PartsHeader | [PartsHeader](#PartsHeader) | Used for secure gossiping of the block during consensus, is the MerkleRoot of the complete serialized block cut into parts (ie. `MerkleRoot(MakeParts(block))`). | Must adhere to the validation rules of [PartsHeader](#PartsHeader) |
```
josef-widder, line 107: Should this be partsheader or partsetheader?

spec/core/data_structures.md
```diff
@@ -4,195 +4,239 @@ Here we describe the data structures in the Tendermint blockchain and the rules
...
+| BlockIDFlag      | [BlockIDFlag](#blockidflag) | Represents the validators participation in consensus: Either voted for the block that received the majority, voted for another block, voted nil or did not vote | Must be one of the fields in the [BlockIDFlag](#blockidflag) enum |
+| ValidatorAddress | [Address](#address)         | Address of the validator                                                                                      | Must be of length 20                                              |
+| Timestamp        | [Time](#time)               | This field will vary from `CommitSig` to `CommitSig`. It represents the timestamp of the validator.               | [Time](#time)                                                     |
+| Signature        | [Signature](#signature)     | Signature corresponding to the validators participation in consensus.                                         | The length of the signature must be > 0 and < than  64            |
```
josef-widder, line 153: This point was never spelled out in the old spec. We should make it more clear. Somehow, 
- a signature, 
- a vote
- height, round, blockID

belong together and there is a check to ascertain that. Somehow, I must be able to check that a signature belongs to a certain height, round, blockID. I never saw that discussed anywhere

spec/core/data_structures.md
```diff
@@ -4,195 +4,239 @@ Here we describe the data structures in the Tendermint blockchain and the rules
...
+
+  // Proposal
+  SIGNED_MSG_TYPE_PROPOSAL = 32;
 }
 ```
 
-There are two types of votes:
-a _prevote_ has `vote.Type == 1` and
-a _precommit_ has `vote.Type == 2`.
+## CanonicalVote
```
josef-widder, line 204: I don't understand how this is linked to the rest of the document.

spec/core/data_structures.md
```diff
@@ -258,14 +285,17 @@ Valid Duplicate Vote Evidence must adhere to the following rules:
```
josef-widder, line 285: "Evidence must not have expired" does not really belong here. Or it should be explained differently: The first 4 bullet points are independent of time, that is, I can still validate them in two years from now. The last one depends on time, and is not a validating rule, but rather a guarantee of Tendermint at the time when Evidence is passed to the application.

spec/core/data_structures.md
```diff
@@ -278,297 +308,53 @@ Valid Light Client Attack Evidence encompasses three types of attack and must ad
 
 - Evidence must not have expired. The height (and thus the time) is taken from the common height.
```
josef-widder, line 310: The issue about being "expired" is the same as above. I am not sure it belongs to a document about data structures.  
Also, all the other rules implicitly assume validation at a full node against a local copy of the blockchain (e.g., "invalid" and "conflicting header" in bullets 1, 2 and 3). This should be made clear.

spec/core/data_structures.md
```diff
@@ -278,297 +308,53 @@ Valid Light Client Attack Evidence encompasses three types of attack and must ad
...
 
-where `pubKey.Verify` performs the appropriate digital signature verification of the `pubKey`
-against the given signature and message bytes.
+## ValidatorSet
 
-## Execution
+| Name       | Type                             | Description                                        | Validation                                                                                                        |
+|------------|----------------------------------|----------------------------------------------------|-------------------------------------------------------------------------------------------------------------------|
+| Validators | Array of [validator](#validator) | List of the active validators at a specific height | The list of validators can not be empty or nil and must adhere to the validation rules of [validator](#validator) |
+| Proposer   | Array of [validator](#validator) | The block proposer for the corresponding block           | The proposer cannot be nil and must adhere to the validation rules of  [validator](#validator)                    |
```
josef-widder, line 336: why is proposer an array? shouldn't there be only one?

spec/core/data_structures.md
marbar3778: I can reword this to be specific for block inclusion.

spec/core/data_structures.md
marbar3778: Canonical Vote is used for signing (privval). It can be seen as implementation specific, but the rust client has implemented it due to its use in tmkms.

josef-widder: I see. My point is that if I read the document from top to bottom, at this point I have no idea what a Canonical Vote is. At the beginning of the document there is a bullet list with data types, but it doesn't show there. Also it is not mentioned. So the role of this data type is just not clearly stated here, I guess.

spec/core/data_structures.md
```diff
@@ -4,195 +4,239 @@ Here we describe the data structures in the Tendermint blockchain and the rules
 
 The Tendermint blockchains consists of a short list of basic data types:
```
josef-widder, line 5:
```suggestion
The Tendermint blockchains consists of a short list of data types:
```
basic data types are typically integer and Boolean...

spec/core/data_structures.md
marbar3778: This addition would be better suited alongside the block. The validation of commitsig just checks the length of the signature.

marbar3778: added in a secondary validation section under the execution section of the block.

cmwaters (approved):
Good stuff :+1:  although I will have to follow up straight after with my changes to the evidence data structures :grimacing:

PR #184, opened 2020-10-08
spec: light client attack detector 002

josef-widder:
Before going into evidence creation, we first cross check the lastest Header, and only go into evidence creation when there is a conflict. (This is part of a revision to address both lightnode and IBC use cases).

I introduced a new version. However, only [LCD-FUNC-DETECTOR.2] is new with the code around `FetchLightBlock`.


cezarad (approved):
looks good.

PR #210, opened 2020-11-04
add Ivy proofs

nano-o:


milosevic (approved):

ivy-proofs/network_shim.ivy
```diff
@@ -0,0 +1,140 @@
...
+
+# Here we define a network module, which is our model of the network, and a
+# shim module that sits on top of the network and which, upon receiving a
+# message, calls the appropriate protocol handler.
+
+include domain_model
+
+# Here we define an enumeration type for identifying the 3 different types of
+# messages that nodes send.
+object msg_kind = { # TODO: merge with step_t
```
konnov, line 15: is it done, or shall we remove the TODO?

ivy-proofs/network_shim.ivy
```diff
@@ -0,0 +1,140 @@
...
+
+        # The state of the network consists of all the packets that have been
+        # sent so far, along with their destination.
+        relation sent(V:msg, N:node)
+
+        after init {
+            sent(V, N) := false
+        }
+
+        before send {
```
konnov, line 53: should it be "after"?

konnov (commented):
The PR contains all the details we have been discussing and that you presented to us a few ago. We just have two small comments. Feel free to resolve them at your discretion. The PR has been merged.

ivy-proofs/network_shim.ivy
nano-o: With `before`, the code is inserted at the beginning of the action. With `after`, it would be inserted at the end of the action. Since the action is atomic, it doesn't change much. It could make a difference if some assertions inside the action depended on the the specification state being updated or not, but I think it's not the case here.

ivy-proofs/network_shim.ivy
nano-o: This can be removed.

PR #222, opened 2020-11-16
RFC: Soft Chain Upgrades

cmwaters:
This is a RFC for the proposal of adding support for soft upgrades, allowing for the lifespan of a chain to be across multiple block versions with upgrading happening whilst the network is still live. 

[Rendered](https://github.com/tendermint/spec/blob/callum/rfc5/rfc/005-soft-chain-upgrades.md)


rfc/005-soft-chain-upgrades.md
```diff
@@ -0,0 +1,227 @@
...
+
+The block version consists of all block related data structures i.e. Header,
+Vote, Validator, Commit. This would be the sole focus for soft upgrades (a full
+list of the exact data structures can be found in
+[data structures](https://github.com/tendermint/spec/blob/master/spec/core/data_structures.md#data-structures).
+A block version could be incremented between minor releases. However, for
+backwards compatibility, all prior versions in a given major release would also
+need to be supported thus TM 1.3 could support block version 9 and 10 and TM 1.4
+could introduce block version 11 but would still need to support 9 and 10. In a
+major release the support of prior versions could be dropped. Continuing with
```
marbar3778, line 59: Could prior versions be kept around? I find this useful if we were to offer RPC querying across versions.

cmwaters: Prior major versions you mean?

marbar3778: prior block versions. If a chain started on block version 9 (TM 1.3) and is now on block version 12 (TM 2.3) as a user I may want to query a block from version 9.

cmwaters: Across major versions i.e. 1.x -> 2.x would still be a hard fork and thus require a new chain (as is the case currently). I would also like to find ways to make hard forks easier on networks but I want to address this in a separate discussion

marbar3778: > Across major versions i.e. 1.x -> 2.x would still be a hard fork and thus require a new chain (as is the case currently).

For p2p upgrades only right? It seems to me this upgrade process would mitigate hard forks for everything minus p2p upgrades.

rfc/005-soft-chain-upgrades.md
```diff
@@ -0,0 +1,227 @@
...
+
+### Positive
+
+- Chains get the benefits of upgrades without having to sacrifice their block
+history
+
+### Negative
+
+- There is a large degree of complexity required in ensuring the flexibility of
+data structures. This could lead to the bloating of the code-base and/or
```
marbar3778, line 217: Putting different version of Tendermint into a tm-version repo could mitigate this. This way tendermint only holds the latest versions and avoids bloating. 

There is some added complexity to this but keeps the repo clean and minimal.

cmwaters: and just import it as a library? Yeah that could be an alternative implementation

marbar3778: > and just import it as a library? Yeah that could be an alternative implementation

Yea!

rfc/005-soft-chain-upgrades.md
```diff
@@ -0,0 +1,227 @@
...
+live and at the discretion of the app.
+
+Depending on the implementation it may be more beneficial to divide block
+protocol versions into more specific abstractions i.e. Block, State and
+Consensus to isolate changes from spilling over multiple modules (normally a
+change to say `commit` would cause us to replicate all the data structures).
+However, increasing the granularity of abstractions would compound the
+permutations between versions which could lead to untamable code, hence, given
+that these changes should be grouped together and occur seldomly it should
+suffice to have a single block protocol version.
```
erikgrinaker, line 75: This ties into the planned versioning for the spec as well. I'd say the protocol version should be given by the spec version, and all data structures should follow that.

rfc/005-soft-chain-upgrades.md
```diff
@@ -0,0 +1,227 @@
...
+
+### Transitions
+
+Networks would be expected to announce upgrades in advance allowing a window for
+individual nodes to upgrade to a software version of Tendermint that would
+comply with the announced block protocol upgrade. Node runners would temporarily
+halt their node, change to the latest binary and start up again. Tendermint
```
erikgrinaker, line 82: It might be nice to be explicit about version expectations for nodes in a network. Do we expect/require all nodes in a network to run the same minor version at all times? Patch version?

rfc/005-soft-chain-upgrades.md
```diff
@@ -0,0 +1,227 @@
...
+Furthermore, although work from [RFC002: Non-Zero Genesis](https://github.com/tendermint/spec/blob/master/rfc/002-nonzero-genesis.md)
+has enabled unique heights throughout the history of a "logical" chain, data
+retrievability across heights remains difficult as blockchains that choose to
+upgrade not only have to coordinate upgrades but they essentially lose all their
+history.
+
+The scope of this RFC is to address the need for Tendermint to have flexibility
+in improving itself whilst offering greater ease of use to the networks running
+on Tendermint. This is done by introducing a division in the types of ways a
+network can be upgraded. There will still remain the need for hard upgrades (the
```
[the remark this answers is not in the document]
cmwaters: > Why will there still remain a need for hard upgrades?

Because I can't bind what future tendermint developers do nor predict how tendermint will evolve.

What I've attempted to do is to split the problem by identifying a set of data structures that are likely to change and give them flexibility to change without requiring a "hard break" or a reset of the block chain. At the moment, Tendermint releases a major version every 6 months but this might allow major versions to only happen every couple of years, thus extending the lifetime of a blockchain. But this is still an experimental field, we could decide to completely change up the consensus mechanism and in that case v2 could look completely different from v1. I find it unreasonable to have a single software version that tries to tie all this together. 

You're right though that we should also be addressing when a major change happens how we retain the appearance of a full immutable history that can be accessed and I've been trying to come up with ideas here https://github.com/tendermint/tendermint/issues/5595#issuecomment-724807340 to essentially improve the ux of a "hard upgrade"

rfc/005-soft-chain-upgrades.md
cmwaters: All nodes in a network must be able to run the same protocol version (p2p, block and app). They can, however, be running different minor versions so long as that minor version supports the protocol versions

rfc/005-soft-chain-upgrades.md
cmwaters: could you import from the same repo but from different releases?

marbar3778: Potentially. Once we make the jump to proper semvar then we will want to look at how go modules recommends versioning: https://blog.golang.org/v2-go-modules. 

```md
Major versions and module paths

Modules formalized an important principle in Go, the import compatibility rule:

If an old package and a new package have the same import path,
the new package must be backwards compatible with the old package.

By definition, a new major version of a package is not backwards compatible with the previous version. This means a new major version of a module must have a different module path than the previous version.
```

Although this section conflicts with allowing block protocol breaking in an X.X release.

rfc/005-soft-chain-upgrades.md
```diff
@@ -0,0 +1,227 @@
...
+Tendermint currently has three protocol versions: Block, P2P and App.
+
+The block version consists of all block related data structures i.e. Header,
+Vote, Validator, Commit. This would be the sole focus for soft upgrades (a full
+list of the exact data structures can be found in
+[data structures](https://github.com/tendermint/spec/blob/master/spec/core/data_structures.md#data-structures).
+A block version could be incremented between minor releases. However, for
+backwards compatibility, all prior versions in a given major release would also
+need to be supported thus TM 1.3 could support block version 9 and 10 and TM 1.4
+could introduce block version 11 but would still need to support 9 and 10. In a
```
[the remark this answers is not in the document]
cmwaters: Block version changes are associated with heights. Is it still breaking if a client request a block at a height 20 with version 1 and the block at that height is actually version 2 so the rpc just returns a version error. This basically means that clients still work but they have an invisible ceiling of heights they can query until they upgrade.

> Make sure the data structures of all new block versions can be represented as all prior supported versions, and that clients will have to opt-in to the new data structures by bumping e.g. the RPC version number of requests.

How feasible do you think this is? Isn't this much the same as just making it backwards-compatible like you suggested by having sermver for block versions.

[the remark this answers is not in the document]
erikgrinaker: > Is it still breaking if a client request a block at a height 20 with version 1 and the block at that height is actually version 2 so the rpc just returns a version error.

Yes, I think so. This will break e.g. block explorers and coin exchanges, and basically all clients that interact with Tendermint that haven't upgraded.

>How feasible do you think this is? Isn't this much the same as just making it backwards-compatible like you suggested by having sermver for block versions.

Depends on what we need. I don't think it buys us much over just "regular" backwards-compatibility, but it might reduce the amount of cruft in newer block versions. E.g. if we rename a field, with backwards compatibility we would have to keep both the old and the new field, while with the other approach we could just map the new field onto the old field for clients that request a previous version and not have the old field present in new versions.

rfc/005-soft-chain-upgrades.md
erikgrinaker: So this means that if we upgrade Tendermint 1.4 -> 1.5, which bumps block version 10 -> 11, then all nodes in the network must upgrade to 1.5 at the same time?

cmwaters: When tendermint releases 1.5, nodes can upgrade whenever they want but they will continue to run block version 10 until the entire network decides to move from 10 to 11 (this is usually done through governance proposals) at a specified height. From that height onwards nodes will switch to block version 11 unless they haven't upgraded in time in which they will error out

rfc/005-soft-chain-upgrades.md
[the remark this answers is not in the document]
erikgrinaker: > I find it unreasonable to have a single software version that tries to tie all this together.

I sort of disagree with this. It depends on what kind of product we want to offer out users, but people don't have to throw their data away when upgrading e.g. Postgres 8 to 13, and I think we should be able to do the same. It doesn't have to be convenient (e.g. a migration process is totally fine), but I think we should offer some way for people to keep their data across upgrades.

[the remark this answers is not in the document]
cmwaters: yeah I did like the idea of turning nodes that were used in older versions into archive nodes which only run say the blockchain reactor and an rpc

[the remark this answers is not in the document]
cmwaters: basically during a major version change, the old node would pass the "baton" i.e. config + genesis to the new node to start up, but it wouldn't quite be left to throw away. Instead it would become an archive node serving data from that "era"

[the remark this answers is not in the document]
erikgrinaker: Yeah, that might work. Problem is that this requires coordination with all users, because they now have to know which heights are located at which servers, and at what height a future upgrade is taking place. If this is the approach we're planning to use, then I feel like we should have some sort of built-in frontend proxy that automatically does this mapping, such that when a network upgrades the governors can make sure all of these proxies have been updated with new network information and relevant block ranges before the upgrade takes place. This way, users/clients don't have to do anything.

Of course, this doesn't solve the problem of having to support old RPC versions for new blocks, since this will still require clients to upgrade or suddenly break. Maybe that's okay once in a while, it depends on the guarantees we want to offer users. But worth thinking through how we want this to work.

rfc/005-soft-chain-upgrades.md
erikgrinaker: Ah right, that's great. 👍

cmwaters: Yeah I will try fit an example in as well. That might help explain things a bit more clearly.

rfc/005-soft-chain-upgrades.md
```diff
@@ -0,0 +1,227 @@
...
+In this case, Block which falls under a data structure of the block protocol
+could have many versions and is represented here as an interface. `State` and
+`BlockExecutor` are structs outside of the block protocol. Breaking changes to
+them could only occur in major releases.
+
+### Alternative Approach: Migration Scripts
+
+Although using migration scripts would have been most likely a simpler solution,
+unfortunately in most cases this would result in breaking the hashes and is
+therefore not a viable approach.
```
erikgrinaker, line 190: Hm, I wonder if it might be possible to come up with something like a "derived" block, where we create a new version of a given height and somehow include the hash of the original block and the previous derived block. If the first block after the upgrade ties into this "derived" block rather than the original block, and we have the entire network come to consensus on it (and thus the entire history of derived blocks) we may get away with doing migrations, which I think are a much simpler solution to this problem.

Someone who knows more about this stuff than me should vet the idea, but maybe something worth exploring?

rfc/005-soft-chain-upgrades.md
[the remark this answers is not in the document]
alexanderbez: Can we enforce that all block version changes are always backwards compatible? i.e. we would never change an existing field from say v10 -> v11, only add fields. I do agree using semver in block versioning as well could be very useful.

rfc/005-soft-chain-upgrades.md
cmwaters: So how would fast sync work in this case?

erikgrinaker: It could fetch either the original or derived blocks, they both contain the transactions and related metadata, and are verifiable.

marbar3778: This is an interesting idea!! One concern may be for the users that are storing blocks in a DB. They will need to rewrite the entire database. It could be considered a minor hiccup in the larger picture.

cmwaters: So a node keeps both the original and derived block?

erikgrinaker: > One concern may be for the users that are storing blocks in a DB. They will need to rewrite the entire database.

Yes, that's the tradeoff with migrations - they take time and disk space. But this is (mostly) a one-time cost, and I think the reduced complexity and operational burden is worth it.

> So a node keeps both the original and derived block?

That could be configurable, but some nodes in the network should keep the originals as well, for auditability. Keeping the original would also allow clients to request the block both in the original version and in the derived version, which would satisfy backwards compatibility with current clients.

cmwaters: would this get compounded over multiple upgrades?

erikgrinaker: >would this get compounded over multiple upgrades?

Depends on the design. The simplest solution would be to allow migrating from a derived block of the previous block version to a derived block of the current block version - this would only require us to support a n to n+1 version migration in the codebase, but nodes would have to rewrite the entire history for each upgrade.

cmwaters: I'm just trying to wrap my brain around this. So in my head migration scripts wouldn't work because if for example we had a block at v1 and we wanted to migrate it to v2 where we decided to remove the evidence hash from the header, then when a new node was to fast sync, the signatures for the commit at that height would be aimed at a different block ID from the one that the new node would compute. How does this solve it?

erikgrinaker: As I said, someone who knows more about this than me needs to vet this. But the derived block needs to contain some necessary data from the original block, such as the original block ID. And historical signatures will implicitly have been verified and signed by the current validators when they commit the block after the upgrade.

cmwaters: and as well as containing the necessary data from the original block, should it also be able to form the new one?

I will mull this over a bit and see where it leads me. I guess my main concern is to remove the possibility that one node can fool a new node that is syncing up

rfc/005-soft-chain-upgrades.md
```diff
@@ -0,0 +1,227 @@
...
+Networks would be expected to announce upgrades in advance allowing a window for
+individual nodes to upgrade to a software version of Tendermint that would
+comply with the announced block protocol upgrade. Node runners would temporarily
+halt their node, change to the latest binary and start up again. Tendermint
+could also offer a tool to easily swap between binaries thus reducing the amount
+of down-time that nodes experience although this is outside the scope of this
+RFC.
+
+The actual switch would occur in the same manner as an App Version change, using
+the `EndBlockResponse{}` ABCI call to indicate to Tendermint that a different
```
melekes, line 88: don't understand why we need to do that. since tendermint will need to be restarted, the new version will be announced to the app via `Info` ABCI method.

melekes (commented):
this is a good start although a lot of details are still missing.

question: if a node A is on block protocol v1.4 and node B is on v1.3 and node A sends a block proposal to node B, how would node B go about handling this? Will it just ignore it since it can't read it yet (because it doesn't yet downloaded a new version)?

rfc/005-soft-chain-upgrades.md
cmwaters: A network can have nodes running different versions at the same time i.e. 1.3 and 1.4 so long as all the different software versions support the agreed upon block protocol which should be deterministically decided by the app. 

As an example: 

When tendermint releases 1.5, nodes can upgrade whenever they want but they will continue to run block version 10 until the entire network decides to move from 10 to 11 (this is usually done through governance proposals) at a specified height. From that height onwards nodes will switch to block version 11 unless they haven't upgraded in time in which they will error out

rfc/005-soft-chain-upgrades.md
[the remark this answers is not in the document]
cmwaters: Yes I think semver versioning makes sense because I think there will also be times where we want to make changes that are not backwards compatible. This is all a juggle between how much freedom we want to give us as developers and how annoying we are to users. 

Quick question: Is adding fields backwards compatible?  If I were to add callums hash to the header in v11 and an external client were to request a block using v10 then this very important hash would be omitted. Is this acceptable?

rfc/005-soft-chain-upgrades.md
cmwaters: Yeah that makes sense

rfc/005-soft-chain-upgrades.md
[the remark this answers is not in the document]
cmwaters: Yup, I agree that ideally we want nothing to change from the perspective of the external client.

The quasi "brute-force" solution I see is that all nodes import their previous versions. That way the latest version is something of a "master" node and, knowing the upgrade heights, can redirect requests to the corresponding archive node. You could probably get away with just running an extra blockchain reactor and have the same rpc wired to that. I'm just a bit concerned that someone that starts a network at v2.3 has this unnecessarily big binary that he/she uses.

rfc/005-soft-chain-upgrades.md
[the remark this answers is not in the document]
erikgrinaker: > This is all a juggle between how much freedom we want to give us as developers and how annoying we are to users.

Yup. It's worth bearing in mind that software exists for its users, so they should take priority.

> Is adding fields backwards compatible?

Generally yes, with a couple of important caveats:

* The field must be optional, and the default value must preserve existing behavior.

* The field cannot affect other fields - in this context, notably, it cannot be included in any existing hashes, since clients will no longer be able to verify those hashes (unless it's very clearly specified and implemented up-front how new fields should be hashed, e.g. by using all Protobuf fields in ascending order).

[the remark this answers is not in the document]
cmwaters: But in the context of the block protocol where almost everything is connected and nothing is really optional, I'd imagine it is quite difficult to add a field that is backwards compatible.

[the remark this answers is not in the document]
erikgrinaker: No-one said this would be easy. :)

cmwaters:
I have been thinking about chain migration and have come up with the following...

# Chain Migration

## Imperatives

- All state modifying data from the original block must be preserved
- Light client and fast sync verification models must be unaffected

## Strategy

Migration scripts will need to be tailored to what the exact changes entail but the general notion is that the validator set directly following the upgrade signs not just the previous block but finalizes on the migration of the entire history. 

## Problems

Whenever we change the header (or anything in the block for that matter as this too will mean the hashes in the header will need to change), then we break the signatures in the commit. At least 2/3 of the validators in the set at that height agreed on a blockID, the hash of the original header, which has since changed. This block ID can be transferred across to the derived block but we need some way of verifying that this new block does actually derived from the original one. This is what the validators at the height of the upgrade need to agree on, however, not just for a single block but the entire history. 

## Possible Solutions

This is a random walk through the solution space at what could be done to achieve version upgrading whilst following the aforementioned imperatives.

### Simple Approach

One basic approach is that each derived version should be capable of producing its original i.e. If the chain is at version 4 and a node is fast syncing and is at block 100 which corresponds originally to version 2 then when it receives a derived block it should be able to extract out the original for verification purposes. However, the derived block should also reflect the structure of the latest version.

The shortcomings of this is that we would expect the derived block stored to be larger than normal as it needs to hold the relevant data for both. This would go against a lot of the proposed plans on the horizon which would aim to reduce the overall size of the block. Verification would take a little longer but we wouldn't need to worry about distant future validators signing on the change as it shouldn't be possible to introduce any byzantine behavior (either the derived block can produce the original block or it can't). 

### Advanced Approach

I've been thinking a lot longer about a second approach in which we don't need to be able to rebuild the original block to verify all the way up to the migration height. Currently we have what can be seen as a core verification path. These are a set of fields or features that are essential for Tendermint's verification process. This allows a node to be able to trust a header. The header is then filled with whatever other hashes that allows the user to verify the other components i.e. data, evidence, last commit, consensus, app. This guarantees state replication.

With chain migration, we implicitly create a division in the types of blocks: Those blocks that have been migrated from a previous version and those that are part of the latest version. These blocks of course ideally have the same structure but will most likely need to be treated differently because a migrated block has broken signatures whereas the latest ones don't.

Consensus, albeit for the actual agreement of a version change, remains largely unaffected from migration. However, fast syncing and light clients (which basically use the same forms of verification) would. To quickly recap how verification works we start with a trusted validator set. From this we search for a commit and header that correlate to the height directly above our current state. We rebuild the votes from the commit and verify that the signatures are all for the hash of the header we received by calling `VerifyCommit`:

```go
func (vals *ValidatorSet) VerifyCommit(chainID string, blockID BlockID,
	height int64, commit *Commit) error
```

We can then trust the header and as aforementioned trust the rest of the block. Then tendermint delivers the txs to the application which will in turn update the validator set (with an extra height delay). We use NextValidatorsHash to verify that we have the correct trusted validator set for the next height.

If we were to migrate data to a new block, then we could take the original blockID across. This would mean that we could verify this BlockID but we would not be able to trust the rest of the contents of the header / block which means we wouldn't be able to know whether the state transition was correct or if the new validator set could be trusted with which to continue verification.

A solution to this would be as follows:

We have a block that has a set of unalterable fields. They can't change and are essential to the core verification path.

```go
type Block {
	...
	DerivedBlockHash // derives from the original block that was signed by
									 // the validator set
	NextValidatorHash
	...
	LastBlockID // This refers to the migrated block header not the original
	LastCommit // or anything that houses the signatures
	...
}
```

I'm assuming the following relationship

```go
f(DerivedBlockHash, NextValidatorHash) = Hash of the original header / BlockID
```

Part of the migration script would then be to calculate this `DerivedBlockHash` or if it was the migration of an already migrated block then just to carry it across.

Verification would be as follows:

1. Starting from a trusted validator set (this could be from genesis)
2. Grab the migrated block at the next height or more specifically the signatures, next validator hash and derived block hash
3. Calculate the original header hash that the signatures should be signing with the next validator hash and the derived block hash.
4. Check that the LastBlockID is equal to the hash of the trusted header (no need to check this is height is 1)
5. Verify that 2/3 signed the original header hash by running `VerifyCommit`. If no error then we can trust at least that the original block ID is correct and extend that to trusting that the `NextValidatorHash` is correct.
6. Apply block to state. At the moment we still can't trust that the txs were legitimate.
7. Check that the state's `ValidatorHash` (at height + 1 now) matches the `NextValidatorsHash`. If so it means that although we can't trust state we can't trust the new validator set.
8. Go back to 1 and recur until we reach a point where the `DerivedBlockHash` is nil. This indicates the crossover from the migrated blocks to the original ones.
9. When validating this block we return to the normal verification process. This means that instead of using the derived block we take the signatures in the commit and verify it against the hash of the entire header. This means that we not only the next validator hash but can trust the entire header. This includes LastBlockID.
10. Verify that LastBlockID equates to the hash of the last migrated block. If this is equal then we have essentially verified the entire migration history and we can trust that the state transitions that have been applied are correct.
11. Proceed as normal until the node can transition to consensus.

NOTE: The process outline is deemed as the minimum fields that need to be migrated across. Currently this process means that we can't trust state until we reach the crossover point. However, this shortcoming isn't necessary. If we decide to migrate some form of state/app hash such that this still holds:

```go
f(AppHash + NextValidatorsHash, DerivedBlockHash) = Original Block ID
```

Then we can incrementally verify state transitions. This means we don't have to download the entire migration part of the blockchain before we know that state is valid.

In this regards `DerivedBlockHash` serves as the remnants of the hashes/data in the old header that we don't want to bring across because we have changed it.

In terms of the actual upgrade, when consensus is reached on a block that causes the block version to increment, the next block proposed should not only have the new block structure but the LastBlockID should be that of the migrated block not the original one. This means that before the next block of the new version can be agreed upon everyone participating in consensus must have migrated the entire history in order to generate the "golden hash", the cumulative hash of the migrated history. If this is three years of blocks this migration could take a long time before consensus can continue hence it's probably an important aspect to attempt to make migration asynchronous so nodes already know this golden hash before the upgrade actually happens.

### TL;DR

Using migration scripts aside from a few caveats could actually work.

rfc/005-soft-chain-upgrades.md
cmwaters: Have written this up in regards to doing migrations: https://github.com/tendermint/spec/pull/222#issuecomment-731139303

erikgrinaker:
Thanks for thinking about this @cmwaters! My instinct is the same as yours, I think this seems viable (although the devil is in the details here).

> The shortcomings of this is that we would expect the derived block stored to be larger than normal as it needs to hold the relevant data for both. This would go against a lot of the proposed plans on the horizon which would aim to reduce the overall size of the block.

At scale the transaction data will presumably be much larger than the block header, so adding on an extra header or some extra fields has a negligible cost that I think will be well worth it.

> Currently this process means that we can't trust state until we reach the crossover point. However, this shortcoming isn't necessary. If we decide to migrate some form of state/app hash.

I think we probably don't want to migrate transactions data (and thus application state), so requiring the app hash seems fine - or even desirable, since we want to know as soon as possible if a state transition is invalid.

We _could_ support migrating transactions too (which might help with the SDK upgrade story as well), in which case the application would have to provide new app hashes for each derived header, but I'm not sure this is something we want to do in the first iteration - worth considering though.

cmwaters:
>I think we probably don't want to migrate transactions data (and thus application state), so requiring the app hash seems fine - or even desirable, since we want to know as soon as possible if a state transition is invalid.

Yes I think, at least to start with, we should make `AppHash` part of the core verification data. 

> A middle ground might be allowing a separate migration script to create the derived structures while Tendermint is also producing blocks, such that most of the history can be rewritten "offline", and at the upgrade height we'll only need to shut down while migrating e.g. the last 1000 blocks or so - which would presumably only take a few minutes.

I think this makes the most sense: migrating perhaps just the derived headers in the background such that consensus has the information it needs to operate continuously then making the full transition of the rest of the block structure after the upgrade height.

rfc/005-soft-chain-upgrades.md
```diff
@@ -0,0 +1,438 @@
...
+ensure that the nodes could transition smoothly. It may be that the actual
+upgrade might not take place till height: h + 2.
+
+As we expect upgrades to always increment the block version by one, rather
+than having the version number being passed in `EndBlockResponse{}`, another
+option could be to have the height where we want the upgrade to occur. With this
+approach validators are not reaching consensus on the version but the window of
+time (as a height) that nodes will have to perform or prepare for the upgrade.
+
+We will now cover the two main methods of executing a soft upgrade.
```
milosevic, line 88: It might help saying these two approaches on the opposite spectrum, and that we could also imagine having some kind of hybrid approaches that would be in between.

rfc/005-soft-chain-upgrades.md
```diff
@@ -0,0 +1,438 @@
...
+the loss of transaction history.
+
+The scope of this RFC is to address the need for Tendermint to have better
+flexibility in improving itself whilst offering greater ease of use to the
+networks running on Tendermint. This is done by introducing a division in the
+way a network can upgrade: either soft (live) or hard (stop, do some changes
+then coordinate a restart). More information on the distinction as well as
+general upgrade terminology can be found in Appendix A. There may still remain
+the need for hard upgrades but this will be focused [elsewhere](https://github.com/tendermint/tendermint/issues/5595)).
+Instead, this RFC will describe what supporting soft upgrades would entail
```
milosevic, line 34: It might be useful explaining briefly goals from the external users perspective (node operators, wallets, light clients, etc), so that potential solutions could be analysed from that context.

rfc/005-soft-chain-upgrades.md
```diff
@@ -0,0 +1,438 @@
...
+than having the version number being passed in `EndBlockResponse{}`, another
+option could be to have the height where we want the upgrade to occur. With this
+approach validators are not reaching consensus on the version but the window of
+time (as a height) that nodes will have to perform or prepare for the upgrade.
+
+We will now cover the two main methods of executing a soft upgrade.
+
+## Method 1: Multi Data Structure Support
+
+This would require Tendermint to support all prior block protocol version. This
```
milosevic, line 92:
```suggestion
This would require Tendermint to support all prior block protocol version since the last hard fork. This
```

rfc/005-soft-chain-upgrades.md
```diff
@@ -0,0 +1,438 @@
...
+correctly. Similarly, fast sync, the block executor and evidence module, whom
+all interact with block protocol data structures also all have access to state.
+The next few sections will explore the aspects of Tendermint that will need to
+be modified to prepare for multi versioning support.
+
+### RPC
+
+Supporting soft upgrades means that the RPC will need to be able to deliver
+information across different versions. The RPC will therefore need to know which
+heights corresponds to which version and then be able to communicate this to
```
milosevic, line 155: Note that the data structure version of the rpc query depends on the content of the query, i.e, height. If I am not wrong, then either rpc queries will be two step (make a query, full node responds with a right version (if wrong), then make a second query using right version), or rpc client will need to have logic conditional on the data itself.

rfc/005-soft-chain-upgrades.md
```diff
@@ -0,0 +1,438 @@
...
+the clients. This would most likely require clients to be versatile to handling
+the different block versions. For example block explorers would also take on the
+burden of supporting all block versions across a chain.
+
+### Light Clients
+
+Similar to RPC, light clients will also need to be wary of block protocol
+changes and be able to handle them accordingly. Depending on these changes,
+there might be the need for special verification logic across signed headers
+with different versions.
```
milosevic, line 165: Similar as above, light client will need to encode versions for height ranges (for example v10 up to height 1000, v11 up to height 2500, v12 above 2500), and then depending on the height for which query is made, right version should be used.

rfc/005-soft-chain-upgrades.md
```diff
@@ -0,0 +1,438 @@
...
+
+### Implications
+
+**Positive:**
+
+- It requires no changes to the database therefore keeps the immutable aspect
```
milosevic, line 171: It would help if this property is mentioned at the beginning of the doc, together with other nice to have properties.

rfc/005-soft-chain-upgrades.md
```diff
@@ -0,0 +1,438 @@
...
+Tendermint handles the actual transition.
+
+**Negative:**
+
+- Increased burden on RPC clients to be able to parse multiple versions
+- Bloating of code base and binary. We would also need to figure out methods
+of maintaining good code hygiene with respect to having all these versions.
+- If we separate versions i.e. ABCI and Block and P2P this could potentially
+compound the problem if we need to consider the different permutations and
+how they can be supported.
```
milosevic, line 183: Correctness analysis and proper testing need to consider cross-version scenarios. With several supported versions, this can be very complex and error prone task.

rfc/005-soft-chain-upgrades.md
```diff
@@ -0,0 +1,438 @@
...
+
+
+## Method 2: Chain Migration
+
+Chain migration might sound counter intuitive in the context of an immutable
```
milosevic, line 188: Is might be simpler to consider options from properties they ensure and not techniques that are used to ensure the properties? Method 1 could be "supporting multi-versioned data structures in a single chain across soft forks", while method 2 is "supporting single-versioned data structures in a single chain across soft forks".

rfc/005-soft-chain-upgrades.md
```diff
@@ -0,0 +1,438 @@
...
+
+**Negative**
+
+- Nodes may have to go through millions of blocks requiring extra memory and
+computation over the transition phase
+- External clients would most likely need to support both block versions over
+the transition period
+- Bootstrapping nodes and light clients require an extra step for verification.
+- We've cornered off some untouchable fields in the block structure. If we
+really wanted to change these we would need a hard upgrade.
```
milosevic, line 374: This option has higher correctness risks as bugs in the migration logic (protocol or code) might lead to safety issues (light client verification returns true where it should be false). If this option is considered, ideally the potential solution should be analysed more formally.

rfc/005-soft-chain-upgrades.md
```diff
@@ -0,0 +1,438 @@
...
+When this happens nodes can't simply stop and restart when they want but must
+coordinate a restart together. There can't be nodes with different versions on
+the same network. Not only this, but upgraded nodes can't read nor verify data
+structures from a prior major version. Thus, an upgrade for a major release
+has so far meant having to create a new chain.
+
+Soft and hard upgrades both refer to major releases but have very different
+properties.
+
+A soft upgrade like minor upgrades can happen asynchronously across the
```
milosevic, line 427: It seems that what is meant with soft upgrade is supporting nodes and clients with multiple versions to be present in the network and operate correctly. Hard fork requires everyone to be on the same version. Is this correct?

rfc/005-soft-chain-upgrades.md
```diff
@@ -0,0 +1,438 @@
...
+network. They are initiated by the application and coordinated by Tendermint's
+consensus mechanism. In one way or another they allow the chain to remain fully
+accessible and verifiable to the nodes with the latest version. In the context
+of this RFC, an example of a soft upgrade could be the removing of timestamps
+in the commit sig.
+
+A hard upgrade must be initiated and coordinated socially. The latest versioned
+node is incapable of parsing and/or verifying the previous versioned data
+structures. Once could imagine a wall between the chain of blocks. An example of
+a hard upgrade in the context of this RFC could be switching from delayed
```
milosevic, line 437: It might be very informative having examples of what would qualify as a soft or hard fork upgrades by looking at previous breaking changes and also anticipated future changes.

rfc/005-soft-chain-upgrades.md
cmwaters: A soft fork is when the latest *software* version i.e 0.34.0 can support an entire blockchain. For example if going from v6 to v7 is a soft fork then v7 must support the same blockchain that v6 supported whether through multi-version support or migrations. By support, I essentially mean be able to verify all data structures. A hard fork means that the latest version no longer supports the prior blockchain. This requires a new blockchain.

rfc/005-soft-chain-upgrades.md
cmwaters: Yeah I can try add some examples

rfc/005-soft-chain-upgrades.md
cmwaters: Hmm so the way I see it is that both of them support changes to software running on an existing blockchain (which is the main goal/ desirable property). The technique which they achieve this broad property is different and because of that there are several side effects that we need to consider.

rfc/005-soft-chain-upgrades.md
cmwaters: Do you think it would be better to describe the properties of the methods first and then the method itself afterwards?

rfc/005-soft-chain-upgrades.md
cmwaters: We expect, with this method, that the rpc client will have the conditional logic to process the data structure correctly. This means it should support all the versions that the full node supports. the version number will be put in the call i.e. /v3/block?height=100 where the version is the latest that rpc client supports. If it is insufficient then the full node will return some versioning error else it will return the data structure and the version (which could be v1 or v2)

cmwaters: I will try make this clearer in the section here

rfc/005-soft-chain-upgrades.md
cmwaters: I just saw that I have an example for each in the appendix

rfc/005-soft-chain-upgrades.md
cmwaters: yup. Although the version is already encoded in the header. The light client just needs to read this once it is requested and then process the header accordingly.

rfc/005-soft-chain-upgrades.md
milosevic: Yes. I think stating properties upfront has the value on its own as someone might figure out different way to address it we haven't thought of. So I would not think necessarily about this as stating properties of the methods, but more properties of the problem you are trying to solve, and then different methods will fulfil some of the properties.

rfc/005-soft-chain-upgrades.md
milosevic: And what about previous versions of software (since the last hard-fork)? Will it still expected to work with at least heights before soft-fork they are lagging behind?

cmwaters: depends on the implementation. With the migrations, you probably won't be able to operate with a lesser version (even at a lower height) because all the blocks will be updated to the latest version and other peers will be serving these latest blocks. Perhaps if you were peered with other nodes that were on the previous version then they would still be serving those blocks. But once you reached the soft upgrade height you wouldn't be able to go any further.

This last bit would be the same with the multi support version. Nodes on prior versions could still process earlier blocks but would halt when they reached the soft upgrade height

cmwaters: I guess something else to consider is that because we're running this off consensus, it means we could also get version-forks if 2/3 of validating power is malicious. This would mean that the network forks with some nodes following version x and others following version y

cmwaters:
@marbar3778 found a blockchain project that is already implementing multi-version support that is quite similar to what I imagined/described above. If anyone is interested in seeing how it would work you can look here: https://github.com/harmony-one/harmony/tree/main/block

rfc/005-soft-chain-upgrades.md
```diff
@@ -0,0 +1,473 @@
...
+revisited in the future.
+
+### Transitions
+
+Networks would be expected to announce upgrades in advance allowing a window for
+node operators to upgrade to a software version of Tendermint that would
+comply with the announced block protocol upgrade. This can be done
+asynchronously with node operators temporarily halting their node, changing to
+the new binary and starting up again. Tendermint could also offer a tool to
+easily swap between binaries thus reducing the amount of down-time that nodes
```
ebuchman, line 91: I guess this is basically what Regen has built in the https://github.com/cosmos/cosmos-sdk/tree/master/cosmovisor ?

rfc/005-soft-chain-upgrades.md
```diff
@@ -0,0 +1,473 @@
...
+The actual switch would occur in the same manner as an App Version change, using
+the `EndBlockResponse{}` ABCI call to indicate to Tendermint that a different
+block protocol be used. In the case that a node failed to upgrade, the node
+would gracefully stop and error.
+
+Given the delayed execution model, careful consideration would be needed to
+ensure that the nodes could transition smoothly. It may be that the actual
+upgrade might not take place till height: h + 2.
+
+As we expect upgrades to always increment the block version by one, rather
```
ebuchman, line 103: Is there a reason to enforce this? I could imagine an upgrade skipping some versions, but maybe the idea is that it will be easier to provide backwards compatibility in the software for 1 rather than 2 or more past versions?

rfc/005-soft-chain-upgrades.md
```diff
@@ -0,0 +1,473 @@
...
+than having the version number being passed in `EndBlockResponse{}`, another
+option could be to have the height where we want the upgrade to occur. With this
+approach validators are not reaching consensus on the version but the window of
```
ebuchman, line 106: Not sure I understand what this is saying, maybe can be clarified?

rfc/005-soft-chain-upgrades.md
```diff
@@ -0,0 +1,473 @@
...
+- Increased burden on RPC clients to be able to parse multiple versions.
+- Bloating of code base and binary. We would also need to figure out methods
+of maintaining good code hygiene with respect to having all these versions.
+- If we separate versions i.e. ABCI and Block and P2P this could potentially
+compound the problem if we need to consider the different permutations and
+how they can be supported. Correctness analysis and proper testing would be
+required to mitigate errors.
+
+
+## Method 2: Chain Migration
```
ebuchman, line 218: I think this sounds pretty complex and I'm not sure it's worth it in the end, especially since it seems you either need a bunch more storage if you want to regenerate the original blocks or you basically need to rerun consensus for the whole blockchain or something? Also need to be careful if you're resigning things to make sure there's no risk of double signing.

ebuchman (commented):
Thanks for writing this up! I think I'd be much more in favour of the first method (Multi Data Structure Support). While it does add some maintenance overhead, it seems less magical/surprising. I also think it could really help to decouple parts of the codebase and gain clarity on the APIs and relationships, which could lead to net maintenance benefits in the long term. 

Of course the devils in the details here, and we would probably benefit from some more preliminary work to actually see what this would look like in a real example or two, and help build a sense of things we could start doing to better prepare the codebase for it. For instance what would this look like for an mvp proposer-timestamp change, or for an mvp header change?

rfc/005-soft-chain-upgrades.md
cmwaters: Yup I was thinking of the cosmosvisor when I wrote this (although I'm not too familiar with it)

rfc/005-soft-chain-upgrades.md
```diff
@@ -0,0 +1,473 @@
...
+```
+
+In this case, Block which falls under a data structure of the block protocol
+could have many versions and is represented here as an interface. `State` and
+`BlockExecutor` are structs outside of the block protocol so would be
+internalized to allow for changes (so long as this wouldn't change the overall
+behavior).
+
+Another similar approach is to have the prior versions in a different repo and
+import it as a library. This may minimize the amount of code in the repo.
```
tessr, line 163: This sounds like it may lead to dependency hell, but if this were a priority, we could structure our codebase to support it.

rfc/005-soft-chain-upgrades.md
tessr: I feel like the big downside to the Multi Data Structure approach is that it puts some maintenance burden on clients and wallets as well. This is OK if they only need to support a handful of versions, but it could be prohibitively complex if we introduce a dozen versions or something

rfc/005-soft-chain-upgrades.md
cmwaters: I was thinking that the app could return, in `EndBlock`, the height with which the upgrade would occur rather than what version to run on (so consensus could agree on the upgrade height) but on second thought I don't think it's a good idea so just scratch this sentence.

rfc/005-soft-chain-upgrades.md
```diff
@@ -0,0 +1,473 @@
...
+prior block version. This would allow external clients on prior block versions
+to still be able to process the blocks.
+- Async upgrading. Nodes choose when they want to upgrade the binary and
+Tendermint handles the actual transition.
+- In the future it may be possible to support application migrations. This is
+where the application can upgrade the Tx data structures. Generally, this option
+might offer a broader set of changes that can be soft upgraded.
+
+**Negative**
+
```
tessr, line 400: I'd add to the negatives that this is also the more complex solution to implement in the short term.

tessr: (This is sort of implied, but good to call out explicitly.)

rfc/005-soft-chain-upgrades.md
cmwaters: Hmm no. I was rather presuming this nature but I agree that it might be beneficial to allow skipping versions (even potentially reverting versions if for example a bug was found)

rfc/005-soft-chain-upgrades.md
tessr: Also: Requires building some kind of separate migration tool, which probably wouldn't get exercised that often and would be at-risk for getting musty and outdated

rfc/005-soft-chain-upgrades.md
```diff
@@ -0,0 +1,473 @@
...
+upgrade](https://blog.cosmos.network/cosmos-hub-3-upgrade-announcement-39c9da941aee)).
+This means that each block protocol version maps to the entire life of a single
+chain. The result is that networks either forego new features for long periods
+of time or have to deal with the pains of upgrading such as the coordination
+effort, state migration, intra-chain security challenges and most importantly,
+the loss of transaction history.
+
+The scope of this RFC is to address the need for Tendermint to have better
+flexibility in improving itself whilst offering greater ease of use to the
+networks running on Tendermint. This is done by introducing a division in the
```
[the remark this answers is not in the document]
cmwaters: Was there a concept of soft upgrades before in tendermint? 

I'm guessing before this, operators would perform minor upgrades to their nodes within a live network (especially if it was a security fix) in their own time - so this would constitute a soft upgrade.

In this case you're right. In the pie of possible upgrades we are trying to get a bigger slice of them to not require a hard reset.

rfc/005-soft-chain-upgrades.md
```diff
@@ -0,0 +1,473 @@
...
+  down-time and intuitive / easy UX when it comes to upgrading.
+
+- Wallets, Block Explorers and other clients: data retrievability which means
+  ensuring that helpful/informative data is always available and that minimal
+  infrastructure is needed to support serving data across the entire lifespan of
+  the chain.
+
+### Protocol Versioning
+
+Tendermint currently has three protocol versions: Block, P2P and App.
```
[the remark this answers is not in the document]
cmwaters: Do you think it would be helpful to add more context on this? I didn't personally come up with the protocol versions but the abstraction does make sense in terms of separate parts of the code.

Another way that I thought this could be done is to look at it from the stakeholders angle. We kind of have three stakeholders: the application developers, the node operators, and the "end users" which could be wallets and block explorers. A nice way to version things would be to indicate whether the changes require an involvement from that stakeholder. So ABCI and App version correlates to application developers, RPC version (the one I proposed) correlates with external clients. I'm not sure where the node operators fit but they would need to make changes to their binaries to support different block versions (and potentially p2p versions).

rfc/005-soft-chain-upgrades.md
cmwaters: > As an example, imagine that Software Version X supports Block Versions Y and Y+1. When you want to introduce Block Version Y+2, you could release Software Version X+1, which would support Y+1 and Y+2 but could drop support for Block Version Y.

Hmm, the problem with this is that all the heights that were made during Block Version Y are now lost (or can only be retrieved by processes that can still read and validate Block Version Y. This is not a problem for light clients that only follow the last 10,000 or so blocks but for nodes fast syncing or state syncing this is impossible unless they download the versions and there are archive nodes still serving blocks of those versions

rfc/005-soft-chain-upgrades.md
cmwaters: Sure we could have the table in the UPGRADES file or somewhere else. It would also be nice to add this in the `tendermint version` command. 

I was thinking that a group of ABCI, block and p2p versions could all correlate to a spec version and then a software version just maps out which spec versions it covers although I'm not sure this would exactly work

rfc/005-soft-chain-upgrades.md
```diff
@@ -0,0 +1,473 @@
...
+to one another and persist to disk have changed in a way that is incompatible
+with prior releases. A block protocol change is always a major release.
+
+When this happens nodes can't simply stop and restart when they want but must
+coordinate a restart together. There can't be nodes with different versions on
+the same network. Not only this, but upgraded nodes can't read nor verify data
+structures from a prior major version. Thus, an upgrade for a major release
+has so far meant having to create a new chain.
+
+Soft and hard upgrades both refer to major releases but have very different
```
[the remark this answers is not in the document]
cmwaters: But I think all soft upgrades are major releases because external clients need to change their code in order to verify the new block format.

I guess my question here is whether an incompatible change constitutes only readability or also verifiability. If it's just about reading the blocks then maybe some soft upgrades could be released in a minor version

rfc/005-soft-chain-upgrades.md
tessr: Ah true, good point!

cmwaters:
I want to provide a quick update as to the current stage of this RFC and a quick outline of how I plan to move it forward. 

It seems that there is a growing preference towards using multi data structure support (MDSS) as a means of enabling soft chain upgrades. With this in mind I would like to give the holiday break as one last period for any opposing views / ideas towards MDSS to surface. If we are all convinced by this approach come the new year, then I will begin rewriting this RFC, narrowing down the scope by moving the migrations section into the alternatives section (with the bulk of it residing in the appendix) and focusing purely on MDSS, with a clearer distinction between what changes are a soft and hard upgrade and how exactly will soft upgrades be executed. 

Following from this, I would then appreciate getting some final feedback and a thumbs up from everyone if they are happy with this and accept the proposal.

The steps following from this RFC will be:
 - To write a page on how Tendermint conducts versioning which will reside in the spec repo. 
 - To write up an ADR for how to implement what has been decided in this RFC (Tendermint-Go) 
 - To write a document (whether it be another RFC or something else) on a proposal for improving the UX / data availability of hard upgrades.

tessr:
That sounds like a great plan, @cmwaters! I'd say the rewrite of this RFC should definitely include some discussion of the alternative to MDSS and perhaps a record of our decision-making, but there's no need to dig deeply into the details there. Also no need to rush into this before the holidays; I'll share this a little more broadly to solicit more feedback during that time.

PR #223, opened 2020-11-24
abci: add abci_version to requestInfo

marbar3778:


melekes (approved):

spec/abci/abci.md
```diff
@@ -237,6 +237,7 @@ via light client.
     - `Version (string)`: The Tendermint software semantic version
     - `BlockVersion (uint64)`: The Tendermint Block Protocol version
     - `P2PVersion (uint64)`: The Tendermint P2P Protocol version
+    - `AbciVersion (string)`: The Tendermint ABCI semantic version
```
erikgrinaker, line 240: We should specify what "semantic version" actually means, I don't see any description of it anywhere in the spec.

erikgrinaker: Also, shouldn't it be `ABCIVersion`, not `AbciVersion`?

marbar3778: I left it as it would be generated in go without gogoproto. I changed it to `ABCI`.

erikgrinaker (approved):

PR #224, opened 2020-11-27
RFC: ReverseSync - fetching historical data

cmwaters:
This RFC addresses the need for Tendermint nodes to be able to fetch, verify and persist historical blocks and state data

[Rendered](https://github.com/tendermint/spec/blob/callum/backfill-rfc/rfc/004-reverse-sync.md)


rfc/006-backfill-blocks.md
```diff
@@ -0,0 +1,155 @@
...
+type snapshot struct {
+ 	Height   uint64
+ 	Format   uint32
+ 	Chunks   uint32
+ 	Hash     []byte
+ 	Metadata []byte
+
+ 	trustedAppHash []byte // populated by light client
+}
+```
```
erikgrinaker, line 46: This is not what is offered to the ABCI application, but the internal representation in the Tendermint Go implementation. The ABCI application receives this Protobuf message:

```protobuf
message Snapshot {
  uint64 height   = 1;  // The height at which the snapshot was taken
  uint32 format   = 2;  // The application-specific snapshot format
  uint32 chunks   = 3;  // Number of chunks in the snapshot
  bytes  hash     = 4;  // Arbitrary snapshot hash, equal only if identical
  bytes  metadata = 5;  // Arbitrary application metadata
}
```

rfc/006-backfill-blocks.md
```diff
@@ -0,0 +1,155 @@
...
+The application will then be expected to parse information in the meta data
+regarding the backfill height and then return it in the response (alongside the
+app's decision):
+
+```proto
+message ResponseOfferSnapshot {
+  Result result = 1;
+  int64 backfill_height = 2;
+}
+```
```
erikgrinaker, line 57: I think the backfill height should be part of the snapshot metadata, and thus given by the node that took the snapshot. Remember, the local application has no state (it is about to be restored from a snapshot), so it may not have  necessary information to calculate the backfill height. The application that took the snapshot does. E.g.:

```protobuf
message Snapshot {
  uint64 height          = 1;  // The height at which the snapshot was taken
  uint32 format          = 2;  // The application-specific snapshot format
  uint32 chunks          = 3;  // Number of chunks in the snapshot
  bytes  hash            = 4;  // Arbitrary snapshot hash, equal only if identical
  bytes  metadata        = 5;  // Arbitrary application metadata
  int64  backfill_height = 6;  // Height to backfill blocks from before starting application (inclusive)
}
```

rfc/006-backfill-blocks.md
```diff
@@ -0,0 +1,155 @@
...
+			return
+		}
+		err = blockStore.SaveSeenCommit(state.LastBlockHeight, commit)
+		if err != nil {
+			ssR.Logger.Error("Failed to store last seen commit", "err", err)
+			return
+		}
+
+    // proposed method for backfilling blocks
+    bcR.BackfillBlocks(backfill_height)
```
erikgrinaker, line 83: Indentation. Also, this is a fallible operation, so it should return an error.

rfc/006-backfill-blocks.md
```diff
@@ -0,0 +1,155 @@
...
+
+The back filling of blocks on start up would occur synchronously, completing
+this action before switching to either fast sync or consensus.
+
+Following with the notion of having the application in control of block
+retention, if for whatever circumstance the application wanted to lower the
+retention height below the current base height (whether to accommodate changing
+consensus parameters, app specific needs, or a local config), this would boot
+up the same backfill mechanism to asynchronously fetch, verify and persist
+prior blocks.
```
erikgrinaker, line 126: I think this probably needs to be considered separately, as a change in the semantics of `retain_height`. But turning a truncated node into an archive node is a valid use-case.

rfc/006-backfill-blocks.md
```diff
@@ -0,0 +1,155 @@
...
+data.
+
+This could be done in one of two ways: use the light client via the
+RPC connection or creating a new channel to send this data through.
+
+However, this separation would require multiple reads to retrieve the header at
+each height and would make it difficult to coordinate with the process running
+on the blockchain reactor. Hence I would lean towards extending the blockchain
+reactor with new channels to request this data structure, to verify it and to
+persist it to the state store.
```
erikgrinaker, line 116: We should consider this in the context of a P2P light client transport as well. We may end up implementing the same functionality both for the blockchain reactor and for a light client reactor, which seems unfortunate.

rfc/006-backfill-blocks.md
```diff
@@ -0,0 +1,155 @@
...
+Proposed
+
+## Consequences
+
+### Positive
+
+- Greater flexibility over the data that a node has. This makes it easier to
+become a node with full history.
+- Ensures that a state synced node has the adequate block history before
+being involved in consensus.
```
erikgrinaker, line 139: Even broader, the application can enforce a minimum block history invariant, for any purpose and situation.

rfc/006-backfill-blocks.md
cmwaters: Sure it depends on the breadth of scope we want this RFC to be. I was initially thinking that this was more broadly asking the question do we want Tendermint nodes to have the capability of retrieving blocks of a prior height. However we can narrow it to the case of should nodes backfill blocks at startup so that they have the prerequisite data to participate in consensus.

rfc/006-backfill-blocks.md
cmwaters: ok I can change that

rfc/006-backfill-blocks.md
cmwaters: Agree. It seems like there are several paths we can go about this. I will try enumerate them instead here to try make it clearer as to what might be the best path to take.

rfc/006-backfill-blocks.md
```diff
@@ -0,0 +1,155 @@
...
+RPC connection or creating a new channel to send this data through.
+
+However, this separation would require multiple reads to retrieve the header at
+each height and would make it difficult to coordinate with the process running
+on the blockchain reactor. Hence I would lean towards extending the blockchain
+reactor with new channels to request this data structure, to verify it and to
+persist it to the state store.
+
+The back filling of blocks on start up would occur synchronously, completing
+this action before switching to either fast sync or consensus.
```
marbar3778, line 119: Blocking joining consensus is specific to validators? Can a full node do this asynchronously as it does not need to vote on blocks?

cmwaters: Do full nodes try to validate blocks?

rfc/006-backfill-blocks.md
```diff
@@ -0,0 +1,155 @@
...
+  int64 backfill_height = 2;
+}
+```
+
+If all chunks are accepted and state sync is successful then the backfill height
+would be returned from state sync to the node so that it can execute the
+backfill process (via the blockchain reactor) to retrieve the necessary blocks.
+A backfill height of 0 would mean that this step is ignored.
+
+```golang
```
marbar3778, line 64: This seems specific to go-tendermint. Should it be moved to an ADR in the tendermint repo?

rfc/006-backfill-blocks.md
```diff
@@ -0,0 +1,155 @@
...
+requesting blocks and validating them by matching the hash of the new header
+to the LastBlockID hash in the trusted header. This would use the same block
+pool and thus would be exclusive with the fast sync process (only one can run
+at a time).
+
+For completeness, the node should also retrieve the `ValidatorSet`s,
+`ConsensusParam`'s and `ABCIResponse`'s for each of the heights. In alignment
+with the separation between state and block, it may makes sense that the state
+sync reactor rather than the blockchain reactor serve, request and verify this
+data.
```
marbar3778, line 107: Could this info be bootstrapped from the block store? Similarly to this issue: https://github.com/tendermint/tendermint/issues/3713

cmwaters: yeah I go on to talk about that

rfc/006-backfill-blocks.md
erikgrinaker: Up to you, I suppose. We do need a general backfill mechanism that can be controlled by the app and/or the operator, so I suppose it's fine to include it here if we want to tackle it at the same time.

rfc/006-backfill-blocks.md
cmwaters: Hmm or perhaps this is implementation-specific content that should more likely reside in an ADR

rfc/006-backfill-blocks.md
marbar3778: In Tendermint-go they do, but do full nodes need to verify evidence? They can if they have the data available but it may not need to a requirement.

cmwaters: Yeah see proposal no.2 I mention here: https://github.com/tendermint/tendermint/issues/5617#issue-736760581

rfc/006-backfill-blocks.md
cmwaters: Yeah I think I will include it in here.

rfc/006-backfill-blocks.md
cmwaters: Yeah on reflection I tend to agree. I have removed a lot of the implementation-specific details in the latest commit

rfc/006-backfill-blocks.md
erikgrinaker: Any reactor changes would involve additional P2P channels and messages, and those are covered by the spec.

cmwaters: Okay but does that necessarily mean we want to discuss the possible implementation paths in this RFC?

cmwaters: I guess there's no other medium with which we can do it in except in a RFC

erikgrinaker: If you want, you could always submit a spec RFC for the protocol changes and an ADR for the implementation of those changes. However, I'm not sure if we really need an ADR, since the implementation will often follow from the protocol changes.

rfc/006-backfill-blocks.md
```diff
@@ -87,15 +88,72 @@ changing consensus parameters, app specific needs, or a local config).
...
+  int64 height = 1;
+}
+```
+
+```proto
+message StateDataResponse {
+  int64 height = 1;
+  ValidatorSet validator_set = 2;
+  ConsensusParams consensus_params = 3;
+  ABCIResponses abci_responses = 4;
```
marbar3778, line 125: Do we need ABCIResponses for previous blocks?

cmwaters: It's part of the state store so for completeness we should include this.

There is however the proposal to remove it: https://github.com/tendermint/tendermint/issues/3533

erikgrinaker (commented):
👍

rfc/006-backfill-blocks.md
```diff
@@ -0,0 +1,187 @@
...
+  bytes data          = 2;
+  int64 retain_height = 3;
+}
+
+Beforehand, a retain height that was less than the nodes current base was
+ignored. With the same backfill mechanism, the application should now be able to
+lower the retention height below the current base height (whether to accommodate
+changing consensus parameters, app specific needs, or a local config).
+This accommodates the use case of a truncated node that wishes to become an
+archive node. Note this could be done simultaneously alongside consensus.
```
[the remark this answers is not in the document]
cmwaters: > How do we determine whether backfill is synchronous or asynchronous?

I'm guessing the node will determine this i.e. if it's during startup after state sync then we need to do this before consensus. Outside of startup we can run it asynchronously so long as the implementation permits it (we may not be able to drop back into fast sync if we are already backfilling).

> There's a footgun here: if the application returns retain_height: 10 on all nodes, such that all nodes delete 1-9, and then returns retain_height: 5 later, backfill will be impossible because no nodes have those blocks.

Yeah good call. It also might be that the blocks still exist but none of our peers have it. I guess we sort of imply (or hope) that applications aren't going to shoot themselves in the foot by doing this. Outside of this we can do the same thing as fastsync where we query the lowest height that our peers have and if it's greater than the retain height then we just terminate the service once we reach that height.

rfc/006-backfill-blocks.md
```diff
@@ -0,0 +1,187 @@
...
+use the embedded light client within state sync instead.
+
+This approach would separate the two forms of data. The state sync reactor would
+be responsible for verifying state data and persisting it whilst the blockchain
+reactor would process only the blocks. Operating asynchronously could cause
+inconsistencies in which data was available (if blockchain reactor is faster
+we have blocks with no validator set; if state sync reactor is faster then we
+have validator sets with no blocks). This would also require duplication in
+headers sent (one for the light client and one within the block sent to the
+blockchain reactor).
```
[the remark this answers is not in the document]
cmwaters: agree it would be consensus or at least the block executor passing a call to state sync to retrieve validator sets and consensus params. Sounds a bit messy

rfc/006-backfill-blocks.md
```diff
@@ -0,0 +1,187 @@
...
+using the hashes in the trusted header to validate the other block and state
+info. This is recursively done till the backfill height.  
+
+### Design
+
+The backfill mechanism would reside in the blockchain reactor. One may view it
+similarly to fast syncs position within the reactor as a service that can be
+started and stopped. Backfill, however, would not be an optional configuration.
+
+Two new messages would be added to allow for the passing of state data:
```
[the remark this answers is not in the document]
cmwaters: Hmm interesting. But this would wade deep into the existing blockchain reactors right?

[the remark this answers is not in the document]
erikgrinaker: I don't know about "deep", basically just some (preferably lightweight) interface to fetch this data.

ValarDragon:
Is there an alternative name for this feature that can be used? "Filling block" suggests to me filling it with txs, in which case backfilling doesn't make sense.

Perhaps `fetch historical blockdata`?

cmwaters:
@ValarDragon I think the initial coining of the term was to mean filling the gap between genesis and the nodes current base with blocks although I can also see how one might think it may mean going back and filling prior blocks with more txs. I'm not attached to the naming so I don't mind renaming this as: "fetch historical blockdata" if everyone else is in accordance

If we want something a little shorter and in alignment with the x-sync nomenclature then we could also call it "reversesync"

rfc/006-backfill-blocks.md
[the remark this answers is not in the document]
cmwaters: Do you think it makes sense to divide these messages into the three individual data types or just clump them together?

ValarDragon:
I think reversesync is a great name as well

rfc/004-reverse-sync.md
```diff
@@ -0,0 +1,218 @@
...
+
+```proto
+message StateDataResponse {
+  uint64 height = 1;
+  ValidatorSet validator_set = 2;
+  ConsensusParams consensus_params = 3;
+  ABCIResponses abci_responses = 4;
+}
+```
+
```
marbar3778, line 165: Not convinced this should introduce new messages. We can reconstruct state from blocks, making this an implementation issue not protocol level issue. 

To me this seems like our implementation does not enable this, so we would like to change the protocol. The blockchain reactor can fetch blocks at any height, the difference being when backfilling, it should not replay the block on the app. I think if we were to refactor tendermint-go to enable cross pkg communication this problem would be solved. 

Just my 2¢

cmwaters: > We can reconstruct state from blocks

Not going backwards I don't think - only forwards. We can't take state at height h-1 give it a block at height h and produce the corresponding state at height h

>  it should not replay the block on the app

How could you get `ABCI Responses` then if you didn't pass it to the application

marbar3778: > Not going backwards I don't think - only forwards. We can't take state at height h-1 give it a block at height h and produce the corresponding state at height h

If you need height 100 you would need to fetch h + 1 as well, i believe 

> How could you get `ABCI Responses` then if you didn't pass it to the application

Oh when backfilling was mentioned to me before it was only for Tendermints case. I thought that is the approach we wanted.

cmwaters: I guess the matter of ABCI responses is a separate concern. I too believe that in the long term we want to extract out ABCI Responses and Tx Indexer for that matter into its own third party thing. I would however leave it in for now and at a later date reassess its purpose.

marbar3778: I would disagree. I am opposed to adding things now if we are not sure if it will be there in the future. This would mean two protocol breaking changes. I think going with the basic approach now (non-breaking) and asses in the future if users would like to have this data.

cmwaters: > I think if we were to refactor tendermint-go to enable cross pkg communication this problem would be solved.

What do you mean by this?

marbar3778: If consensus sees evidence it tells the blockchain reactor to fetch the needed data to verify it. I didnt want to use the term `reactor` communication because the communication could include non-reactors as well.

cmwaters: > I am opposed to adding things now if we are not sure if it will be there in the future.

So the argument against this is a matter of completeness. It would be odd if a node had all the block data for a certain height yet didn't have the corresponding `ABCI Responses`.

As for the other state structures: `ValidatorSet` and `ConsensusParams`. We need this for evidence verification. There is no way currently that we can derive it from block data hence we need to introduce a new message to gossip it

rfc/004-reverse-sync.md
```diff
@@ -0,0 +1,218 @@
...
+  uint64 height = 1;
+}
+```
+
+```proto
+message StateDataResponse {
+  uint64 height = 1;
+  ValidatorSet validator_set = 2;
+  ConsensusParams consensus_params = 3;
+  ABCIResponses abci_responses = 4;
```
melekes, line 162: not sure that all applications store this info.

rfc/004-reverse-sync.md
```diff
@@ -0,0 +1,218 @@
...
+}
+```
+
+This would be sent across a new channel most likely within the blockchain reactor 
+(one can view ReverseSync as a similar process to FastSync and thus should most
+likey be in the same domain).
+
+## Alternative Solutions
+
+If we do not want to extend the blockchain's functionality, it is possible to
```
melekes, line 172: yeah, I would rather see backfilling implemented as an extention of state-sync or blockchain protocols THAN ABCI protocol (if possible). The latter puts more pressure on the ABCI app developer I believe.

rfc/004-reverse-sync.md
```diff
@@ -0,0 +1,218 @@
...
+message RequestEndBlock {
+  Height uint64
+  Base uint64
+}
+```
+
+### Verification
+
+Nodes will need to verify these prior blocks. This can be achieved by first
+retrieving the header at the base height from the block store. The node then
```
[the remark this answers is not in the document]
cmwaters: Yup true. I guess the case I talk about here is when you already have a node running and you want to reverse sync. For the case when you've just state synced and have no blocks, I believe you will have State instead which also contains the LastBlockID which you can use to start the process.

rfc/004-reverse-sync.md
```diff
@@ -0,0 +1,218 @@
...
+checks that the `LastBlockID` corresponds with the hash calculated from the
+header in the new block directly below:
+
+```go
+header[height].LastBlockID == hash(header[height-1])
```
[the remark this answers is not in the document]
cmwaters: Yeah we can. It's basically the same process of checking the LastBlockID and then once we trust the Header, verifying the rest of the data through the hashes

rfc/004-reverse-sync.md
cmwaters: I think we may have covered this in the call, but just for a written record here, I will say it here. These three data structures are what is saved in the state store. We won't be querying the application at all when we reverse sync. Executing blocks, so far, can only happen going forwards so we rely on other nodes' state stores to send us the relevant data.

rfc/004-reverse-sync.md
```diff
@@ -0,0 +1,218 @@
...
+where `Snapshot` contains the new field `backfill_height` :
+
+```proto
+message Snapshot {
+  uint64 height          = 1;  // The height at which the snapshot was taken
+  uint32 format          = 2;  // The application-specific snapshot format
+  uint32 chunks          = 3;  // Number of chunks in the snapshot
+  bytes  hash            = 4;  // Arbitrary snapshot hash, equal only if identical
+  bytes  metadata        = 5;  // Arbitrary application metadata
+  uint64 backfill_height = 6;  // Height to backfill blocks from before starting application (inclusive)
```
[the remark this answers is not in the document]
cmwaters: That's a valid concern. In RFC001 on pruning, it was decided to create a retainHeight variable and leave it up to the application to decide how much they wanted to prune as it was believed that they would have access to the most information (there may be some app specific invariants that might mean for example that they would need to keep the last 2 months of blocks). Basically we already decided to put the responsibility in the hand of the app. This is just a continuation of this.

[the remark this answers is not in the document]
cmwaters: Something I haven't considered is whether this could be used as a sort of attack. Like what if I create a snapshot that insists other nodes must reverse sync back to genesis. Perhaps we're also implying that this can be verified. 

Another idea may be to deliver all the chunks to the application first and then after the last chunk, ask the application what the retain height should be and then reverse sync to there

rfc/004-reverse-sync.md
```diff
@@ -0,0 +1,218 @@
...
+This accommodates the use case of a truncated node that wishes to become an
+archive node.
+
+#### Choosing between Strong or Weak Gaurantees about Data
+
+It can be said previously, that block pruning, with respect to the application, 
+offered strong gaurantees about data i.e. if the application pruned to height 100, 
+the operation was synchronous and thus the application knew it would prune to 100 
+before the next height. This meant that the application always knew the height and 
+base of the blockchain at every step. The opposite, weak gaurantees, implies that
```
ValarDragon, line 103:
```suggestion
base of the blockchain at every step. The opposite, weak guarantees, implies that 
```

rfc/004-reverse-sync.md
```diff
@@ -0,0 +1,218 @@
...
+archive node.
+
+#### Choosing between Strong or Weak Gaurantees about Data
+
+It can be said previously, that block pruning, with respect to the application, 
+offered strong gaurantees about data i.e. if the application pruned to height 100, 
+the operation was synchronous and thus the application knew it would prune to 100 
+before the next height. This meant that the application always knew the height and 
+base of the blockchain at every step. The opposite, weak gaurantees, implies that 
+the application can't gaurantee that what it asked from Tendermint, with respect
```
ValarDragon, line 104:
```suggestion
the application can't guarantee that what it asked from Tendermint, with respect 
```

rfc/004-reverse-sync.md
```diff
@@ -0,0 +1,218 @@
...
+
+It can be said previously, that block pruning, with respect to the application, 
+offered strong gaurantees about data i.e. if the application pruned to height 100, 
+the operation was synchronous and thus the application knew it would prune to 100 
+before the next height. This meant that the application always knew the height and 
+base of the blockchain at every step. The opposite, weak gaurantees, implies that 
+the application can't gaurantee that what it asked from Tendermint, with respect 
+to data, has happened.
+
+The downside of strong gaurantees is that it is blocking. If a validator decided
```
ValarDragon, line 107:
```suggestion
The downside of strong guarantees is that it is blocking. If a validator decided 
```

rfc/004-reverse-sync.md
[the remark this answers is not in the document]
ValarDragon: There is a difference in that the light client's `VerifyBackwards` should allow for far more parallelism in verifying / checking old blocks, whereas this method enforces a sequential check before you can ensure you've been given a valid header. This should matter in terms of bounding how much garbage data an adversarial peer can give you right?

rfc/004-reverse-sync.md
```diff
@@ -0,0 +1,218 @@
...
+
+Proposed
+
+## Consequences
+
+### Positive
+
+- Greater flexibility over the data that a node has. This makes it easier to
+become a node with full history.
+- Ensures a minimum block history invariant.
```
ValarDragon, line 192: I don't understand how this has been ensured, isn't it still up to honesty/benevolence assumptions? (Like nothing on the chain / consensus enforces this right?)

rfc/004-reverse-sync.md
ValarDragon: > So the argument against this is a matter of completeness. It would be odd if a node had all the block data for a certain height yet didn't have the corresponding ABCI Responses.

This sounds like the ABCI responses are being treated as part of a blocks state, just not in the state root. Are the ABCI responses merkelized into the header already, or do you have to re-execute the block to get them?

cmwaters: Right, I looked into this a bit further and this is a valid point. I initially thought the `LastResultsHash` in the header could be used to verify `ABCIResponses`. But now I've just realised that the root hash is just for the `ResponseDeliverTxs`. 
```
LastResultsHash:                  ABCIResponsesResultsHash(abciResponses),
```
where
```
func ABCIResponsesResultsHash(ar *tmstate.ABCIResponses) []byte {
	return types.NewResults(ar.DeliverTxs).Hash()
}
```
Which means it doesn't include `BeginBlock` and `EndBlock` events. This is odd to me because I thought that it would have included all events that resulted from executing the block. Wouldn't this mean that if applications somehow fired different begin block and end block events that were not part of the `AppHash`, that Tendermint nodes would still agree on it? 🤔 

I would like to avoid executing prior blocks on the application. This, coupled with the growing sense that we shouldn't be saving `ABCIResponses` (as this is something that either the application or the TxIndexer should have/ be in control of), make me inclined to not worry about backfilling prior `ABCIResponses`. 

However, speaking of the txIndexer, I think it would be useful, when backfilling, to also batch the corresponding txs (`abci.TxResult`) to the TxIndexer. This means that although we won't need the entire `ABCIResponse`, we will at least need all the `DeliverTxResults`.

rfc/004-reverse-sync.md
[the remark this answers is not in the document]
cmwaters: Although it's possible to do it in a skipping manner (i.e `VerifySkipping` but backwards), `VerifyBackwards` currently only verifies sequentially which means that there is no parallelism and that it must step through every header. Thus, these algorithms are basically the same.

In case your wondering, we never got around to implementing the verify backwards skipping algorithm because we figured that most light clients would start on a trusted light block and naturally keep verifying forwards. There's probably an open issue for this but perhaps `VerifyBackwardsSkipping` is something that we can add to a future light client

rfc/004-reverse-sync.md
cmwaters: Sure, the amount of history a node has isn't enforced on chain, but this means that no honest/correct node will be capable of pruning past a certain threshold which may cause their node to panic.

ValarDragon: That makes sense, could this be changed to 
` - Ensures a minimum block history invariant given a single correct node` to reflect this?

rfc/004-reverse-sync.md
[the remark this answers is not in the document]
ValarDragon: That makes sense, didn't realize `VerifyBackwardsSkipping` is a TODO atm.

rfc/004-reverse-sync.md
```diff
@@ -0,0 +1,198 @@
...
+- Usage
+- Verification
+- Design
+- Termination
+
+### Usage
+
+For now, we focus purely on the case of a state syncing node, whom after
+syncing to a height will need to verify an adequate amount of previous data in
+order to be capable of processing new blocks and advancing sate. We can call
```
alexanderbez, line 51: I find this section a bit confusing and hard to conceptualize. Perhaps it's just me. 

> ... will need to verify an adequate amount of previous data in
order to be capable of processing new blocks and advancing sate.

What is adequate? What data exactly? I would state that we need data for blocks from the state sync snapshot back until the max evidence age, in order to process evidence and auxiliary information.

rfc/004-reverse-sync.md
```diff
@@ -0,0 +1,198 @@
...
+this the `max_historical_height`/`time` (max as in earliest/highest possible
+time/height). Both height and time are necessary as this maps to the BFT time
+used for evidence expiration. After acquiring `State`, we calculate these
+parameters as:
+
+```go
+max_historical_height = max(state.InitialHeight, state.LastBlockHeight - state.ConsensusParams.EvidenceAgeHeight)
```
alexanderbez, line 58: Shouldn't this be `min`? I'm not sure what `state.InitialHeight` is, but what if that height is past the minimum bound for processing evidence?

rfc/004-reverse-sync.md
```diff
@@ -0,0 +1,198 @@
...
+```
+
+Before starting either fast sync or consensus, we then run the following
+synchronous process:
+
+```go
+func ReverseSync(
+  max_historical_height int64
+  max_historical_time time.Time
+) error
```
alexanderbez, line 69: nit: just make this one line.

rfc/004-reverse-sync.md
```diff
@@ -0,0 +1,198 @@
...
+```
+
+Where we fetch and verify blocks until a block `A` where
+`A.Height <= max_historical_height` and `A.Time <= max_historical_time`.
+
+Upon successfully reverse syncing a node can now safely continue. As this
```
alexanderbez, line 75:
```suggestion
Upon successfully reverse syncing, a node can now safely continue. As this
```

rfc/004-reverse-sync.md
```diff
@@ -0,0 +1,198 @@
...
+```
+
+### Design
+
+This section will provide a high level overview of some of the more important
+characteristics of the design, saving the more tedious details as an ADR. This
+section is divided into two approaches that can be used to achieve the defined
+backfill mechanism based on the two networking protocols: RPC or P2P.
+
+#### RPC
```
alexanderbez, line 108: I'm confused, why does RPC need to be involved in the fetching and verification process at all?

alexanderbez: Ohh this is a "considered" approach?

rfc/004-reverse-sync.md
```diff
@@ -0,0 +1,198 @@
...
+```proto
+message ValidatorSetResponse {
+  uint64 height = 1;
+  ValidatorSet validator_set = 2;
+}
+```
+
+In addition, Tendermint currently sends messages for entire blocks only. This
+would be inconvenient if we only wanted to acquire the header (although there
+may be benefits in saving the entire block). Hence we may want to consider
```
alexanderbez, line 137: Yup, I think this is a valid point, but would reside in the blockchain reactor, no? If so, how do we tell the blockchain reactor to just get headers only during reverse syncing. Seems like it'll add to the complexity.

Perhaps we just keep the logic as-is and get the entire block, even though we only need the header.

rfc/004-reverse-sync.md
```diff
@@ -0,0 +1,198 @@
...
+characteristics of the design, saving the more tedious details as an ADR. This
+section is divided into two approaches that can be used to achieve the defined
+backfill mechanism based on the two networking protocols: RPC or P2P.
+
+#### RPC
+
+The aforementioned data is already available via the following RPC endpoints:
+`/blockchain` or `/commit` for `Header`'s' and `/validators` for
+`ValidatorSet`'s'. It may be the best option to fetch and verify these resources
+over RPC. Statesync already requires a list of RPC endpoints, so these could be
```
melekes, line 113: I think we should really consider Erik's light client reactor

rfc/004-reverse-sync.md
cmwaters: the max just means we won't try to verify below the initial height (as no heights below initial height exist). It's merely a technicality

rfc/004-reverse-sync.md
cmwaters: Well we use RPC for state sync and for the light client. So it already has previous use cases (even if it is frail)

cmwaters: Although, expanding the p2p gossip protocol will allow both the light client and state sync to use p2p instead of RPC

rfc/004-reverse-sync.md
cmwaters: Yeah we can decide this at a later date. Personally I feel splitting out the components of the block as individual messages makes sense for a few reasons:
- Headers and Commits represent a fraction of the size of an entire block - we don't want to be sending large amount of data/load across the network that isn't actually used.
- Division of the block aligns with the structure of the blockstore which stores commits, headers and the rest of the block in different compartments.
- It aligns with the abstraction between verification and execution (at least in a delayed execution model) where headers and commits are used solely for verification and the txs in a block are used for execution
- It aligns with the use case of state sync and light clients. If we want to integrate these components into the p2p layer, where these components only require headers and commits and not the entire block, then it makes sense that we divide these messages.

rfc/004-reverse-sync.md
cmwaters: Yeah if we went with extending the p2p layer then I think Erik's approach would be the strongest candidate when it comes to the actual implementation.

rfc/004-reverse-sync.md
cmwaters: In any case, I think I can try reword this better

rfc/004-reverse-sync.md
alexanderbez: How do we use RPC state syncing??

cmwaters: In order to verify the app hash of snapshot it runs an internal light client which uses RPC endpoints to verify the header at the height of the snapshot and thus the app hash

rfc/004-reverse-sync.md
cmwaters: There's this closed PR https://github.com/tendermint/tendermint/pull/4508

cmwaters: That's about all the documentation we have on that

rfc/004-reverse-sync.md
alexanderbez: It uses a `proxy.AppConnQuery`. Does that mean RPC is used?

josef-widder:
For the protocols I recently worked on, for 
- evidence handling 
- handling lightnode requests for headers

having a complete record (up to the trusting period in the past) of the headers is sufficient. Are there cases where we need the complete blocks?

In case we only need the headers, I agree that conceptually the problem is similar to backwards verification in the lightclient.

More generally, I can imagine that several design decisions in the past are based on the assumption that full nodes have a complete history. Did someone do a systematic review of the protocols where such an implicit assumption may be used?

rfc/004-reverse-sync.md
```diff
@@ -0,0 +1,198 @@
...
+
+## Context
+
+Two new features: [Block pruning](https://github.com/tendermint/tendermint/issues/3652)
+and [State sync](https://github.com/tendermint/tendermint/blob/master/docs/architecture/adr-042-state-sync.md)
+meant nodes no longer needed a complete history of the blockchain. This
+introduced some challenges of its own which were covered and subsequently
+tackled with [RFC-001](https://github.com/tendermint/spec/blob/master/rfc/001-block-retention.md).
+The RFC allowed applications to set a block retention height; an upper bound on
+what blocks would be pruned. However nodes who state sync past this upper bound
```
milosevic, line 22: I don't really understand what you want to say here. Can you be a bit more precise about the motivating use case?

rfc/004-reverse-sync.md
```diff
@@ -0,0 +1,198 @@
...
+(which is necessary as snapshots must be saved within the trusting period for
```
milosevic, line 23: Is it correct saying that retention height and unbonding period are not correlated at all currently, i.e., we don't ensure that honest full nodes keep blocks during the unbonding period?

rfc/004-reverse-sync.md
```diff
@@ -0,0 +1,198 @@
...
+- Usage
+- Verification
+- Design
+- Termination
+
+### Usage
+
+For now, we focus purely on the case of a state syncing node, whom after
+syncing to a height will need to verify historical data in order to be capable
+of processing new blocks. We can denote the earliest height that the node will
```
milosevic, line 51: I don't get this. Why you need to verify historical data (what do you mean by historical data here?) to be able to process new blocks?

rfc/004-reverse-sync.md
milosevic: I also find naming confusing here. I guess you are trying to identify max block such that both block height and block time satisfy evidence age parameters as you have written below. It might be easier to understand if you move property from line 69 up front.

rfc/004-reverse-sync.md
```diff
@@ -0,0 +1,198 @@
...
+
+```go
+func ReverseSync( max_historical_height int64, max_historical_time time.Time) error
+```
+
+Where we fetch and verify blocks until a block `A` where
+`A.Height <= max_historical_height` and `A.Time <= max_historical_time`.
+
+Upon successfully reverse syncing, a node can now safely continue. As this
+feature is only used as part of state sync, one can think of this as merely an
```
milosevic, line 73: Do we ensure that this property always hold, i.e., that retentionHeight does not lead to removing blocks needed to verify evidence?

rfc/004-reverse-sync.md
```diff
@@ -0,0 +1,198 @@
...
+extension to it.
+
+In the future we may want to extend this functionality to allow nodes to fetch
+historical blocks for reasons of accountability or data accessibility.
```
milosevic, line 77: Fork accountability (if this is what you mean by accountability) operates in the same security model, so the same lower bound for needed block applies also in that case. What do you mean by data accessibility here?

rfc/004-reverse-sync.md
```diff
@@ -0,0 +1,198 @@
...
+
+### Verification
+
+ReverseSync is used to fetch and verify the following data structures:
+- `Header`
+- `ValidatorSet`
```
milosevic, line 83: I am not sure to understand this. Given a trusted block B, you can use backward verification to verify previous blocks by following hash links in the past. Why you need to check valsets explicitly?

rfc/004-reverse-sync.md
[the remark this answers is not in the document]
milosevic: How you can implement VerifyBackwardsSkipping? My understanding was that backward verification is strictly sequential as we need to follow the hash links.

cmwaters:
> having a complete record (up to the trusting period in the past) of the headers is sufficient. Are there cases where we need the complete blocks?

An earlier revision of this RFC proposed to retrieve the complete block in order to give nodes the option to go from having a truncated history (i.e. when starting from state sync) to having full history. This idea was abandoned and I simplified the RFC to look purely at the problem where a node that has state synced doesn't have all the headers and validator sets within the unbonding period.

We need both headers and validator sets. Headers are used for verification and are used as the trusted header in the event of `LightClientAttackEvidence` and validator sets are used to confirm that the malicious validator is still bonded (within the evidence time).

> More generally, I can imagine that several design decisions in the past are based on the assumption that full nodes have a complete history. Did someone do a systematic review of the protocols where such an implicit assumption may be used?

When pruning and starting at an initial height was introduced we tried as far as the implementation went to sweep the entire code for assumptions that were now invalid. I believe we got most of them but I still see some cases where the assumption that a node has full block history is made.

rfc/004-reverse-sync.md
cmwaters: Yes, this is the crux of the problem

rfc/004-reverse-sync.md
cmwaters: Yup.

By historical data I mean `Header` and `ValidatorSet`. I explain this later on but for now I was intentionally trying to keep this high level to start with. I can add exactly what data is needed at the start if you think this is a better structure.

josef-widder: I mentioned that somewhere else. In general we should do a closer review about where in the current protocol we assume full nodes to have all past blocks. For instance, in Fastsync, I can also ask peers for blocks. If they don't have historical data they cannot respond. They might be removed by the syncing node as peer, although they are correct. I suspect this implicit assumption might be around in several places.

rfc/004-reverse-sync.md
cmwaters: One of the original motivations for ReverseSync was to take a node that had truncated history and to be able to turn it into an archive node. This is a node that has all the blocks from genesis height. 

When pruning and state sync was introduced, a chief concern was that it incentivises nodes to keep only what they need. The most alarming result to this is that if every node truncated their history then all data below that height would be lost forever. New nodes would not be able to fast sync from genesis and no transactional records during that era would exist. Accountability and data accessibility refer to this problem - that the network should have a healthy amount of data available.

rfc/004-reverse-sync.md
```diff
@@ -0,0 +1,198 @@
...
+also doesn't need to rely on RPC as the adequate data can now be transmitted
+through P2P.
+
+### Termination
+
+ReverseSync draws a lot of parallels with fast sync. An important consideration
+for fast sync that also extends to ReverseSync is termination. ReverseSync will
+finish it's task when one of the following conditions have been met:
+
+1. It reaches a block `A` where `A.Height <= max_historical_height` and
```
milosevic, line 151: What we use for as an upper bound of the unbonding period window, now or bft time of the state synced state? In the former case, we have a moving target for termination.

rfc/004-reverse-sync.md
cmwaters: Currently we don't, but this RFC would introduce one.

rfc/004-reverse-sync.md
cmwaters: Yeah so I initially wasn't sure with which perspective to apply this. It might be intuitive to think of the minimum amount of headers that a node needs. But as this measure is in terms of height, I tended towards the highest height that nodes can prune to. I can change the nomenclature if we can think of a more intuitive way to describe it

milosevic (changes requested):
Good work!I believe there are still few open points that should be clarified/agreed upon before we can merge this. More particularly, it would be good if there is more clarity what exactly we want to achieve with reversesync (my understanding is supporting evidence handling is the main use case), and how we want to do it can be left for ADR.

rfc/004-reverse-sync.md
cmwaters: Checking the val sets has nothing to do with the backwards verification algorithm itself, it's more because we want validator sets as a necessary data structure when verifying evidence. Thus when a node sends us a validator set we need to check that we can trust it.

rfc/004-reverse-sync.md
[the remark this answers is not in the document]
cmwaters: My thinking is that it should be possible but I could be wrong.

Say you wanted to verify backwards from height 100 to height 10 (Both of them are within the trusting period). If 1/3+ of the validators in height 100 were also present in the validator set at height 10 and signed the header at that height we should be able to trust the header in the same way that we can trust via skipping verification. We would then still run header comparison with witnesses. If 1/3+ had performed a lunatic attack and a witness was able to produce a conflicting header then we would create evidence and slash the common validators that were in height 100 and signed in height 10

rfc/004-reverse-sync.md
cmwaters: bft time of the state synced state

cmwaters:
I just did another pass through of the evidence logic. It seems that we will also need the `Commit`. This means that reverse sync needs to be able to fetch and verify:
- Header
- Commit
- Validator Set

>  it would be good if there is more clarity what exactly we want to achieve with reversesync (my understanding is supporting evidence handling is the main use case)

This is the only use case at the moment.

Prior we thought about adding it as a tool to turn a node with truncated history to become an archive node with full history but we decided to simplify the proposal. (See https://github.com/tendermint/spec/pull/224#discussion_r598571567).

Given that we narrowed the scope of this RFC, then perhaps we don't need to call it something like reverse sync as it will just be an additional component of state sync

milosevic (approved):

cmwaters:
I have polished up this RFC to include some of the discussions held at the dev call a few weeks back where we decided to use the p2p layer and thus introduce two new messages.

I will look to merge this by the end of the day in case anyone wants to look over it one last time

PR #225, opened 2020-11-30
Draft of evidence handling for discussion

josef-widder:
This is the first draft of the English specification. It will be accompanied by a TLA+ model showing completeness of the approach.

### DIFF
--- a/rust-spec/lightclient/attacks/isolate-attackers_001_draft.md
+++ b/rust-spec/lightclient/attacks/isolate-attackers_001_draft.md
@@ -46,7 +46,7 @@ type LightClientAttackEvidence struct {
 The isolator is a function that gets as input evidence `ev`
 and a prefix of the blockchain `bc` at least up to height `ev.ConflictingBlock.Header.Height + 1`. The output is a set of *peerIDs* of validators.
 
-We assume that the full node is synchronized with be blockchain and has reached the height `ev.ConflictingBlock.Header.Height + 1`.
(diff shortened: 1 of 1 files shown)


konnov (commented):
Small comments so far

rust-spec/lightclient/attacks/isolate-attackers_001_draft.md
```diff
@@ -0,0 +1,216 @@
...
+    } 
+}
+```
+- Implementation comment
+    - If the full node has only reached height `ev.conflictingBlock.Header.Height` then `bc[ev.conflictingBlock.Header.Height + 1].Header.LastCommit` refers to the locally stored commit for this height. (This commit must be present by the precondition on `length(bc)`.)
+    - We check in the precondition that the unbonding period is not expired. However, since time moves on, before handing the validators over Cosmos SDK, the time needs to be checked again to satisfy the contract which requires that only bonded validators are reported. This passing of validators to the SDK is out of scope of this specification.
+- Expected precondition
+    - `length(bc) >= ev.conflictingBlock.Header.Height`
+    - `ValidAndVerifiedUnbonding(bc[ev.CommonHeight], ev.ConflictingBlock) == SUCCESS`
+    - `ev.ConflictingBlock.Header != bc[ev.ConflictingBlock.Header.Height]`
```
konnov, line 108:
```suggestion
    - `ev.ConflictingBlock.Header != bc[ev.ConflictingBlock.Header.Height]`
    - TODO: input light blocks pass basic validation
```

konnov (approved):
Let's merge to unblock the other PRs

PR #227, opened 2020-12-01
rfc: add P2P stream proposal

erikgrinaker:
[Rendered](https://github.com/tendermint/spec/blob/erik/rfc-p2p-streams/rfc/004-p2p-streams.md)

Proposes two changes to the MConnection protocol to better accomodate multi-stream protocols such as QUIC.

### DIFF
--- a/rfc/004-p2p-streams.md
+++ b/rfc/004-p2p-streams.md
@@ -1,0 +1,204 @@
+# RFC 004: P2P Stream Support
+
+## Changelog
+
+- 2020-12-01: Initial draft
+
+## Author(s)
+
+- Erik Grinaker (@erikgrinaker)
+
+## Context
+
+The Tendermint Core team is currently refactoring the P2P stack, with a planned architecture outlined in [ADR 062](https://github.com/tendermint/tendermint/blob/master/docs/architecture/adr-062-p2p-architecture.md). In parallel, Informal Systems is planning to implement P2P support in Tendermint-rs, and have expressed interest in using QUIC as a transport protocol instead of the current proprietary MConnection protocol.
+
+In order to take full advantage of QUIC streams, the proposed new P2P architecture has first-class support for network streams. Each reactor channel maps onto a separate network stream. However, the current MConnection protocol, which uses multiplexing, cannot map onto an idiomatic stream API due to a couple of impedance mismatches. This leaves three options:
+
+1. Make two changes to the MConnection protocol to fit with an idiomatic network stream API.
+
+2. Use a non-idiomatic network API that is message-oriented rather than byte-oriented.
+
+3. Don't support network streams (and by extension, don't take full advantage of QUIC).
+
+### What's Multiplexing, and What's Wrong With It?
+
+The current Tendermint P2P transport protocol is connection-oriented. This means that when communicating with a peer over the network it uses a single TCP connection with a single IO stream. In Go, this connection has the following (reduced) interface, with similar abstractions in other languages:
+
+```go
+type Connection interface {
+    // Read reads incoming bytes into the given byte slice.
+    Read([]byte) (int, error)
+
+    // Write writes outgoing bytes to the connection.
+    Write([]byte) (int, error)
+
+    // Close closes the connection.
+    Close() error
+}
+```
+
+The Tendermint MConnection protocol has a concept of logical channels, used to separate messages passed by different reactors for different purposes. For example, consensus vote messages are passed via channel `0x22` and mempool transaction messages are passed via channel `0x30`. These channels are [multiplexed](https://en.wikipedia.org/wiki/Multiplexing) onto a single TCP connection by fragmenting and sending data in Protobuf packet messages written to the TCP stream in a FIFO (first-in first-out) manner:
+
+```protobuf
+message PacketMsg {
+  int32 channel_id = 1; // The logical channel ID
+  bool  eof        = 2; // Whether this packet is the last packet of a logical message
+  bytes data       = 3; // The packet payload
+}
+```
+
+One weakness with multiplexing is that if e.g. the mempool reactor is slow to process transaction messages then this may delay consensus vote messages that are queued behind them. This is known as [head-of-line blocking](https://en.wikipedia.org/wiki/Head-of-line_blocking).
+
+### What Are QUIC Streams, and Why Do We Want Them?
+
+[QUIC](https://en.wikipedia.org/wiki/QUIC) is a transport protocol originally designed to address head-of-line blocking in HTTP. HTTP/2 supports sending HTTP requests in parallel across multiple streams multiplexed onto a single TCP connection. However, as with Tendermint channel messages, a slow request at the head of the line will delay all requests queued behind it. QUIC does away with multiplexing, and instead provides fully-independent streams that do not affect each other, forming the basis of HTTP/3.
+
+QUIC is therefore of interest to Tendermint since it will allow e.g. consensus messages to be exchanged independently of less-important messages for other components. A Tendermint reactor channel would map directly onto a separate QUIC stream.
+
+The idiomatic way to expose such streams in network APIs is as follows:
+
+```go
+// Connection represents a single connection or session against a peer.
+type Connection interface {
+    // Stream creates a new stream within the connection.
+    Stream() (Stream, error)
+
+    // Close closes the connection and all streams within it.
+    Close() error
+}
+
+// Stream represents a single logical stream within a connection.
+type Stream interface {
+    // Read reads incoming bytes into the given byte slice.
+    Read([]byte) (int, error)
+
+    // Write writes outgoing bytes to the stream.
+    Write([]byte) (int, error)
+
+    // Close closes the stream.
+    Close() error
+}
+```
+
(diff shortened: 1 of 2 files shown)


marbar3778 (commented):
Thank you for the well written RFC!! I am in favor of this change.

cmwaters (commented):
This looks like sound reasoning to me. 

I don't know the space too well. Are there any single streamed transports we would want to consider adopting. It seems relatively incongruent with the current format of reactors to not have a transport that more natively supports multiple streams (without HOL blocking and so forth)

rfc/004-p2p-streams.md
```diff
@@ -0,0 +1,211 @@
...
+
+```diff
+  message PacketMsg {
+    int32 channel_id = 1; // The logical channel ID
+-   bool  eof        = 2; // Whether this packet is the last packet of a logical message
+    bytes data       = 3; // The packet payload
+  }
+```
+
+P2P reactor messages should instead be framed by the message sender, using length-prefixed encoding (the de-facto standard for Protobuf). This allows using e.g. a standard `protoio.Reader` and `protoio.Writer` to automate framing. An additional benefit is that the length of a message is known prior to receiving it, such that an appropriately sized buffer can be initialized immediately.
```
alexanderbez, line 118: > This allows using e.g. a standard `protoio.Reader` and `protoio.Writer` to automate framing.

This is a very important detail. Might want to add a sentence or two on how this happens.

alexanderbez (commented):
Great explanation! I'm in favor of the main/ideal proposal. I think adopting QUIC/stream-oriented transports is preferable for many reasons.

xla:
Following the developments of ADR-062 and working on the new p2p stack in tendermint-rs, I'm surprised to see this proposal. While stream based transports are on the horizon, it seems odd that it would necessitate retrofitting as it indicates a leak of abstraction at some point in the stack. Unless this is in preparation for the *Router* to be generic over the transport plugged. Yet if that is the desired outcome, a breaking protocol change for that internal convenience seems induldgent. If it is the declared outcome indeed, a less intrusive approach is to wrap the existing MConn with some thing that quacks and walks like a QUIC.

And there is a lot of mention of quic and its streams, these are byte streams which is not a sensible interface to expose up to reactors, which only depend on PeerID (origin) and message (payload). So the Router should really guard against the transport details. I refer to the Router here as it is the entity in the ADR-062 diagrams. Which also might be the angle (the layer where the Router sits) where innovation has impact as it would determine how ready the codebase is to be migrated onto something like QUIC.

It is mentioned that streams can be set up in an ad-hoc way. Is that something that has been decided/settled on? It seems like a protoocl violation if a peer is trying to initiate communication for a capability not advertised. It feels worthy to flag that this is not just a decision that should be made based on technical merit.

Lastly is it anticipated to solve any of the shortcomings with package transmission/delivery of the current implementation? Which would seem like an attempt to reimplement something like gRPC.

erikgrinaker:
>While stream based transports are on the horizon, it seems odd that it would necessitate retrofitting as it indicates a leak of abstraction at some point in the stack. Unless this is in preparation for the Router to be generic over the transport plugged.

The router, or rather the peer connection, will be "generic" over the transport - the router can communicate with different peers using different transports. We therefore need a transport abstraction. The proposal in ADR-062 is for that abstraction to be byte streams (plural), which are the de-facto low-level networking abstraction across most languages (although normally singular). The main proposal here is to change MConnection from a message-oriented multi-stream transport to a byte-oriented multi-stream transport, which is the same model that QUIC uses.

>Yet if that is the desired outcome, a breaking protocol change for that internal convenience seems induldgent. If it is the declared outcome indeed, a less intrusive approach is to wrap the existing MConn with some thing that quacks and walks like a QUIC.

Yes, that is option 2: writing a shim around MConnection that takes reassembled messages and passes them through the P2P transport API as byte streams using length-prefix framing. This has increased resource and complexity costs, so I think this is best suited as a temporary measure as long as we are fairly confident that we want to migrate to QUIC, but it is a viable option.

>these are byte streams which is not a sensible interface to expose up to reactors, which only depend on PeerID (origin) and message (payload). So the Router should really guard against the transport details.

Indeed. Reactors will exchange Protobuf messages across Go channels, using peer ID addressing. Only the router cares about low-level network transports, and it is its job to serialize messages, frame them, and pass them over the appropriate byte stream.

>It is mentioned that streams can be set up in an ad-hoc way. Is that something that has been decided/settled on?

No, we could do all stream setup during the initial transport handshake, and use QUIC-specific stream handshakes. But explicitly opening and closing channels allows them (and thus reactors) to be dynamic, and I think this is an improvement over the current state where we announce that we have some capability (e.g. consensus) but then just ignore received messages when necessary (e.g. during fast sync). This is somewhat orthogonal though, and we can defer this discussion until later without any real impact on the P2P refactor.

>Lastly is it anticipated to solve any of the shortcomings with package transmission/delivery of the current implementation? Which would seem like an attempt to reimplement something like gRPC.

I'm not sure if I understand the question. Internal message scheduling and prioritization will be revamped as part of the router, e.g. to provide fair queueing of messages received from different peers and prioritize certain reactors over others. Apart from this, the MConnection transport characteristics will largely remain unchanged. Could you elaborate on the gRPC comparison?

erikgrinaker:
>Are there any single streamed transports we would want to consider adopting. It seems relatively incongruent with the current format of reactors to not have a transport that more natively supports multiple streams (without HOL blocking and so forth).

We do sort of use a single-streamed transport currently, in the sense that TCP is single-streamed and MConnection multiplexes multiple logical streams onto a single TCP stream. UNIX sockets are another example. Most transport protocols, and thus must networking APIs, have traditionally been single-streamed, which is why it's slightly painful to introduce something like QUIC that breaks this convention by using multiple streams.

tessr:
> While stream based transports are on the horizon, it seems odd that it would necessitate retrofitting as it indicates a leak of abstraction at some point in the stack. 

I think we're basically dealing with an abstraction mismatch here, between the abstractions we already have and the abstractions we _want_ to have. The mconnection protocol necessitates that multiplexing happen within the "application" abstraction, while something like QUIC expects multiplexing to happen within the "transport" abstraction. 

There had been a lot of enthusiasm for a "transport-agnostic" P2P layer, since it seemed like that meant that we could move all the internals around while leaving the protocol in place. This also meant that we could postpone any decisions about what the transport layers should actually be. But I think we've realized that the preexisting mconnection protocol is not compatible with the concept of "transport," at least if transports behave like QUIC, since both QUIC and mconnection expect to handle multiplexing. 

It sounds like, as an alternative to this RFC, mconnection and QUIC could be shoehorned together. But this would be at the cost of a lot of complexity (and also likely a performance hit), and that would be a little bit of a shame since the whole point of this P2P refactor is to reduce complexity. 

All of that said, if we _don't_ expect to adopt QUIC (or another multi-stream transport), this whole thing is moot. Perhaps we should begin by trying to build confidence around that decision.

erikgrinaker:
To unblock the P2P refactor I'm going to implement a message-oriented transport API as a temporary solution, and we can revisit once we get alignment on a long-term strategy.

ebuchman:
I'm not sure I really understand why we would make any changes to the MConnection in a first pass here. The reactors currently use `peer.Send(chID, msgBytes)` - my impression was that `peer` would be capable of mapping that API onto either an underlying MConnection (as is) or a new QUIC connection (with streams for channel ids). In the former case, nothing would change. In the latter case, it would select the appropriate QUIC stream (putting aside the stream id problem), add a length prefix, and send it off.

Is that what Option 2 is proposing? I didn't get that from `Yes, that is option 2: writing a shim around MConnection that takes reassembled messages and passes them through the P2P transport API as byte streams using length-prefix framing`. Maybe the problem is that I'm thinking about this like, "leave the current system more or less in place without changing it, and just add support for QUIC", while it's otherwise being though more coherently as part of the larger refactor. At least to me it seems like this could be done in a straight forward non protocol breaking way, and maybe cleaned up in the larger breaking refactor when MConnection is ultimately deprecated (presumably it will be ?). But maybe I'm missing something obvious?

erikgrinaker:
Yeah, that's basically right. Taking full advantage of QUIC requires a somewhat different processing/scheduling model in the P2P stack (basically one goroutine per stream rather than one goroutine per peer), and the most natural way to do this is to expose streams as a first-class transport concept. It's just a question of whether we implement this now, or defer until we build the QUIC transport.

> Is that what Option 2 is proposing?

No, that is sort of the opposite. Option 2 instead proposes to implement the full stream-based transport API with per-stream goroutines, and shoehorn MConnection into this by basically rewriting the transport protocol on-the-fly (which is about as unsatisfying as it sounds).

I'm going to do basically what you outline for now: implement a higher-level `SendMessage(chID, msg)` and `ReceiveMessage() (chID, msg)` transport API that fits with the current MConnection semantics and protocol, and also allows us to experiment with a QUIC transport using a different framing/handshake protocol (although inefficiently since there will be a single goroutine consuming all streams per peer). Then we can defer refactoring of the transport bits until we start working on QUIC.

erikgrinaker:
Closing this, will be superseded once QUIC work begins.

PR #228, opened 2020-12-02
BFT algorithms require less than 1/3 faulty validators

shahankhatch:
I've gone through and made some changes to the wording regarding faulty validator constraints. 

BFT algorithms limit the number of faulty validators to less than 1/3, not 1/3 or more. 

See for example http://pmg.csail.mit.edu/papers/osdi99.pdf, which states that "The algorithm offers both liveness and safety provided at most floor((n-1)/3) out of a total of n replicas are simultaneously faulty".


shahankhatch:
To reference "The latest gossip on BFT consensus" paper, the equation is `n > 3f` which is equivalent to `n/3 > f`.

konnov (approved):
Hi @shahankhatch! Thank you for making this document more precise. You are right: "less than 1/3 of voting power" and "1/3 of voting power or more" does not leave room for ambiguity.

konnov:
Just FYI, we have more detailed specifications of the light client in English and TLA+ here: https://github.com/tendermint/spec/tree/master/rust-spec/lightclient. In TLA, we have to precisely write "<", e.g., see the [blockchain specification](https://github.com/tendermint/spec/blob/master/rust-spec/lightclient/verification/Blockchain_003_draft.tla#L101-L109).

konnov:
Actually, I think the problem is in the inherent imprecision of the words "1/3 of the voting power". The Tendermint paper fixes `n` to be `3 * f + 1`. The faulty validators need at least `f+1` units of the voting power to violate the security model. In this case, we are talking about **more than 1/3** (`> f`, or `>= f+1`). Here `n/3` is integer division. If we understand by "1/3 of the voting power" the rational number `n / 3`, then we are talking about **at least 1/3**.

That is why the English specifications are quite tricky to write. The interpretation is always up to the reader.

shahankhatch:
@konnov Thanks so much for reviewing, for the TLA+ spec links, and for your comments on the implied integer division. Much appreciated!

PR #231, opened 2020-12-11
The TLA+ specification of the attackers detection

konnov:
This is the intermediate spec for the larger PR.


josef-widder (approved):
amazing! looks great!!

PR #235, opened 2020-12-17
p2p: update frame size

melekes:
Reflect the change made in https://github.com/tendermint/tendermint/pull/5805

The MTU (Maximum Transmission Unit) for Ethernet is 1500 bytes.
The IP header and the TCP header take up 20 bytes each at least (unless
optional header fields are used) and thus the max for (non-Jumbo frame)
Ethernet is 1500 - 20 -20 = 1460
Source: https://stackoverflow.com/a/3074427/820520


erikgrinaker (approved):

xla (approved):
💺 🎢 🎩 🔍

PR #237, opened 2021-01-05
abci: rewrite to proto interface

### LINKED ISSUE #89: Document ABCI as a Protobuf interface instead of a Go interface
The ABCI spec currently uses field names and data types from the Go interface rather than the Protobuf interface (e.g. `AppHash` instead of `app_hash`, and `[]byte` instead of `bytes`). If we want ABCI to be an implementation-agnostic protocol, perhaps the spec should document the Protobuf interface instead?

marbar3778:
# Description

Rewrite ABCI documentation to adhere to protobuf types instead of go. 

closes #89


melekes (approved):
💛

PR #242, opened 2021-01-13
spec: remove reactor section

marbar3778:
# Description

- Remove reactor section
- Add reactor messages and channels to p2p section

closes #242


spec/p2p/messages/README.md
```diff
@@ -0,0 +1,15 @@
+---
+order: 1
+parent:
+  title: Messages
+  order: 1
+---
+
+# Messages
+
+Tendermint-Core consists of many components, while many parts of these components are implementation specific the p2p messages are not. In this section we will be covering all the p2p messages of the various components of Tendermint-Core.
```
tessr, line 10:
```suggestion
Tendermint Core consists of many components. While many parts of these components are implementation specific, the p2p messages are not. In this section we will be covering all the p2p messages of the various components of Tendermint Core.
```

This actually leads to an interesting question for me - what is the difference between Tendermint Core vs. Tendermint Consensus? I kind of think of "Tendermint Core" as "the Go implementation of Tendermint Consensus" but maybe that's not fair

marbar3778: I will reword to go with "implementation" wording. 

I also view Tendermint Core as the Go implementation, but the unfairness is right.

melekes (approved):
👍

PR #244, opened 2021-01-21
non-critical bugfix in the TLA+ spec (found by new version of apalache)

konnov:
The new version of apalache has found a bug by checking the inductive invariant, which has not been uncovered by the old version. The bug does not affect safety and accountability.


josef-widder:
Looks good. Thanks. We will check that we also fix that in proposer time...

josef-widder (approved):

PR #246, opened 2021-01-21
proto: add files

### LINKED ISSUE #245: proto: move the required proto files here
Would like to hold some discussion on if we should move proto files to the spec repo and work on path to get this done.

marbar3778:
## Description

- Add proto files from tendermint to the spec repo. 
- Until we remove gogoproto from tendermint we should keep the annotations to avoid adding them every time we fetch them. 

closes #245


proto/abci/types.proto
```diff
@@ -0,0 +1,392 @@
...
+
+message ResponseInitChain {
+  ConsensusParams          consensus_params = 1;
+  repeated ValidatorUpdate validators       = 2 [(gogoproto.nullable) = false];
+  bytes                    app_hash         = 3;
+}
+
+message ResponseQuery {
+  uint32 code = 1;
+  // bytes data = 2; // use "value" instead.
```
marbar3778, line 175: In the process of moving to 1.0 of the spec, should we reset all the field numbers?

proto/README.md
```diff
@@ -0,0 +1,16 @@
+# Protocol Buffers
+
+This sections defines the types and messages shared across implementations. The definition of the data structures are located in the [core/data_structures](../spec/core/data_structures.md) for the core data types and ABCI definitions are located in the [ABCI](../spec/abci/README.md) section.
+
+## Process of Updates
+
+The `.proto` files within this section are core to the protocol and updates must be treated as such.
+
+### Steps
+
```
marbar3778, line 10: This section needs to be worked on, but this is how I imagine the process going.

melekes (approved):

proto/abci/types.proto
melekes: I'm in favor of leaving as is. The benefit from resetting seems low comparing to breaking API again (even if it's Tendermint-go v0.35 or v1).

marbar3778: Sounds good. I can see this question coming up in the future. I'll try to find why we have it this way and define it in the readme

brapse (approved):
Fully agree with the efforts here. Great stuff 👍

proto/README.md
marbar3778: How does versioning play into these steps?

PR #248, opened 2021-01-26
params: remove block timeiota

marbar3778:
- remove blocktimeiota from params
- merge consensusparams and abci params


cmwaters (approved):
:+1:

PR #251, opened 2021-01-29
spec: add accountability to spec section

marbar3778:
moves from accountability from rust-spec to spec


marbar3778:
this should be under light client actually

PR #252, opened 2021-02-01
spec: merge rust-spec

marbar3778:
## Description

- merge rust-spec and spec directories


spec/light-client/README.md
```diff
@@ -5,66 +5,201 @@ parent:
...
+The [English specification](attacks/isolate-attackers_002_reviewed.md)
+defines the protocol that is executed on a full node upon receiving attack [evidence](detection/detection_003_reviewed.md#tmbc-lc-evidence-data1) from a lightclient. In particular, the protocol handles three types of attacks
+
+- lunatic
+- equivocation
+- amnesia
+
+We discussed in the [last part](attacks/isolate-attackers_002_reviewed.md#Part-III---Completeness) of the English specification
+that the non-lunatic cases are defined by having the same validator set in the conflicting blocks. For these cases,
+computer-aided analysis of  [Tendermint Consensus in TLA+][tendermint-accountability] shows that equivocation and amnesia capture all non-lunatic attacks.
```
cmwaters, line 197: Is this supposed to be a link?

cmwaters (approved):
Great work Marko!

I would think that eventually we may want to merge the `attacks`, `detection` and `supervisor` subdirectories as I think they all correspond to the same domain. But I will leave this to be decided upon at a later date.

melekes (approved):
👍

ebuchman:
Note the rust spec is still linekd from the Readme. Might be good to add the old rust-spec directory back and add a file notifying about the redirect at least so links dont break.

marbar3778:
Will change the readme, we don't render the Readme on any sites so we dont need to add it back.

josef-widder:
Our problem is that in https://github.com/informalsystems/tendermint-rs/tree/master/docs/spec we kept the old specs and added links (e.g. [here](https://github.com/informalsystems/tendermint-rs/blob/master/docs/spec/lightclient/verification/verification.md) to the (back then up-to-date) rust-spec. (There are several published papers that reference (https://github.com/informalsystems/tendermint-rs/tree/master/docs/spec).

marbar3778:
ah, okay. ill add it back

PR #253, opened 2021-02-01
proto: modify height int64 to uint64

marbar3778:
## Description

- change `height`, `base` from int64 to uint64


cmwaters:
I think `proposer_priority` needs to remain a int64. I believe it goes negative after a validator has just proposed. I'm sure we could offset by a value so that all proposer priorities are non negative numbers but I think this will involve tweaking the algorithm.

Also do you plan to do other number types i.e for `round` int32 -> uint32 in a separate PR

By `base`, do you mean `initial_height`?

marbar3778:
> I think `proposer_priority` needs to remain a int64. I believe it goes negative after a validator has just proposed. I'm sure we could offset by a value so that all proposer priorities are non negative numbers but I think this will involve tweaking the algorithm.

Was just looking into tendermint to get a better understanding if its possible. Thanks for the comment, saved me some time.

proto/abci/types.proto
```diff
@@ -322,13 +322,13 @@ message TxResult {
 message Validator {
   bytes address = 1;  // The first 20 bytes of SHA256(public key)
   // PubKey pub_key = 2 [(gogoproto.nullable)=false];
-  int64 power = 3;  // The voting power
+  uint64 power = 3;  // The voting power
```
marbar3778, line 325:
```suggestion
  int64 power = 3;  // The voting power
```

marbar3778:
making this pr only contain changes to height

proto/abci/types.proto
```diff
@@ -90,7 +90,7 @@ message RequestDeliverTx {
 }
 
 message RequestEndBlock {
-  int64 height = 1;
```
cmwaters, line 93: This is a bit of an aside and perhaps not relevant to this PR, but don't you think that this file should have comments on what exactly the data structures and methods are for? Or is this covered elsewhere in the spec

cmwaters (approved):
Thumbs up from me 👍 

You just missed `LightClientAttackEvidence.CommonHeight` as far as I can see

melekes (approved):
👍

proto/abci/types.proto
marbar3778: The displayed docs are located here: https://docs.tendermint.com/master/spec/p2p/messages/. We can add comments here, they will add value to us and tm-rs team

PR #259, opened 2021-02-24
accountability-new approach

jovankomatovic:
This represents an alternative approach to the fork accountability protocol.
Namely, we propose a way that will remove completely the "query-response" subprotocol in the current version.
The benefits of the new approach are:
(1) It does not demand synchrony assumptions.
(2) Less information need to be exchanged.
(3) Always produces a self-contained proof of a misbehavior.


josef-widder:
This is a great idea! We should add a discussion of the following points to the document:
1. Does this change the structure of a block? Right now a block contains `2f+1` `precommit` messages. Would this require us to add `prevote` messages? (given that we are moving towards aggregate signatures, I don't think this will mean prohibitive space requirements.) However, my current understanding is that this is not the case, because only the evidence `Z` has this information. But it would be good to clarify.
2. What is the end-to-end expectation about accountability? Is the mentioned witness `Z` supposed to submit the `2f+1` `precommit` and `2f+1` `prevote` as evidence?

PR #261, opened 2021-03-02
Proposer-based timestamp specification

istoilkovska:
This is the first draft of the English and TLA+ specifications of proposer-based timestamp, containing:

 - [Proposer-Based Time](https://github.com/tendermint/spec/blob/ilina/proposer-based-timestamp/spec/consensus/proposer-based-timestamp/pbts_001_draft.md), the document used in the presentation during the Tendermint Dev session,
 - [Part I - System Model and Properties](https://github.com/tendermint/spec/blob/ilina/proposer-based-timestamp/spec/consensus/proposer-based-timestamp/pbts-sysmodel_001_draft.md), detailed specification of the assumed system model and the desired properties,
 - [Part II - Protocol Specification](https://github.com/tendermint/spec/blob/ilina/proposer-based-timestamp/spec/consensus/proposer-based-timestamp/pbts-algorithm_001_draft.md), detailed specification of the modified protocol,
 - [TLA+ Specification](https://github.com/tendermint/spec/blob/ilina/proposer-based-timestamp/spec/consensus/proposer-based-timestamp/tla/TendermintPBT_001_draft.tla), a TLA+ specification, which extends and modifies the existing [Tendermint TLA+ specification](https://github.com/tendermint/spec/blob/master/spec/light-client/accountability/TendermintAcc_004_draft.tla).

### DIFF
--- a/spec/consensus/proposer-based-timestamp/pbts-algorithm_001_draft.md
+++ b/spec/consensus/proposer-based-timestamp/pbts-algorithm_001_draft.md
@@ -4,7 +4,7 @@
 
 ### Outline
 
-The algorithm in the [arXiv paper](arXiv) evaluates rules of the received messages without making explicit how these messages are received. In our solution, we will make some message filtering explicit. We will assume that there are message reception steps (where messages are received and possibly stored locally for later evaluation of rules) and processing steps (the latter roughly as described in a way similar to the pseudo code of the arXiv paper).
+The algorithm in the [arXiv paper][arXiv] evaluates rules of the received messages without making explicit how these messages are received. In our solution, we will make some message filtering explicit. We will assume that there are message reception steps (where messages are received and possibly stored locally for later evaluation of rules) and processing steps (the latter roughly as described in a way similar to the pseudo code of the arXiv paper).
 
 In contrast to the original algorithm the field `proposal` in the `PROPOSE` message is a pair `(v, time)`, of the proposed consensus value `v` and the proposed time `time`. 
 
--- a/spec/consensus/proposer-based-timestamp/pbts-algorithm_001_draft.md
+++ b/spec/consensus/proposer-based-timestamp/pbts-algorithm_001_draft.md
@@ -1,0 +1,136 @@
+# Proposer-Based Time - Part II
+
+## Updated Consensus Algorithm
+
+### Outline
+
+The algorithm in the arXiv paper evaluates rules of the received messages without making explicit how these messages are received. In our solution, we will make some message filtering explicit. We will assume that there are message reception steps (where messages are received and possibly stored locally for later evaluation of rules) and processing steps (the latter roughly as described in a way similar to the pseudo code of the arXiv paper).
+
+In contrast to the original algorithm the field `proposal` in the `PROPOSE` message is a pair `(v, time)`, of the proposed consensus value `v` and the proposed time `time`. 
+
+#### **[PBTS-RECEPTION-STEP.0]**
+
+In the reception step at process `p` at local time `now_p`, upon receiving a message `m`:
+- if the message `m` is of type `PROPOSE` and satisfies `now_p - PRECISION <  m.time < now_p + PRECISION + MSGDELAY`, then mark the message as `timely`
+> if `m` does not satisfy the constraint consider it `untimely`
+ 
+
+#### **[PBTS-PROCESSING-STEP.0]**
+In the processing step, based on the messages stored, the rules of the algorithms are
+executed. Note that the processing step only operates on messages
+for the current height. The consensus algorithm rules are defined by the following updates to arXiv paper.
+
+#### New `StartRound`
+
+There are two additions
+- in case the proposer's local time is smaller than the time of the previous block, the proposer waits until this is not the case anymore (to ensure the block time is monotonically increasing)
+- the proposer sends its time `now_p` as part of its proposal
+
+#### **[PBTS-ALG-STARTROUND.0]**
+```go
+function StartRound(round) {
+  round_p ← round
+  step_p ← propose
+  if proposer(h_p, round_p) = p {
+    wait until now_p > block time of block h_p - 1 // new wait condition
+    if validValue_p != nil {
+      proposal ← (validValue_p, now_p) // added "now_p"
+    }
+    else {
+      proposal ← (getValue(), now_p)   // added "now_p"
+    }
+    broadcast ⟨PROPOSAL, h_p, round_p, proposal, validRound_p⟩
+  }
+  else {
+    schedule OnTimeoutPropose(h_p,round_p) to be executed after timeoutPropose(round_p)
+  }
+}
+```
+
+#### New Rule Replacing Lines 22 - 27
+
+- a validator prevotes for the consensus value `v` **and** the time `t`
+- the code changes as the `PROPOSAL` message carries time (while `lockedValue` does not)
+
+#### **[PBTS-ALG-UPON-PROP.0]**
+```go
+upon timely(⟨PROPOSAL, h_p, round_p, (v,t), −1⟩) from proposer(h_p, round_p) while step_p = propose do {
+  if valid(v) ∧ (lockedRound_p = −1 ∨ lockedValue_p = v) {
+    broadcast ⟨PREVOTE, h_p, round_p, id(v,t)⟩ 
+  }
+  else {
+    broadcast ⟨PREVOTE, h_p, round_p, nil⟩ 
+  }
+  step_p ← prevote
+}
+```
+
+#### New Rule Replacing Lines 28 - 33
+
+In case consensus is not reached in round 1, in `StartRound` the proposer of future rounds may propose the same value but with a different time. 
+Thus, the time `tprop` in the `PROPOSAL` message need not match the time `tvote` in the (old) `PREVOTE` messages. 
+A validator may send `PREVOTE` for the current round as long as the value `v` matches. 
+This gives the following rule:
+
+#### **[PBTS-ALG-OLD-PREVOTE.0]**
+```go
+upon timely(⟨PROPOSAL, h_p, round_p, (v, tprop), vr⟩) from proposer(h_p, round_p) AND 2f + 1 ⟨PREVOTE, h_p, vr, id((v, tvote)⟩ 
+while step_p = propose ∧ (vr ≥ 0 ∧ vr < round_p) do {
+  if valid(v) ∧ (lockedRound_p ≤ vr ∨ lockedValue_p = v) {
+    broadcast ⟨PREVOTE, h_p, roundp, id(v, tprop)⟩
+  }
+  else {
+    broadcast ⟨PREVOTE, hp, roundp, nil⟩
+  }
+}
+```
+
+#### New Rule Replacing Lines 36 - 43
+
+- As above, in the following `(v,t)` is part of the message rather than `v`
+- the stored values (i.e., `lockedValue`, `validValue`) do not contain the time
+
+#### **[PBTS-ALG-NEW-PREVOTE.0]**
+```go
+upon timely(⟨PROPOSAL, h_p, round_p, (v,t), ∗⟩) from proposer(h_p, round_p) AND 2f + 1 ⟨PREVOTE, h_p, round_p, id(v,t)⟩ while valid(v) ∧ step_p ≥ prevote for the first time do {
+  if step_p = prevote {
+    lockedValue_p ← v
+    lockedRound_p ← round_p
+    broadcast ⟨PRECOMMIT, h_p, round_p, id(v,t))⟩ 
+    step_p ← precommit
+  }
+  validValue_p ← v 
+  validRound_p ← round_p
+}
+```
+
+#### New Rule Replacing Lines 49 - 54
+- we decide on `v` as well as on the time from the proposal message
+- here we do not care whether the proposal was received timely. 
+> In particular we need to take care of the case where the proposer is untimely to one correct validator only. We need to ensure that this validator decides if all decide.
+
+#### **[PBTS-ALG-DECIDE.0]**
+```go
+upon ⟨PROPOSAL, h_p, r, (v,t), ∗⟩ from proposer(h_p, r) AND 2f + 1 ⟨PRECOMMIT, h_p, r, id(v,t)⟩ while decisionp[h_p] = nil do {
+  if valid(v) {
+    decision_p [h_p] = (v,t) // decide on time too
+    h_p ← h_p + 1
+    reset lockedRound_p , lockedValue_p, validRound_p and validValue_p to initial values and empty message log 
+    StartRound(0)
+  }
+}
+```
+
+**All other rules remains unchanged.**
+
+Back to [main document][main].
+
+[main]: ./pbts_001_draft.md
+
+[arXiv]: https://arxiv.org/abs/1807.04938
+
+[tlatender]: https://github.com/tendermint/spec/blob/master/rust-spec/tendermint-accountability/README.md
+
+[bfttime]: https://github.com/tendermint/spec/blob/439a5bcacb5ef6ef1118566d7b0cd68fff3553d4/spec/consensus/bft-time.md
+
+[lcspec]: https://github.com/tendermint/spec/blob/439a5bcacb5ef6ef1118566d7b0cd68fff3553d4/rust-spec/lightclient/README.md
--- a/spec/consensus/proposer-based-timestamp/pbts-algorithm_001_draft.md
+++ b/spec/consensus/proposer-based-timestamp/pbts-algorithm_001_draft.md
@@ -26,13 +26,28 @@ There are two additions
 - in case the proposer's local time is smaller than the time of the previous block, the proposer waits until this is not the case anymore (to ensure the block time is monotonically increasing)
 - the proposer sends its time `now_p` as part of its proposal
 
+We update the timeout for the `PROPOSE` step according to the following reasoning:
+
+- If a correct proposer needs to wait to make sure its proposed time is larger than the `blockTime` of the previous block, then it sends by realtime `blockTime + ACCURACY` (By this time, its local clock must exceed `blockTime`)
+- the receiver will receive a `PROPOSE` message by `blockTime + ACCURACY + MSGDELAY`
+- the receiver's local clock will be `<= blockTime + 2 * ACCURACY + MSGDELAY`
+- thus when the receiver `p` enters this round it can set its timeout to a value `waitingTime => blockTime + 2 * ACCURACY + MSGDELAY - now_p`
+
+So we should set the timeout to `max(timeoutPropose(round_p), waitingTime)`.
+
+> If, in the future, a block delay parameter `BLOCKDELAY` is introduced, this means 
+that the proposer should wait for `now_p > blockTime + BLOCKDELAY` before sending a `PROPOSE` message. 
+Also, `BLOCKDELAY` needs to be added to `waitingTime`.
+
 #### **[PBTS-ALG-STARTROUND.0]**
 ```go
 function StartRound(round) {
+  blockTime ← block time of block h_p - 1
+  waitingTime ← blockTime + 2 * ACCURACY + MSGDELAY - now_p
   round_p ← round
   step_p ← propose
   if proposer(h_p, round_p) = p {
-    wait until now_p > block time of block h_p - 1 // new wait condition
+    wait until now_p > blockTime // new wait condition
     if validValue_p != nil {
       proposal ← (validValue_p, now_p) // added "now_p"
     }
@@ -42,7 +57,7 @@ function StartRound(round) {
     broadcast ⟨PROPOSAL, h_p, round_p, proposal, validRound_p⟩
   }
   else {
-    schedule OnTimeoutPropose(h_p,round_p) to be executed after timeoutPropose(round_p)
+    schedule OnTimeoutPropose(h_p,round_p) to be executed after max(timeoutPropose(round_p), waitingTime)
   }
 }
 ```
@@ -82,6 +97,7 @@ while step_p = propose ∧ (vr ≥ 0 ∧ vr < round_p) do {
   else {
     broadcast ⟨PREVOTE, hp, roundp, nil⟩
   }
+  step_p ← prevote
 }
 ```
 
--- a/spec/consensus/proposer-based-timestamp/pbts-algorithm_001_draft.md
+++ b/spec/consensus/proposer-based-timestamp/pbts-algorithm_001_draft.md
@@ -4,7 +4,7 @@
 
 ### Outline
 
-The algorithm in the arXiv paper evaluates rules of the received messages without making explicit how these messages are received. In our solution, we will make some message filtering explicit. We will assume that there are message reception steps (where messages are received and possibly stored locally for later evaluation of rules) and processing steps (the latter roughly as described in a way similar to the pseudo code of the arXiv paper).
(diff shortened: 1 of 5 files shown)


josef-widder:
@konnov is going to do an in-depth verification effort of the TLA+ spec in the beginning of April. This PR, and subsequent planning work (ADR) should not be blocked by that.

spec/consensus/proposer-based-timestamp/pbts_001_draft.md
```diff
@@ -0,0 +1,268 @@
...
+1. **Aggregate signatures.** The `precommit` messages free of time, which **allows** for aggregate signatures.
+
+### Protocol Overview
+
+#### Proposed Time
+We assume that the field `proposal` in the `PROPOSE` message is a pair `(v, time)`, of the proposed consensus value `v` and the proposed time `time`. 
+
+#### Reception Step
+In the reception step at node `p` at local time `now_p`, upon receiving a message `m`:
+- **if** if the message `m` is of type `PROPOSE` and satisfies `now_p - PRECISION <  m.time < now_p + PRECISION + MSGDELAY`, then mark the message as `timely`.
```
alexanderbez, line 57:
```suggestion
- **if** the message `m` is of type `PROPOSE` and satisfies `now_p - PRECISION <  m.time < now_p + PRECISION + MSGDELAY`, then mark the message as `timely`.  
```

alexanderbez, line 57: Should this not be `now_p - PRECISION <= m.time <= ...`?

spec/consensus/proposer-based-timestamp/pbts_001_draft.md
```diff
@@ -0,0 +1,268 @@
...
+### Protocol Overview
+
+#### Proposed Time
+We assume that the field `proposal` in the `PROPOSE` message is a pair `(v, time)`, of the proposed consensus value `v` and the proposed time `time`. 
+
+#### Reception Step
+In the reception step at node `p` at local time `now_p`, upon receiving a message `m`:
+- **if** if the message `m` is of type `PROPOSE` and satisfies `now_p - PRECISION <  m.time < now_p + PRECISION + MSGDELAY`, then mark the message as `timely`.  
+(`PRECISION` and `MSGDELAY` being system parameters, see [below](#safety-and-liveness))
+> after the presentation in the dev session, we realized that a different semantics for the reception step is closer aligned to the implementation. Instead of dropping propose messages, we keep all them, and mark timely ones.
```
alexanderbez, line 59:
```suggestion
> after the presentation in the dev session, we realized that different semantics for the reception step is closer aligned to the implementation. Instead of dropping propose messages, we keep all of them, and mark timely ones.
```

Can you elaborate a bit one this?

spec/consensus/proposer-based-timestamp/pbts_001_draft.md
```diff
@@ -0,0 +1,268 @@
...
+
+<td>
+
+```go
+function StartRound(round) {
+ round_p ← round
+ step_p ← propose
+ if proposer(h_p, round_p) = p {
+  // new wait condition
+  wait until now_p > block time of block h_p - 1
```
alexanderbez, line 106: This is just to ensure we do not create a proposal with a smaller time than the last block?

spec/consensus/proposer-based-timestamp/pbts_001_draft.md
```diff
@@ -0,0 +1,268 @@
...
+- Other rules are extended in a similar way, or remain unchanged
+
+### Property Overview
+
+#### Safety and Liveness
+
+For safety (Point 1, Point 2, Point 3i) and liveness (Point 4) we need
+the following assumptions:
+
+- There exists a system parameter `PRECISION` such that for any two correct validators `V` and `W`, and at any real-time `t`, their local times `C_V(t)` and `C_W(t)` differ by less than `PRECISION` time units,
```
alexanderbez, line 231: Why not at most `PRECISION` time units?

spec/consensus/proposer-based-timestamp/pbts_001_draft.md
```diff
@@ -0,0 +1,268 @@
...
+For safety (Point 1, Point 2, Point 3i) and liveness (Point 4) we need
+the following assumptions:
+
+- There exists a system parameter `PRECISION` such that for any two correct validators `V` and `W`, and at any real-time `t`, their local times `C_V(t)` and `C_W(t)` differ by less than `PRECISION` time units, 
+i.e., `|C_V(t) - C_W(t)| < PRECISION`
+- The message end-to-end delay between a correct proposer and a correct validator (for `PROPOSE` messages) is less than `MSGDELAY`.
+
+#### Relation to Real-Time
+
+For analyzing real-time safety (Point 5), we use a system parameter `ACCURACY`, such that for all real-times `t` and all correct validators `V`, we have `| C_V(t) - t | < ACCURACY`.
```
alexanderbez, line 237: How and where is `ACCURACY` used? Or is it just a metric we can observe within the system?

spec/consensus/proposer-based-timestamp/pbts_001_draft.md
```diff
@@ -0,0 +1,268 @@
...
+2. **Effect of faulty validators.** If more than `1/2` of the voting power (which is in fact more than one third and less than two thirds of the voting power) is held by faulty validators, then the time is under total control of the faulty validators. (This is particularly challenging in the context of [lightclient][lcspec] security.)
+3. **Proposer influence on block time.** The proposer of the next block has a degree of freedom in choosing the `bfttime`, since it computes the median time based on the timestamps from `precommit` messages sent by 
+`2f + 1` correct validators.
+   1. If there are `n` different timestamps in the  `precommit` messages, the proposer can use any subset of timestamps that add up to `2f + 1` 
+   of the voting power in order to compute the median.
+   1. If the validators decide in different rounds, the proposer can decide on which round the median computation is based.
```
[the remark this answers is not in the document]
josef-widder: It is not. We just wanted to make this explicit in the write-up, as we were not sure people are aware of that (I was not). 

When analyzing the new solution we saw that it had that property, and we were concerned that we introduced uncertainty, but then realized it was already there in the old solution.

spec/consensus/proposer-based-timestamp/pbts_001_draft.md
josef-widder: @istoilkovska, I guess we should check in the TLA+ spec whether we have `<` or `<=`. For safety `<` is more conservative and safe, but for the precise liveness requirements we should double-check

spec/consensus/proposer-based-timestamp/pbts_001_draft.md
josef-widder: Yes.

spec/consensus/proposer-based-timestamp/pbts_001_draft.md
josef-widder: I think this is related to your comment above related to `now_p - PRECISION <= m.time <= ...?`: we either should have strict inequalities in both places or `<=` in both placed. 

Is there a reason you would prefer `<=`?

spec/consensus/proposer-based-timestamp/pbts_001_draft.md
josef-widder: It does not show up in the code, but if there is a value of `ACCURACY` that satisfied `| C_V(t) - t | < ACCURACY`, we can say something about the relation of block-time and real-time. This can be used when observing the chain, e.g., to understand how old a block is.

spec/consensus/proposer-based-timestamp/pbts_001_draft.md
josef-widder: After reading a comment below, I remembered that the strict inequality derives from the strict inequality used in the definition of `PRECISION`. We might change both if this seems more desirable...

spec/consensus/proposer-based-timestamp/pbts_001_draft.md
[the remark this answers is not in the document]
alexanderbez: Thanks for the clarification!

spec/consensus/proposer-based-timestamp/pbts_001_draft.md
alexanderbez: It just seems more natural to me based on my understanding of `PRECISION` and how it acts as a bound for clock skew. In the end, maybe it does not matter.

spec/consensus/proposer-based-timestamp/pbts-algorithm_001_draft.md
```diff
@@ -0,0 +1,136 @@
+# Proposer-Based Time - Part II
+
+## Updated Consensus Algorithm
+
+### Outline
+
+The algorithm in the arXiv paper evaluates rules of the received messages without making explicit how these messages are received. In our solution, we will make some message filtering explicit. We will assume that there are message reception steps (where messages are received and possibly stored locally for later evaluation of rules) and processing steps (the latter roughly as described in a way similar to the pseudo code of the arXiv paper).
```
marbar3778, line 7:
```suggestion
The algorithm in the [arXiv paper](arXiv) evaluates rules of the received messages without making explicit how these messages are received. In our solution, we will make some message filtering explicit. We will assume that there are message reception steps (where messages are received and possibly stored locally for later evaluation of rules) and processing steps (the latter roughly as described in a way similar to the pseudo code of the arXiv paper).
```

not sure if this is right, but it would be nice to have the link here as well.

cmwaters (commented):
I have a few questions to help my understanding of this algorithm.

1. What happens if 2/3+ precommit on a block that has a timestamp that a node deems to be untimely. Does it a) still commit it or b) panic
2. Can a proposer in round 2 propose a block with a time less than a block proposed in round 1 (I'm guessing this is fine so long as it is still timely). 
3. Should there be any relation between timeout propose and PRECISION since a proposer waiting for the time to be ready to send out their proposal could timeout and thus miss their block (or if PRECISION is too relaxed a proposer could set their block time well in the future, where the proposer of the next height will be unable to propose a block before the timeout). 
**EDIT**
4. What is the validation like for nodes that are still catching up and are thus validating blocks "from the past". Do they just ignore time (except maybe for being greater than the time of the previous block)

spec/consensus/proposer-based-timestamp/pbts-algorithm_001_draft.md
```diff
@@ -0,0 +1,136 @@
+# Proposer-Based Time - Part II
+
+## Updated Consensus Algorithm
+
+### Outline
+
+The algorithm in the [arXiv paper](arXiv) evaluates rules of the received messages without making explicit how these messages are received. In our solution, we will make some message filtering explicit. We will assume that there are message reception steps (where messages are received and possibly stored locally for later evaluation of rules) and processing steps (the latter roughly as described in a way similar to the pseudo code of the arXiv paper).
```
josef-widder, line 7:
```suggestion
The algorithm in the [arXiv paper][arXiv] evaluates rules of the received messages without making explicit how these messages are received. In our solution, we will make some message filtering explicit. We will assume that there are message reception steps (where messages are received and possibly stored locally for later evaluation of rules) and processing steps (the latter roughly as described in a way similar to the pseudo code of the arXiv paper).
```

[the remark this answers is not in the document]
josef-widder: Ah, thanks! I replaces parentheses by brackets. That solves it.

spec/consensus/proposer-based-timestamp/pbts_001_draft.md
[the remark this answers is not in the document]
josef-widder: Yes, if +2/3 precommit in round 1 but timeout before they receive these precommits (that is, before they decide) they go to the next round. In round 2 they might precommit again. By agreement it will be the same block, but the times in their precommits will be different.

So you have two rounds with +2/3 precommits. Both are valid commit for the block. The proposer can pick which one it prefers.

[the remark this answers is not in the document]
cmwaters: Yup thanks

spec/consensus/proposer-based-timestamp/pbts_001_draft.md
```diff
@@ -0,0 +1,268 @@
...
+### Property Overview
+
+#### Safety and Liveness
+
+For safety (Point 1, Point 2, Point 3i) and liveness (Point 4) we need
+the following assumptions:
+
+- There exists a system parameter `PRECISION` such that for any two correct validators `V` and `W`, and at any real-time `t`, their local times `C_V(t)` and `C_W(t)` differ by less than `PRECISION` time units, 
+i.e., `|C_V(t) - C_W(t)| < PRECISION`
+- The message end-to-end delay between a correct proposer and a correct validator (for `PROPOSE` messages) is less than `MSGDELAY`.
```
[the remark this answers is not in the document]
istoilkovska: For the analysis, there is some freedom in choosing `MSGDELAY`, but in principle `MSGDELAY = ∆` is the obvious choice.

spec/consensus/proposer-based-timestamp/pbts_001_draft.md
```diff
@@ -0,0 +1,268 @@
...
+1. **Aggregate signatures.** The `precommit` messages free of time, which **allows** for aggregate signatures.
+
+### Protocol Overview
+
+#### Proposed Time
+We assume that the field `proposal` in the `PROPOSE` message is a pair `(v, time)`, of the proposed consensus value `v` and the proposed time `time`. 
+
+#### Reception Step
+In the reception step at node `p` at local time `now_p`, upon receiving a message `m`:
+- **if** the message `m` is of type `PROPOSE` and satisfies `now_p - PRECISION <  m.time < now_p + PRECISION + MSGDELAY`, then mark the message as `timely`.
```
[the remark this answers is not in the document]
istoilkovska: We formalize the conditions under which liveness holds in the [LIVENESS](https://github.com/tendermint/spec/blob/ilina/proposer-based-timestamp/spec/consensus/proposer-based-timestamp/pbts-sysmodel_001_draft.md#liveness) section. 

To give an easy answer, if >1/3 of the validators are not in sync, there are no guarantees. (Liveness may still be preserved if >1/3 of the validators are out of sync, but the messages are delivered very fast.)

istoilkovska:
>     1. What happens if 2/3+ precommit on a block that has a timestamp that a node deems to be untimely. Does it a) still commit it or b) panic

Short answer: a). Long answer: Following the discussion after the presentation, we changed the protocol a bit. Rather than dropping `PROPOSE` messages when they are untimely, we store them and record whether they have been received on time or not. The rule to decide on a block is then based on receiving any `PROPOSE` message, regardless if it was on time or not. 

>     2. Can a proposer in round 2 propose a block with a time less than a block proposed in round 1 (I'm guessing this is fine so long as it is still timely).

Yes. Still, these two times are greater than the time of the last block.  
 
>     3. Should there be any relation between timeout propose and PRECISION since a proposer waiting for the time to be ready to send out their proposal could timeout and thus miss their block (or if PRECISION is too relaxed a proposer could set their block time well in the future, where the proposer of the next height will be unable to propose a block before the timeout).

You're right, we overlooked this. We will need `timeoutPropose(r)>2∆+timeoutPrecommit(r−1) + x`, where `x` is the maximum difference between the last block time and the local clocks. We will look at the exact value of `x` more closely next week. 

>     4. What is the validation like for nodes that are still catching up and are thus validating blocks "from the past". Do they just ignore time (except maybe for being greater than the time of the previous block)

As far as we understand, currently the nodes that do validation check BFT time against the median of the commits in the same block. Analogously, the proposer-based timestamp is checked by checking that all nodes that signed the commit also signed the proposer-based timestamp.

spec/consensus/proposer-based-timestamp/pbts_001_draft.md
istoilkovska: The TLA+ spec is consistent with the English spec w.r.t. the inequalities. Since there are no strong reasons to have non-strict inequalities, we should keep the strict ones.

spec/consensus/proposer-based-timestamp/pbts-algorithm_001_draft.md
```diff
@@ -0,0 +1,136 @@
...
+```
+
+#### New Rule Replacing Lines 22 - 27
+
+- a validator prevotes for the consensus value `v` **and** the time `t`
+- the code changes as the `PROPOSAL` message carries time (while `lockedValue` does not)
+
+#### **[PBTS-ALG-UPON-PROP.0]**
+```go
+upon timely(⟨PROPOSAL, h_p, round_p, (v,t), −1⟩) from proposer(h_p, round_p) while step_p = propose do {
```
ValarDragon, line 57: What is this timely function?

josef-widder: In the [reception step](https://github.com/tendermint/spec/blob/ilina/proposer-based-timestamp/spec/consensus/proposer-based-timestamp/pbts_001_draft.md#reception-step) we mark a `propose` message as timely if it arrives in a given interval. This just expresses that this was the case for that message. 

Does that help?  Do you think we should write that in a more clear way?

ValarDragon: That helps! I think I just missed the definition in reading the raw markdown, its pretty clear in the rendered version =)

spec/consensus/proposer-based-timestamp/pbts-algorithm_001_draft.md
```diff
@@ -0,0 +1,136 @@
...
+```go
+upon timely(⟨PROPOSAL, h_p, round_p, (v, tprop), vr⟩) from proposer(h_p, round_p) AND 2f + 1 ⟨PREVOTE, h_p, vr, id((v, tvote)⟩ 
+while step_p = propose ∧ (vr ≥ 0 ∧ vr < round_p) do {
+  if valid(v) ∧ (lockedRound_p ≤ vr ∨ lockedValue_p = v) {
+    broadcast ⟨PREVOTE, h_p, roundp, id(v, tprop)⟩
+  }
+  else {
+    broadcast ⟨PREVOTE, hp, roundp, nil⟩
+  }
+}
```
ValarDragon, line 85: I believe this is missing the line `  step_p ← prevote`

spec/consensus/proposer-based-timestamp/pbts_001_draft.md
```diff
@@ -0,0 +1,268 @@
...
+- Other rules are extended in a similar way, or remain unchanged
+
+### Property Overview
+
+#### Safety and Liveness
+
+For safety (Point 1, Point 2, Point 3i) and liveness (Point 4) we need
+the following assumptions:
+
+- There exists a system parameter `PRECISION` such that for any two correct validators `V` and `W`, and at any real-time `t`, their local times `C_V(t)` and `C_W(t)` differ by less than `PRECISION` time units,
```
cmwaters, line 231: When you say system parameter, does this imply that `PRECISION` is a consensus parameter that is agreed upon in consensus or an individual node parameter?

cmwaters: Same q for `ACCURACY`

spec/consensus/proposer-based-timestamp/pbts_001_draft.md
cmwaters, line 57: There has been some talk about giving the application some degree of control over the block times (https://github.com/tendermint/tendermint/issues/5911). Such control might be valuable when some transactions / loads require more time to process. In the case this idea was to come to any fruition would we redefine `timely` as:
```
now_p - PRECISION + BLOCKDELAY < m.time < now_p + PRECISION + MSGDELAY + BLOCKDELAY
```

cmwaters (approved):
This all looks good at least from my angle 👍 . I see that there could be a couple of small things that may need to be ironed out but IMO I think this spec is well thought out. My question is how do we define when this is ready to be merged and what steps need to be taken following the merge of this PR.

spec/consensus/proposer-based-timestamp/pbts_001_draft.md
josef-widder: `PRECISION` needs to be agreed upon in consensus, as it will be part of the code. Moreover, in our analysis we assume it is the same on every node.

`ACCURACY` is not visible in the code. But it is crucial to relate `blocktime` to `real-time`. It required, e.g., by a lightclient to understand how old a block really is. In principle here a lightclient may subjectively say, "I assume accuracy is 10 sec". Then its safety depends on the correctness of this assumption. It would therefore be good if the validators would provide an estimate how large their `ACCURACY` is.

spec/consensus/proposer-based-timestamp/pbts_001_draft.md
josef-widder: I don't think it interferes here. 

Rather it interferes with the `TimeoutPropose`: I guess the cleanest way to do this is that the proposer waits until the `BLOCKDELAY` has expired before sending the `PROPOSE` message.

In https://github.com/tendermint/spec/issues/265#issuecomment-803976562 I have estimated how `TimeoutPropose` needs to be computed to ensure monotonic block times, that is 
- `block(i+1).Time > block(i).Time`

This would need to be generalized to
- `block(i+1).Time > block(i).Time + BLOCKDELAY`

That is to compute `T2` in this case, we would need to add `BLOCKDELAY`. I guess we can already add a comment in the spec right now to be ready for such a change. What do you think?

cmwaters: Thanks. Yeah this makes sense. If you want, you can mention it but I wouldn't say it's that the idea is well developed enough to warrant any changes to the math

spec/consensus/proposer-based-timestamp/pbts_001_draft.md
cmwaters: Right. We could also persist `real-time` values for each node when they store their blocks and add this as part of an RPC endpoint. Thus, participants of the network could query nodes for their `ACCURACY`. Would doing something like that makes sense?

josef-widder: I am not sure that would help. 

But you are right: there is a bigger question: how are validators synchronizing their clocks now (I guess they use NTP) and what values for `ACCURACY` can we assume from there.

Or even more generally: right now there are a lot of implicit assumptions about time and clocks. We need to make them more explicit (as we tried in this PR). And once we made these assumptions explicit, what are suitable values for the parameters. For instance, the timeout values of Tendermint consensus implementation work pretty well. I wonder how these values where set/estimated initially. A similar approach we should take for the parameters introduced in this PR.

Also, I think we can be quite conservative here. Even if the precision/accuracy of the clocks with NTP is in the order of several hundred microseconds, perhaps it is enough to set the parameters in the code to 10 seconds. After all, given that unbonding and trusting periods are in the order of weeks, we just need to maintain some relation of real-time to block time.

The point about these parameters are not to be as precise as possible, but to understand "how precise" the solution guarantees to be in order to have sound guarantees at the system interfaces, e.g., between full nodes and light nodes. The point is to guarantee some value of precision that is sufficient for the application purpose rather to expose the actual precision that the system could achieve.

spec/consensus/proposer-based-timestamp/pbts_001_draft.md
istoilkovska: We have added a comment in the spec about how `BLOCKDELAY` can be incorporated in the future.

PR #266, opened 2021-03-17
Fix list of RFCs

ebuchman:


ValarDragon (approved):
LGTM

I'm surprised this wasn't caught by the markdown linter

josef-widder (approved):
thanks for fixing!

PR #269, opened 2021-03-22
core/types: header hashing

### LINKED ISSUE #175: Cleanup Header Hashing
The way we hash the header is a bit crazy right now. Each field is stuffed into a struct so it can be proto encoded, and then those proto encodings are put in Merkle tree.

I suppose the extra struct wrapping is an artifact of needing a struct to call Marshal on in Go. It seems for instance in Rust we can actually get a proto encoding of a type (eg. bytes, string, int) without wrapping it in a struct first.

At the least we need to do a better job of documenting this weirdness in the spec. Currently we have a short reference in https://github.com/tendermint/spec/blob/master/spec/core/encoding.md#merkleroot about how fields are amino encoded first, so this needs to be updated to clarify that they are wrapped in a struct and then Proto encoded.

But ultimately I would question this whole complex hashing enterprise in the first place. It's not clear that any one gets any value from merkle proofs of fields in the header, so we might save a lot of trouble by just serializing the header as a whole object and hashing it once. 

Somewhat related https://github.com/tendermint/tendermint/issues/2530

marbar3778:
## Description

Cleanup header hashing in favor of proto encoding the struct and hashing the output.

closes #175

### DIFF
--- a/spec/core/data_structures.md
+++ b/spec/core/data_structures.md
@@ -125,6 +125,18 @@ the data in the current block, the previous block, and the results returned by t
 | EvidenceHash      | slice of bytes (`[]byte`) | MerkleRoot of the evidence of Byzantine behaviour included in this block.                                                                                                                                                                                                                                                                                                             | Must  be of length 32                                                                                                                                                                            |
 | ProposerAddress   | slice of bytes (`[]byte`) | Address of the original proposer of the block. Validator must be in the current validatorSet.                                                                                                                                                                                                                                                                                         | Must  be of length 20                                                                                                                                                                            |
 
+### Header Hashing
+
+The hash of the header is used to identify the block a node is downloading or verifying against. To get the correct representation of the header in hash form you must protobuf encode the entire struct, then sha256 the resulting encoding. The field order of the header matters, please see [header struct](../../proto/types/types.proto) for more information on how to construct the header.
+
+```go
+  bz := proto.Marshal(Header)
+
+  sha256.New()
+  h := sha256.Sum256(bz)
+
+```
+
 ## Version
 
 | Name  | type   | Description                                                                                                     | Validation                                                                                                         |


cmwaters (approved):

marbar3778:
@ebuchman @milosevic, could I get an approval from one of you

ebuchman:
Hey @marbar3778 this is a proposed breaking change to the Block Protocol? It should have a referencable RFC or ideally CIP now

marbar3778:
> Hey @marbar3778 this is a proposed breaking change to the Block Protocol? It should have a referencable RFC or ideally CIP now

How we have RFC's setup is that if there is not an understanding or agreement on the change, an RFC would help explain things related to it. This change was suggested by many people so I went with there is an understanding and agreement with the approach. As for CIP, I can assist in writing a CIP for the hub. Who on the Gaia team is responsible for this?

marbar3778:
> Also, how do we manage versions for this repo. If this gets merged here before the change in Tendermint, won't ppl reading the spec be misled? Should we start versioning the spec and even including a changelog? Though ideally a changelog would in the end be a list of CIPs added

Version 0.1 should be cut soon. Not sure what procedure is wanted to do this. There are a couple PRs that would be useful to merge before this tho. A PR that falls into this category is https://github.com/tendermint/spec/pull/276, but its a nice to have.

ValarDragon:
Why are headers protobuf encoded btw, are all fields of fixed length? If not, won't this be hard to parse on other chains?

marbar3778:
> Why are headers protobuf encoded btw, are all fields of fixed length? If not, won't this be hard to parse on other chains?

They were previously amino encoded, this was to keep similar design to the previous one. I don't think there is a valid reason.

liamsi:
> They were previously amino encoded, this was to keep similar design to the previous one. I don't think there is a valid reason.

@marbar3778 I think the fixed length encoding is a good thing independent from proto/amino (and if I understand @ValarDragon's comment correctly he is also advocating for that to make the header easily parseable by other chains). If all fields are fixed size, encoding and decoding is simple: just the raw bytes (and documenting the endianness for integers).

PR #271, opened 2021-03-22
core/types: merge evidence into data

marbar3778:
## Description:

Merge evidence list into the data field. This reduces one of the hashes in the header.


josef-widder:
Is the reason for this just reducing the size? I wonder whether this mixes conceptually different things. As far as I understand, the data field until now was under the control of the application, and evidence under the control of Tendermint. This merge seems to mix this. For instance, the application will not be able to validate all the information in the data, while this might have been the case until now.

marbar3778:
> Is the reason for this just reducing the size? I wonder whether this mixes conceptually different things. As far as I understand, the data field until now was under the control of the application, and evidence under the control of Tendermint. This merge seems to mix this. For instance, the application will not be able to validate all the information in the data, while this might have been the case until now.

Yes, it reduces the header by 32 bytes. This is not large, but it is a stepping stone. I agreed with you at first in regard to two different things being housed under one struct. The reason I changed my thinking is that evidence, while created by tendermint, is also an application concern. Tendermint creates the evidence, but the application handles it. The same could be said for transactions, they are externally created and only handled by the application. A second reason why I decided to go with this approach is that evidence and txs are sent to the app, it is better to group them together since they share the same path and final destination.

josef-widder:
There are still other hashes in the header. What is the rationale of keeping some of them separated and putting others together?

marbar3778:
I was planning on using this as a stepping stone to https://github.com/tendermint/spec/issues/199. I can do them together if you think it's better?

josef-widder:
I just tried to understand the rationale. The link to #199 was helpful to get some context. Thanks!

Also I try to understand how we make such decisions. For instance, in #199 it is proposed that `ValidatorsHash` is removed, while afaik the current IBC light client used that field.

marbar3778:
> Also I try to understand how we make such decisions. For instance, in #199 it is proposed that `ValidatorsHash` is removed, while afaik the current IBC light client used that field.

What is removed needs to be reevaluated since the issue was opened so long ago. I was planning on making a proposal taking into consideration the new uses of the hashes.

marbar3778:
this will be merged after there is a release on this repo.

PR #275, opened 2021-03-25
genesis: remove app_hash 

marbar3778:
## Description

Remove app_hash from genesis as it is not used and returned by the app on initchain.


cmwaters (approved):
Yup this makes sense to me 👍

josef-widder:
So that was once used and is not used anymore, or what is the reason it was there in the first place?

marbar3778:
I looked back through some issues and couldn't find why it was there. Maybe @ebuchman could shed some light. I mainly know its not used now.

marbar3778:
@milosevic do you have thoughts here?

marbar3778:
@ebuchman could I get your approval here?

PR #283, opened 2021-04-07
ABCI++ markdown integration into Tendermint (for spec)

ValarDragon:
This PR adds a markdown version of how ABCI++ aims to hook into Tendermint from a spec perspective, based on a prior discussion with Igor, Josef and Zarko.

The structure of the code is as follows:

v0.md contains my transcription of the Tendermint BFT paper into markdown directly.
v1.md contains how I think ABCI hooks in.
v2.md contains a conversion to single-threaded ABCI++, with no optimization for splitting up verification of the header and block data
v3.md contains the optimization for splitting up verification of the header and block data 
v4.md contains the multi-threaded optimization for delaying the end of ProcessProposal.

The structure for reviewing this is to review it commit by commit. Each commit shows the update to one file, so that it may be incrementally reviewed. (As a side effect, it also involves copying the file as well).

So the v0 commit contains the code for v0, and a duplicate called v1.md.
The v1 commit contains the diff on v1.md for the change, and a duplicate of v1.md called v2.md.
The v2 commit contains the diff on v2.md for its change, and a duplciate called v3.md, etc.

~~Left as a draft until the complete v3.md and v4.md are included.~~


spec/abci++/v0.md
```diff
@@ -0,0 +1,161 @@
...
+```go
+upon ⟨PROPOSAL, h_p, round_p, v, vr⟩
+ from proposer(h_p, round_p)
+ AND 2f + 1 ⟨PREVOTE, h_p, vr, id(v)⟩ 
+while step_p = propose ∧ (vr ≥ 0 ∧ vr < round_p) do {
+ if valid(v) ∧ (lockedRound_p ≤ vr ∨ lockedValue_p = v) {
+  broadcast ⟨PREVOTE, h_p, round_p, id(v)⟩
+ } else {
+  broadcast ⟨PREVOTE, h_p, round_p, nil⟩
+ }
```
liamsi, line 65: Just a minor/optional-to-address nit: sometimes the `else` goes into a separate line and sometimes not. I don't have any preference but would make it consistent as it is distracting otherwise.

spec/abci++/v0.md
```diff
@@ -0,0 +1,161 @@
...
+```go
+function startRound(round) {
+    round_p ← round
+    step_p ← propose
+    if proposer(h_p, round_p) = p {
+        if validValue_p != nil {
+            proposal ← validValue_p
+        }
+        else {
+            proposal ← getValue()
```
liamsi, line 28: Again, orthogonal to this PR:
The same holds for the bft paper but `getvalue()` is not introduced or explained anywhere. This if-else-block could potenially be simplified / remove by explaining getValue.

liamsi (commented):
Left some nits on v0.md will go through the others shortly.

spec/abci++/v1.md
```diff
@@ -0,0 +1,166 @@
...
+while decision_p[h_p] = nil do {
+ if (valid(v)) {
+	 decision_p[h_p] = v
+	 h_p = h_p + 1
+	 reset lockedRound_p, lockedValue_p,validRound_p and validValue_p to initial values
+	 lockedRound_p ← round_p
+   ABCI.BeginBlock()
+   ABCI.DeliverTxs()
+   ABCI.EndBlock()
+	 StartRound(0)
```
liamsi, line 161: There seems to be an indentation glitch here. Other than that, this looks good imo.

spec/abci++/v3.md
```diff
@@ -0,0 +1,202 @@
...
+function startRound(round) {
+    round_p ← round
+    step_p ← propose
+    if proposer(h_p, round_p) = p {
+        if validValue_p != nil {
+            proposal ← validValue_p
+        }
+        else {
+            txdata ← mempool.GetBlock()
+            // getBlockProposal fills in header
```
liamsi, line 31: Comment outdated.

spec/abci++/v3.md
```diff
@@ -0,0 +1,202 @@
...
+            unpreparedProposal = getUnpreparedBlockProposal(blockdata)
```
liamsi, line 32: Shouldn't this be txdata instead?
```suggestion
            unpreparedProposal = getUnpreparedBlockProposal(txdata)
```

spec/abci++/v3.md
```diff
@@ -0,0 +1,202 @@
...
+upon ⟨PROPOSAL, h_p, round_p, v, -1⟩
+ from proposer(h_p, round_p)
+ AND 2f + 1 ⟨ PRECOMMIT, h_p, vr, id(v)⟩ 
+while decision_p[h_p] = nil do {
+ if (valid(v)) {
+	 decision_p[h_p] = v
+	 h_p = h_p + 1
+	 reset lockedRound_p, lockedValue_p,validRound_p and validValue_p to initial values
+	 lockedRound_p ← round_p
+   ABCI.FinalizeBlock()
```
liamsi, line 196: Again, an indentation glitch. Or is it on purpose that the ABCI calls are idented differently? 🤔

spec/abci++/v3.md
```diff
@@ -0,0 +1,202 @@
...
+	schedule OnTimeoutPrevote(h_p, round_p) to be executed after timeoutPrevote(round_p)
+}
+```
+
+with OnTimeoutPrevote defined as:
+
+```go
+def OnTimeoutPrevote(height, round) {
+	if (height = h_p && round = round_p && step_p = prevote) {
+    precommit_extension = ABCI.ExtendVote(h_p, round_p, nil)
```
liamsi, line 117: indentation

spec/abci++/v3.md
```diff
@@ -0,0 +1,202 @@
...
+upon 2f + 1 ⟨PREVOTE, h_p, vr, -1⟩ 
+ with step_p = prevote for the first time, do {
+	schedule OnTimeoutPrevote(h_p, round_p) to be executed after timeoutPrevote(round_p)
+}
+```
+
+with OnTimeoutPrevote defined as:
+
+```go
+def OnTimeoutPrevote(height, round) {
```
liamsi, line 115: Why not use `function` or `func` everywhere? Or `def` everywhere (I prefer `function`).

spec/abci++/v3.md
```diff
@@ -0,0 +1,202 @@
...
+```go
+function startRound(round) {
+    round_p ← round
+    step_p ← propose
+    if proposer(h_p, round_p) = p {
+        if validValue_p != nil {
+            proposal ← validValue_p
+        }
+        else {
+            txdata ← mempool.GetBlock()
```
liamsi, line 30: Speaking about the mempool, I think it would be great if the mempool connection was also mentioned soemwhere (not necessarily in the pseudo code and just for the sake of completeness).

spec/abci++/v1.md
ValarDragon: oh huh, I think my editor wasn't showing tabs. I was actually overall pretty confused with what was going on with indentation locally before haha. (Alot of the markdown came from the great work in #261 )

spec/abci++/v3.md
ValarDragon: Thanks for the catch!

spec/abci++/v3.md
ValarDragon: You're right, thanks! I'll switch both to be txData

spec/abci++/v3.md
ValarDragon: Agreed, I'll switch to function everywhere

spec/abci++/v0.md
liamsi: OK, that makes sense.

spec/abci++/v0.md
```diff
@@ -0,0 +1,161 @@
...
+### Initialization
+
+```
+h_p = 0
+round_p = 0
+step_p is one of {propose, prevote, precommit}
+decision_p = nil
+lockedValue_p = nil
+validValue_p = nil
+validRound_p = -1
```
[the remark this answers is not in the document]
ValarDragon: Hrmm, cc @milosevic @konnov @josef-widder for thoughts on this. 

Its not super clear to me, since I think removing it may make it easier to conclude that when height / round increments, it increments for every validator.

(Its also not entirely clear to me, how / if ever this markdown will get merged)

spec/abci++/v0.md
```diff
@@ -0,0 +1,156 @@
...
+
+This translates the latex code for Tendermint consensus from the Tendermint paper into markdown.
+
+### Initialization
+
+```go
+h_p ← 0
+round_p ← 0
+step_p is one of {propose, prevote, precommit}
+decision_p ← nil
```
konnov, line 11: In the Tendermint paper, `decision_p` is a dynamically growing vector of values. I guess, that is why the authors wrote `decision_p[] := nil` in the paper. Can we write:

```
decision_p ← Vector()
```

Or `Array()`, or `List()`. I am fine with any choice here.

spec/abci++/v0.md
```diff
@@ -0,0 +1,156 @@
...
+```
+
+### ReceiveProposal
+
+In the case where the local node is not locked on any round, the following is ran:
+
+```go
+upon ⟨PROPOSAL, h_p, round_p, v, −1) from proposer(h_p, round_p) while step_p = propose do {
+    if valid(v) ∧ (lockedRound_p = −1 ∨ lockedValue_p = v) {
+        broadcast ⟨PREVOTE, h_p, round_p, id(v,t)⟩
```
konnov, line 43: What is `t` in `id(v, t)`? I guess, it is a typo.

spec/abci++/v0.md
```diff
@@ -0,0 +1,156 @@
...
+    }
+}
+```
+
+### Receiving enough prevotes to precommit
+
+The following code is ran upon receiving 2f + 1 prevotes for the same block
+
+```go
+upon ⟨PROPOSAL, h_p, round_p, v, -1⟩
```
konnov, line 92:
```suggestion
upon ⟨PROPOSAL, h_p, round_p, v, *⟩
```

That's what I see in line 36 in the paper.

spec/abci++/v0.md
```diff
@@ -0,0 +1,156 @@
...
+    step_p ← precommit
+}
+```
+
+### Precommit timeout
+
+Upon receiving 2f + 1 precommits, setup a timeout.
+
+```go
+upon 2f + 1 ⟨PRECOMMIT, h_p, vr, -1⟩ for the first time, do {
```
konnov, line 122:
```suggestion
upon 2f + 1 ⟨PRECOMMIT, h_p, round_p, *⟩ for the first time, do {
```

In the paper, they have a star to indicate that the value of `validRound` is irrelevant.

spec/abci++/v0.md
```diff
@@ -0,0 +1,156 @@
...
+    schedule OnTimeoutPrecommit(h_p, round_p) to be executed after timeoutPrecommit(round_p)
+}
+```
+
+with OnTimeoutPrecommit defined as:
+
+```go
+function OnTimeoutPrecommit(height, round) {
+    if (height = h_p && round = round_p && step_p = prevote) {
```
konnov, line 131:
```suggestion
    if (height = h_p && round = round_p) {
```

I don't think a process has to be in `prevote`, in order to progress to the next round. At least, it is not the case in the paper.

spec/abci++/v0.md
```diff
@@ -0,0 +1,156 @@
...
+    }
+}
+```
+
+### Upon Receiving 2f + 1 precommits
+
+The following code is ran upon receiving 2f + 1 precommits for the same block
+
+```go
+upon ⟨PROPOSAL, h_p, round_p, v, -1⟩
```
konnov, line 142:
```suggestion
upon ⟨PROPOSAL, h_p, r, v, *⟩
```

By writing `r`, they do implicit pattern matching. So `r` can be different from `round_p`.

spec/abci++/v0.md
```diff
@@ -0,0 +1,156 @@
...
+  from proposer(h_p, round_p)
```
konnov, line 143:
```suggestion
  from proposer(h_p, r)
```

spec/abci++/v0.md
```diff
@@ -0,0 +1,156 @@
...
+  AND 2f + 1 ⟨ PRECOMMIT, h_p, vr, id(v)⟩
```
konnov, line 144:
```suggestion
  AND 2f + 1 ⟨ PRECOMMIT, h_p, r, id(v)⟩ 
```

spec/abci++/v0.md
```diff
@@ -0,0 +1,156 @@
...
+```go
+upon ⟨PROPOSAL, h_p, round_p, v, -1⟩
+  from proposer(h_p, round_p)
+  AND 2f + 1 ⟨ PRECOMMIT, h_p, vr, id(v)⟩ 
+  while decision_p[h_p] = nil do {
+    if (valid(v)) {
+        decision_p[h_p] ← v
+        h_p ← h_p + 1
+        reset lockedRound_p, lockedValue_p,validRound_p and validValue_p to initial values
+        lockedRound_p ← round_p
```
konnov, line 150:
```suggestion
```

It's not in the paper

spec/abci++/v0.md
```diff
@@ -0,0 +1,156 @@
+# Tendermint v0 Markdown pseudocode
+
+This translates the latex code for Tendermint consensus from the Tendermint paper into markdown.
```
konnov, line 3: The case of `OnTimeoutPropose` is missing.

konnov (commented):
Made a pass over `v0.md`. There are several inconsistencies with the original pseudo-code. I am not sure, whether some of them were intended, or they are just typos.

spec/abci++/v1.md
```diff
@@ -0,0 +1,163 @@
...
+  from proposer(h_p, round_p)
+  AND 2f + 1 ⟨ PRECOMMIT, h_p, vr, id(v)⟩ 
+  while decision_p[h_p] = nil do {
+    if (valid(v)) {
+        decision_p[h_p] ← v
+        h_p ← h_p + 1
+        reset lockedRound_p, lockedValue_p,validRound_p and validValue_p to initial values
+        lockedRound_p ← round_p
+        ABCI.BeginBlock()
+        ABCI.DeliverTxs()
```
konnov, line 156: Shall we retrieve `txdata` from mempool, as in `StartRound`?

spec/abci++/v2.md
```diff
@@ -0,0 +1,180 @@
...
+```go
+function startRound(round) {
+    round_p ← round
+    step_p ← propose
+    if proposer(h_p, round_p) = p {
+        if validValue_p != nil {
+            proposal ← validValue_p
+        } else {
+            txdata ← mempool.GetBlock()
+            // getBlockProposal fills in header
```
konnov, line 29: seems to be an outdated comment from `v1`

spec/abci++/v2.md
```diff
@@ -0,0 +1,180 @@
...
+            unpreparedProposal ← getUnpreparedBlockProposal(txdata)
+            proposal ← ABCI.PrepareProposal(unpreparedProposal)
```
konnov, line 31:
```suggestion
            // ABCI++: the proposer may reorder/update transactions in `unpreparedProposal`
            proposal ← ABCI.PrepareProposal(unpreparedProposal)
```

spec/abci++/v2.md
```diff
@@ -0,0 +1,180 @@
...
+}
+```
+
+### ReceiveProposal
+
+In the case where the local node is not locked on any round, the following is ran:
+
+```go
+upon ⟨PROPOSAL, h_p, round_p, v, −1) from proposer(h_p, round_p) while step_p = propose do {
+    if valid(v) ∧ ABCI.ProcessProposal(h_p, v).accept ∧ (lockedRound_p = −1 ∨ lockedValue_p = v) {
```
konnov, line 46: why do we need both `valid(v)` and `ABCI.ProcessProposal(h_p, v)`?

spec/abci++/v0.md
josef-widder: looks like copy/paster overlook from the proposer-based time

spec/abci++/v2.md
konnov, line 46: Trying to clarify things for myself: Is `v` understood as a block header or a complete block? It would be great to have a comment that explains the expected behavior of `ABCI.ProcessProposal(h_p, v)`.

spec/abci++/v2.md
```diff
@@ -0,0 +1,180 @@
...
+        broadcast ⟨PREVOTE, h_p, round_p, id(v,t)⟩ 
+    } else {
+        broadcast ⟨PREVOTE, h_p, round_p, nil⟩ 
+        // Include any slashing evidence that may be sent in the process proposal response
+        for evidence in ABCI.ProcessProposal(h_p, v).evidence_list {
```
konnov, line 51: what are the assumptions about `ABCI.ProcessProposal(h_p, v).evidence_list`? For instance, if the ABCI endpoint is Byzantine, what guarantees do we have that the code cannot make up the evidence?

spec/abci++/v2.md
```diff
@@ -0,0 +1,180 @@
...
+    schedule OnTimeoutPrevote(h_p, round_p) to be executed after timeoutPrevote(round_p)
+}
+```
+
+with OnTimeoutPrevote defined as:
+
+```go
+function OnTimeoutPrevote(height, round) {
+    if (height = h_p && round = round_p && step_p = prevote) {
+        precommit_extension ← ABCI.ExtendVote(h_p, round_p, nil)
```
konnov, line 95: Could you add an example of `precommit_extension` in a comment?

Also, why would one need to extend the vote on `nil`? Is it a form of piggy-backing?

spec/abci++/v2.md
```diff
@@ -0,0 +1,180 @@
...
+
+```go
+upon ⟨PROPOSAL, h_p, round_p, v, -1⟩
+  from proposer(h_p, round_p)
+  AND 2f + 1 ⟨PREVOTE, h_p, vr, id(v)⟩ 
+  while valid(v) ∧ step_p >= prevote for the first time do {
+    if (step_p = prevote) {
+        lockedValue_p ← v
+        lockedRound_p ← round_p
+        precommit_extension ← ABCI.ExtendVote(h_p, round_p, id(v))
```
konnov, line 114: Same here. An example of `precommit_extension` would be great.

spec/abci++/v2.md
```diff
@@ -0,0 +1,180 @@
...
+  while step_p = prevote do {
+    precommit_extension ← ABCI.ExtendVote(h_p, round_p, nil)
+    broadcast ⟨PRECOMMIT, h_p, round_p, nil, precommit_extension⟩
+    step_p ← precommit
+}
+```
+
+### Upon receiving a precommit
+
+Upon receiving a precommit `precommit`, we ensure that `ABCI.VerifyVoteExtension(precommit.precommit_extension) = true`
```
konnov, line 136: should the call to `ABCI.VerifyVoteExtension` include the height `h_p` and value `v` as parameters?

spec/abci++/v2.md
```diff
@@ -0,0 +1,180 @@
...
+```
+
+### Upon Receiving 2f + 1 precommits
+
+The following code is ran upon receiving 2f + 1 precommits for the same block
+
+```go
+upon ⟨PROPOSAL, h_p, round_p, v, -1⟩
+  from proposer(h_p, round_p)
+  AND 2f + 1 ⟨ PRECOMMIT, h_p, vr, id(v)⟩
```
konnov, line 167: Here the difference between the original pseudo-code and `v0-v2` is crucial. In the original pseudo-code, a process could catch up with a round `r` that is different from `round_p`. I guess, you would like to avoid catch up here, as the application has partially processed the block for `round_p`.

konnov: If it is the case, we need a comment here.

spec/abci++/v2.md
konnov: Also, does `ProcessProposal` interact with mempool?

spec/abci++/v2.md
```diff
@@ -0,0 +1,180 @@
...
+upon ⟨PROPOSAL, h_p, round_p, v, -1⟩
+  from proposer(h_p, round_p)
+  AND 2f + 1 ⟨ PRECOMMIT, h_p, vr, id(v)⟩ 
+  while decision_p[h_p] = nil do {
+    if (valid(v)) {
+        decision_p[h_p] ← v
+        h_p ← h_p + 1
+        reset lockedRound_p, lockedValue_p,validRound_p and validValue_p to initial values
+        lockedRound_p ← round_p
+        ABCI.FinalizeBlock()
```
konnov, line 174: Since the application has processed a block, we have to make sure that `v` is indeed the block that the ABCI layer has processed.

konnov (commented):
Versions 1-2 helped me a lot to see the envisioned changes. Several comments on v1-v2.

spec/abci++/v4.md
```diff
@@ -0,0 +1,201 @@
...
+In the case where the node is locked on a round, the following is ran:
+
+```go
+upon ⟨PROPOSAL, h_p, round_p, v, vr⟩
+  from proposer(h_p, round_p)
+  AND 2f + 1 ⟨PREVOTE, h_p, vr, id(v)⟩ 
+  while step_p = propose ∧ (vr ≥ 0 ∧ vr < round_p) do {
+    if valid(v) ∧ ABCI.VerifyHeader(h_p, v.header) ∧ (lockedRound_p ≤ vr ∨ lockedValue_p = v) {
+        // We fork process proposal into a parallel process
+        Fork ABCI.ProcessProposal(h_p, v)
```
konnov, line 67: Can a single validator run two instances of `ProcessProposal` at the same time? If so, what do we assume about the state of the ABCI app?

spec/abci++/v4.md
```diff
@@ -0,0 +1,201 @@
...
+        lockedRound_p ← round_p
+        processProposalOutput ← Join ABCI.ProcessProposal(h_p, v)
+        // If the proposal is valid precommit as before.
+        // If it was invalid, precommit nil.
+        // Note that ABCI.ProcessProposal(h_p, v).accept is deterministic for all honest nodes.
+        precommit_value ← nil
+        if processProposalOutput.accept {
+            precommit_value ← id(v)
+        }
+        precommit_extension ← ABCI.ExtendVote(h_p, round_p, id(v))
```
konnov, line 125: Should `ExtendVote` also accept `processProposalOutput`?

konnov (commented):
Having read `v4`, I can say that this is a great presentation of the proposed changes to ABCI++. Added a few comments.

spec/abci++/v0.md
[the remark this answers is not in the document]
konnov: I agree that the subscript `_p` is not optimal. However, it is crucial to see who is executing the code. This prefix is the only indication that the code is executed by a process `p`.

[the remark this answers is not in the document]
konnov: Given that some `upon` rules work as pattern matching, it is very important to distinguish the process variables (e.g., `round_p`) from those "temporary" variables (e.g., `vr`, `r`).

spec/abci++/v0.md
ValarDragon: That is what it is, my bad!

spec/abci++/v0.md
ValarDragon: Set to vector for now, but I wonder if for technical correctness we should actually put `Map`, since its indexed by `h_p`, and we don't initialize at 0

spec/abci++/v0.md
ValarDragon: You are right, thanks for the catch!

spec/abci++/v0.md
ValarDragon: You're right!

spec/abci++/v2.md
ValarDragon: `v` here is understood as the complete block in `v2` and `v3`. ProcessProposal does not interact with the mempool, I'll add some more comments around here

ValarDragon: The intention was `valid(v)` was something managed by Tendermint, whereas the ABCI method is something only the app does things with. (In particular, the app wouldn't guarantee any structure that Tendermint would expect)

spec/abci++/v2.md
ValarDragon: It can make up evidence. The state machine will have still have to interpret this evidence whenever it lands in a block. This gossip is just specifying that ABCI will trigger this message to gossiped around the network, and potentially included according to the {unspecified} mempool processes.

spec/abci++/v2.md
ValarDragon: Sure, I'll add an example of `precommit_extension`!

You have a good point, I don't actually foresee a reason to extend a `nil` vote (since that precommit won't make it into a block). 

That would only make sense in a consensus algorithm in which nil precommits made in time still get included in the next block, which doesn't happen at the moment. The only reasons I can imagine for this seem pretty contrived, so I'll remove it

spec/abci++/v4.md
ValarDragon: I _think_ it should be impossible per the spec at the moment for this to happen? (I believe every process proposal process gets joined prior to incrementing rounds locally)

spec/abci++/v2.md
konnov: I think that having `v_header` would help!

spec/abci++/v4.md
konnov: For instance, the last rule "Upon Receiving 2f + 1 precommits" can fire anytime. This would advance the height and start a new round. What happens to the `ProcessProposal` thread that is still running?

I think we should understand how the ABCI app is supposed to work with these requests:
 
 1. The app processes requests one-by-one, that is, only a single `ProcessProposal` is working till it is joined, or
 1. The app maintains potential future states (produced by different transactions) and then it finalizes one of them.

It looks like the pseudo code is written with Scenario 1 in mind. Then we have to be very careful about not sending concurrent requests to the app.

>

spec/abci++/v2.md
konnov: I see. That makes sense.

spec/abci++/v4.md
konnov: When I am saying "the app" here, I really mean the layer below Tendermint. So it is either the app or the ABCI layer resolving concurrency for the app.

spec/abci++/v4.md
```diff
@@ -0,0 +1,200 @@
...
+
+### ReceiveProposal
+
+In the case where the local node is not locked on any round, the following is ran:
+
+```go
+upon ⟨PROPOSAL, h_p, round_p, v, −1) from proposer(h_p, round_p) while step_p = propose do {
+    if valid(v) ∧ ABCI.VerifyHeader(h_p, v.header) ∧ (lockedRound_p = −1 ∨ lockedValue_p = v) {
+        // We fork process proposal into a parallel process
+        Fork ABCI.ProcessProposal(h_p, v)
```
ebuchman, line 49: We discussed in slack but just noting here the concern with this is that we could end up with a POLKA for a block that turns out to be invalid (ie. fails ProcessProposal) and this could have liveness consequences. 

So we need to carefully work through the consequences of this and adjust as necessary or else just suffer the latency cost of completing ProcessProposal before prevoting.

spec/abci++/v2.md
milosevic: Why evidence is being rebroadcasted here?

spec/abci++/v2.md
milosevic: I guess you are suggesting to pass id(v) here?

spec/abci++/v4.md
milosevic: It might be fine prevoting for v without waiting ProcessProposal(v) to return, but we should probably not set locked/valid value if ProcessProposal(v) hasn't returned successfully.  Therefore, we ensure that validValue/lockedValue is never set to invalid block, and this ensures that invalid block is never committed.

spec/abci++/v4.md
```diff
@@ -0,0 +1,200 @@
...
+```go
+upon ⟨PROPOSAL, h_p, r, v, *⟩
+  from proposer(h_p, r)
+  AND 2f + 1 ⟨ PRECOMMIT, h_p, r, id(v)⟩ 
+  while decision_p[h_p] = nil do {
+    if (valid(v)) {
+        decision_p[h_p] ← v
+        h_p ← h_p + 1
+        reset lockedRound_p, lockedValue_p,validRound_p and validValue_p to initial values
+        ABCI.FinalizeBlock()
```
milosevic, line 194: What happens with precommit_extension upon commit?

spec/abci++/v4.md
```diff
@@ -0,0 +1,200 @@
+# Tendermint v3 Markdown pseudocode
```
JayT106, line 1: should be v4?

JayT106: and the updates in the following lines 3~5?

spec/abci++/v4.md
ValarDragon: They are handled by PrepareProposal, the data may be included within the block (either in the header or tx data), or not included at all and just have been there as an intermediate step.

PR #284, opened 2021-04-19
core: update a few sections 

cmwaters:
This PR tidies up a few parts in the `spec/core` section.

- Update the state file, adding information on validator set and consensus params update as well as removing redundant information covered in data structures
- fix the table of contents in `spec/core`
- minor touch ups

### DIFF
--- a/spec/core/data_structures.md
+++ b/spec/core/data_structures.md
@@ -26,10 +26,7 @@ The Tendermint blockchains consists of a short list of data types:
 - [`Validator`](#validator)
 - [`ValidatorSet`](#validatorset)
 - [`Address`](#address)
-- [`ConsensusParams](#consensusparams)
-- [`EvidenceParams`](#evidenceparams)
-- [`ValidatorParams`](#validatorparams)
-- [`VersionParams`](#versionparams)
+- [`ConsensusParams`](#consensusparams)
 
 ## Block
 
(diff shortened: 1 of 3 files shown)


spec/core/data_structures.md
```diff
@@ -26,10 +26,7 @@ The Tendermint blockchains consists of a short list of data types:
 - [`Validator`](#validator)
 - [`ValidatorSet`](#validatorset)
 - [`Address`](#address)
-- [`ConsensusParams](#consensusparams)
-- [`EvidenceParams`](#evidenceparams)
-- [`ValidatorParams`](#validatorparams)
-- [`VersionParams`](#versionparams)
```
marbar3778, line 32: why remove these?

cmwaters: I considered them under the umbrella of `ConsensusParams`. We can add them back again but in that case we are missing `BlockParams`

marbar3778: Yes, block params is missing. I would be in favor of leaving them, technically header, evidence data is under the umbrella of block, but we still have them.

cmwaters: Sure, I will update them

spec/core/data_structures.md
```diff
@@ -446,21 +452,21 @@ func SumTruncated(bz []byte) []byte {
...
 
 | Name          | Type            | Description                                                           | Field Number |
 |---------------|-----------------|-----------------------------------------------------------------------|--------------|
 | pub_key_types | repeated string | List of accepted public key types. Uses same naming as `PubKey.Type`. | 1            |
 
-### VersionParams
+## VersionParams
 
 | Name        | Type   | Description                   | Field Number |
 |-------------|--------|-------------------------------|--------------|
```
cmwaters, line 472: What's the `Proof` below here for. Is it part of `PartSet`?

cmwaters: Sorry I can't comment the part that I'm referring to directly

marbar3778: Yes, its part of the part of PartSet (🤔).

marbar3778 (approved):

spec/core/data_structures.md
cmwaters: Is `PartSet` something that should be in the spec? Do we have a section on block propagation?

PR #286, opened 2021-04-23
Update README.md

ebuchman:


ValarDragon (approved):

ValarDragon:
Tested that the link works for me, though it points to a README that says its deprecated, and just forwards you on elsewhere. Perhaps this should just direct to the final document one should be reading?

konnov (approved):
Thanks for fixing the link!

PR #299, opened 2021-05-11
rpc: add chunked genesis endpoint

tychoish:
reflects the changes in: https://github.com/tendermint/tendermint/pull/6445

### DIFF
--- a/spec/rpc/README.md
+++ b/spec/rpc/README.md
@@ -855,20 +855,21 @@ Get the genesis document in a chunks to support easily transfering larger docume
 
 #### Parameters
 
-- `chunkID`: the index number of the chunk that you wish to fetch.
+- `chunk` (integer): the index number of the chunk that you wish to
+  fetch. These IDs are 0 indexed.
 
 #### Request
 
 ##### HTTP
 
 ```sh
-curl  http://127.0.0.1:26657/genesis_chunked?chunkID=0
+curl  http://127.0.0.1:26657/genesis_chunked?chunk=0
 ```
 
 ##### JSONRPC
 
 ```sh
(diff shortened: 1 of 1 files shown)


spec/rpc/README.md
```diff
@@ -848,6 +849,42 @@ curl -X POST https://localhost:26657 -d "{\"jsonrpc\":\"2.0\",\"id\":1,\"method\
 }
 ```
 
+### GenesisChunked
+
+Get the genesis document in a chunks to support easily transfering larger documents.
+
+#### Parameters
+
+- `chunkID`: the index number of the chunk that you wish to fetch.
```
cmwaters, line 858: Personally, I think just calling it `chunk` would be better, but this is a super minor detail so I don't really mind.

cmwaters, line 858:
```suggestion
- `chunkID (integer)`: the index number of the chunk that you wish to fetch.
```

cmwaters (approved):
LGTM 👍 

If a genesis file is greater than the size of one chunk then calling `/genesis` will return an error right? Do we also want to include a note of this in the `Genesis` section

spec/rpc/README.md
alexanderbez: It should also state if it's one-based or zero-based.

alexanderbez (approved):

tychoish:
I have not changed the behavior of the genesis endpoint, yet, but we could do that.

cmwaters (approved):

marbar3778 (approved):
thanks for making this change

marbar3778:
@tychoish can you fix the linter and then we can merge this

marbar3778:
@tychoish can you fix the linting in this PR. Then we can merge

PR #310, opened 2021-06-28
Add C++ code generation and test scenario

nano-o:
This adds the ability to generate C++ code to "run" the model and therefore test various scenarios. We provide a basic scenarios in which nodes reach a decision, showing that the model is not vacuous. Note that the generated C++ code cannot be run as is in a distributed system; at the moment it's only useful to call event handlers sequentially so as to create one execution of the model.

This also substitutes `require` for `assume` in the model. `assume` does not place requirements on callers (because it's an assumption!). This meant that an issue slipped through: a handler was assuming that some field was -1, when in fact messages were sent with a different field value and could therefore not be received. This did however not impact the safety proof.


josef-widder (approved):
Looks good. Thanks!

PR #312, opened 2021-07-07
PEX: add V2 messages

cmwaters:
Adds the new V2 messages for the PEX service.

We should also document the algorithm somewhere - probably as a separate PR


marbar3778 (approved):
The algorithm seems more implementation specific right?

PR #314, opened 2021-07-09
rfc: Propose PEX extensions for future use-cases

xla:
* add RFC
* update spec
* update proto files

[Rendered](https://github.com/tendermint/spec/blob/xla/rfc-pex-info/rfc/006-pex-info.md)

### DIFF
--- a/proto/p2p/pex.proto
+++ b/proto/p2p/pex.proto
@@ -27,7 +27,7 @@ message PexNode {
 message PexRequestInfo {
   // Optional network/chain id to restrict the query to only return nodes which
   // are part of that chain.
-  string network;
+  string network = 1;
 }
 
 message PexResponseInfo {
--- a/proto/p2p/pex.proto
+++ b/proto/p2p/pex.proto
@@ -17,21 +17,28 @@ message PexResponse {
   repeated PexAddress addresses = 1 [(gogoproto.nullable) = false];
 }
 
-message PexAddressV2 {
-  string url = 1 [(gogoproto.customname) = "URL"];
+message PexNode {
+  // Encoded endpoint including protocol, id and path.
+  string   url  = 1 [(gogoproto.customname) = "URL"];
+  // Unaltered NodeInfo obtained during handshake with the peer.
+  NodeInfo info = 2;
 }
 
-message PexRequestV2 {}
+message PexRequestInfo {
+  // Optional network/chain id to restrict the query to only return nodes which
+  // are part of that chain.
+  string network;
+}
 
-message PexResponseV2 {
-  repeated PexAddressV2 addresses = 1 [(gogoproto.nullable) = false];
+message PexResponseInfo {
+  repeated PexNode nodes = 1 [gogoproto.nullable) = false];
 }
 
 message PexMessage {
   oneof sum {
-    PexRequest    pex_request     = 1;
-    PexResponse   pex_response    = 2;
-    PexRequestV2  pex_request_v2  = 3;
-    PexResponseV2 pex_response_v2 = 4;
+    PexRequest      pex_request       = 1;
+    PexResponse     pex_response      = 2;
+    PexRequestInfo  pex_request_info  = 3;
+    PexResponseInfo pex_response_info = 4;
   }
 }
--- a/rfc/006-pex-info.md
+++ b/rfc/006-pex-info.md
@@ -1,0 +1,92 @@
+# RFC 006: Extend PEX messages with additional info
+
+## Changelog
+
+- 2021-07-07: Initial draft
+
+## Author
+
+- @xla
+
+## Context
+
+The majority of tendermint deployments run nodes with varying responsibilities
+offering subsets of all available channels/reactors. While this is accounted
+for in the p2p stack by not sending messages to a peer which did not advise the
+corresponding channel, it depends on the `NodeInfo` exchanged during
+a handshake after a connect. Consequently there currently isn't a protocol
+native way to do rudimentally discovery to find and connect to nodes offering
+certain capabilities.
+
+## Proposal
+
+This document proposes a change to the current set of PEX messages to support
+discovery based on advertised channels. In a way that is backward compabitible
+to avoid it requiring a hard-fork of any running networks.
+
+At the core is the introduction of a new request-response pair which besides
+the endpoint information (ip:port/url) carries the `NodeInfo` exchanged during
+handshake in the response. As that carries the channels active/supported on the
+peer a receiving node can apply filtering appropriate for local use-cases.
+Furthermore any extensions to the `NodeInfo` will be carried with the response
+enabling follow-up changes extending the ABCI to involve the application in the
+node discovery.
+
+### Message changes
+
+The recently introduced [v2 messages](https://github.com/tendermint/spec/pull/312)
+SHALL be replaced with versions which carry additional information about known
+peers.
+
+``` proto
+message PexNode {
+  // Encoded endpoint including protocol, id and path.
+  string   url  = 1 [(gogoproto.customname) = "URL"];
+  // Unaltered NodeInfo obtained during handshake with the peer.
+  NodeInfo info = 2;
+}
+
+message PexRequestInfo {
+  // Optional network/chain id to restrict the query to only return nodes which
+  // are part of that chain.
+  string network;
+}
+
+message PexResponseInfo {
+  repeated PexNode nodes = 1 [gogoproto.nullable) = false];
+}
+```
+
+The resulting sum of pex messages SHALL look as follows:
+
+``` proto
+message PexMessage {
+  oneof sum {
+    PexRequest      pex_request       = 1;
+    PexResponse     pex_response      = 2;
+    PexRequestInfo  pex_request_info  = 3;
+    PexResponseInfo pex_response_info = 4;
+  }
+}
+```
+
+## Status
+
+Proposed
+
+## Consequences
+
+### Positive
+
+- enable discovery use-cases for heterogenous node networks
+- build foundation for multi-chain deployments
+- backwards compatible improvement
+
+### Negative
+
+- address book implementations have to adapt and will consequently become more
+  complex
+
+### Neutral
+
+- application driven use-cases depend on a follow up extension to ABCI
--- a/rfc/006-pex-info.md
+++ b/rfc/006-pex-info.md
@@ -49,7 +49,7 @@ message PexNode {
 message PexRequestInfo {
   // Optional network/chain id to restrict the query to only return nodes which
   // are part of that chain.
-  string network;
+  string network = 1;
 }
 
 message PexResponseInfo {
--- a/rfc/README.md
+++ b/rfc/README.md
@@ -27,3 +27,4 @@ Some RFC's will be presented at a Tendermint Dev Session. If you are an outside 
 - [003-ed25519-verification](./003-ed25519-verification.md)
 - [004-abci++](./004-abci++.md)
 - [005-reverse-sync](./005-reverse-sync.md)
+- [006-pex-nodeinfo](./006-pex-nodeinfo.md)
--- a/spec/p2p/messages/pex.md
+++ b/spec/p2p/messages/pex.md
@@ -14,8 +14,10 @@ Pex has one channel. The channel identifier is listed below.
 
 ## Message Types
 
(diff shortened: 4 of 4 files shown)


rfc/006-pex-info.md
```diff
@@ -0,0 +1,92 @@
...
+
+``` proto
+message PexNode {
+  // Encoded endpoint including protocol, id and path.
+  string   url  = 1 [(gogoproto.customname) = "URL"];
+  // Unaltered NodeInfo obtained during handshake with the peer.
+  NodeInfo info = 2;
+}
+
+message PexRequestInfo {
```
marbar3778, line 49: Could this message be extended to request peers with certain channels/reactors?

rfc/006-pex-info.md
```diff
@@ -0,0 +1,92 @@
...
+  // Optional network/chain id to restrict the query to only return nodes which
+  // are part of that chain.
+  string network;
```
marbar3778, line 52:
```suggestion
  string network = 1;
```

proto/p2p/pex.proto
```diff
@@ -17,21 +17,28 @@ message PexResponse {
...
+  string   url  = 1 [(gogoproto.customname) = "URL"];
+  // Unaltered NodeInfo obtained during handshake with the peer.
+  NodeInfo info = 2;
 }
 
-message PexRequestV2 {}
+message PexRequestInfo {
+  // Optional network/chain id to restrict the query to only return nodes which
+  // are part of that chain.
+  string network;
```
marbar3778, line 30:
```suggestion
  string network = 1;
```

rfc/006-pex-info.md
xla: Debated this as well out-of-band, @liamsi and I concluded that the complexity is not worth the trade-off until there is a tipping point where the smaller replies are significant in bandwidth saving.

liamsi: As far as I understand the reply contains the node infos, which contain the channels. So you can still do the filtering you want (by channels/reactors), locally on the receiving end.

marbar3778: I would say it's mainly a bandwidth saving optimization. 

What is the complexity this would add? Mainly the address book would either order peers accordingly or loop through them?

rfc/006-pex-info.md
```diff
@@ -0,0 +1,92 @@
...
+The recently introduced [v2 messages](https://github.com/tendermint/spec/pull/312)
+SHALL be replaced with versions which carry additional information about known
+peers.
+
+``` proto
+message PexNode {
+  // Encoded endpoint including protocol, id and path.
+  string   url  = 1 [(gogoproto.customname) = "URL"];
+  // Unaltered NodeInfo obtained during handshake with the peer.
+  NodeInfo info = 2;
```
cmwaters, line 46: Whilst `NodeInfo` does provide helpful information in regards to `Channels` and `Network`, there is a lot of information that is quite irrelevant such as:
```
// ASCIIText fields
	Moniker string        `json:"moniker"` // arbitrary moniker
	Other   NodeInfoOther `json:"other"`   // other application specific data
```
also AFAIU, the `ListenAddr` and `NodeID` are already within the URL. 

In addition, there could be other information that could be worth considering s i.e. How many peers they are connected with and how many connections can they support. 

It might be better therefore to form a different struct to get passed rather than `NodeInfo`

cmwaters: We've also had a debate internally as to whether channels should be able to dynamically change or whether they should remain static. They're currently static and not expected to change throughout the lifespan of the connection but making this something that is dynamic better supports node start up phases (where we may only run state sync or fast sync until we're ready to resume all services) as well as better handle conditions, for example, where a node begins to lag behind consensus and may temporarily close down services in reaction. This debate is something to also consider when thinking about which information to gossip amongst peers.

cmwaters (commented):
Nice work! A few comments I have regarding the use of `NodeInfo`

Also, a final things to consider. What happens in adversarial conditions. Imagine a node receives a PEX message from peer A which says that peer B has 5 channels open. Then at a later point, nefarious peer C says that peer B has no channels open. Do we update our model on peer B? Do we still connect to peer B?

rfc/006-pex-info.md
xla: > I would say it's mainly a bandwidth saving optimization.

Yeah that is the angle look at as well if in favour, but I'd like to hear the perspective of node operators on this, as the PEX messages are not frequent afaiu.

> What is the complexity this would add? Mainly the address book would either order peers accordingly or loop through them?

The complexity I'm most vary of his that of the public interface, because beyond a nodes address book now providing and storing the extra info it also has to offer the filtering, which in turn means that needs to be specific and protocolised. Where if it's up to the receiver it can be treated as an implementation detail.

rfc/006-pex-info.md
alexanderbez: @xla any thoughts on the above?

marbar3778: In the above scenario, something like periodic checkins, like what some DHTs do, with peers to get an updated look at the network would be the best way to solve this. To a certain extent, we would be recreating a dht at this point because nodes may want a network wide outlook.

xla: > In addition, there could be other information that could be worth considering s i.e. How many peers they are connected with and how many connections can they support.

In my understanding on the role of `NodeInfo` these would be great candidates to extend `NodeInfo` with. I'd favour investing in this structure as it has a significant role in the protocol already and is guaranteed to be provided between two connected/handshaked peers. This makes it a strong vehicle to expand what is communicated and in turn what can be built on that information down the road.

> We've also had a debate internally as to whether channels should be able to dynamically change or whether they should remain static. They're currently static and not expected to change throughout the lifespan of the connection but making this something that is dynamic better supports node start up phases (where we may only run state sync or fast sync until we're ready to resume all services) as well as better handle conditions, for example, where a node begins to lag behind consensus and may temporarily close down services in reaction. This debate is something to also consider when thinking about which information to gossip amongst peers.

While in the back of my head as well, this is an orthogonal concern and should be designed in isolation and luckily doesn't negatively impact this document if left out. It's conceivable that an interactive protocol build in the future that communicates the latest active channel on a node to peers results in the local `NodeInfo` for that peer being updated. One thing is important no matter if the channels for a node are changing over the lifetime of a connection, when a PEX request-response interaction happen what comes through in the response is always gonna be a snapshot, therefore improvements like you hinting at can be plugged in later without the mechanism being different.

> @xla any thoughts on the above?

See paragraph above.

xla: > In the above scenario, something like periodic checkins, like what some DHTs do, with peers to get an updated look at the network would be the best way to solve this. To a certain extent, we would be recreating a dht at this point because nodes may want a network wide outlook.

Could you expand further what this DHT would key on?

xla:
> Also, a thing to consider. What happens in adversarial conditions. Imagine a node receives a PEX message from peer A which says that peer B has 5 channels open. Then at a later point, nefarious peer C says that peer B has no channels open. Do we update our model on peer B? Do we still connect to peer B?

What's the exact order of events here? The point here raised is extremely valid, I wonder where to capture that tho, as it is not really impacting the shape of the new PEX messages but the protocol for how to make sense of these messages, would you agree to that?

alexanderbez (approved):

cmwaters:
> it is not really impacting the shape of the new PEX messages but the protocol for how to make sense of these messages, would you agree to that?

This is true but I would want to have an idea of how the protocol makes sense of the messages before agreeing on the shape of the messages. If we can't really use the information in the message then it doesn't really make sense to have them. I also think that in this case, if the information is incorrectly acted on, it could cause a malicious user to craft themselves a situation where they could perform some type of eclipse attack so I would prefer if this RFC was explicit in the way that information such as `channels` should be handled.

So it's obvious that sharing information about which `channels` a node has is helpful because it can help us decide who we should connect with. Same can be said with the amount of connections a node has open. We don't want to connect to peers that are just going to outright reject us because they don't have the capacity. The problem here is that if a peer misinforms us we have no way of correctly knowing that and being able to punish that peer. 

As a way of sidestepping some of potential pitfalls, we could include some basic heuristics i.e. ignore a peer's `NodeInfo` if it contains no channels.

To me, a better way to do this would be for every peer to cryptographically secure their own `NodeInfo` such that no malicious peer could tamper with it and misinform surrounding peers. This could be done the same way that votes get signed.  If a node changes their own `NodeInfo`, they could increment a nonce so that surrounding peers know what the most accurate information on that node is.

xla:
@cmwaters The current address book/pex in Go is a port of the btcd implementation if my memory doesn't trick me and its internal behaviour is not part of the spec, or more specifically the behaviour when a `PexResponse` is received. I don't see strong arguments so far why we should specify the implementation specific behaviour besides close to the implementation for documentation.

> This is true but I would want to have an idea of how the protocol makes sense of the messages before agreeing on the shape of the messages. If we can't really use the information in the message then it doesn't really make sense to have them. I also think that in this case, if the information is incorrectly acted on, it could cause a malicious user to craft themselves a situation where they could perform some type of eclipse attack so I would prefer if this RFC was explicit in the way that information such as `channels` should be handled.

Could you expand how this attack vector is different/special from the current potential to eclipse a peer with PEX messages?

> To me, a better way to do this would be for every peer to cryptographically secure their own `NodeInfo` such that no malicious peer could tamper with it and misinform surrounding peers. This could be done the same way that votes get signed. If a node changes their own `NodeInfo`, they could increment a nonce so that surrounding peers know what the most accurate information on that node is.

I do like this idea, but I fail to see how the signature would be meaningful prior to being connected to the peer which signed the `NodeInfo`.

cmwaters:
> Could you expand how this attack vector is different/special from the current potential to eclipse a peer with PEX messages?

The attack is still the same. It's just that prior, we connected to new peers from the address book at random. Whereas now I imagine we'll use the NodeInfo as a heuristic to choose who out of the new addresses we should dial. This makes it a little easier for an attacker because they can try manipulate our address book to favour dialing nodes that are theirs (mainly by making other nodes seem undesirable to connect with). 

> I do like this idea, but I fail to see how the signature would be meaningful prior to being connected to the peer which signed the NodeInfo.

Hmm let me think about this a bit more. Say we are connected to two peers: `NodeA` (faulty) and `NodeB` (honest) and they are sending info about `NodeC` (faulty) and `NodeD` (honest). Let's say that you can send arbitrary information which makes one peer look more attractive to connect to than another peer (channels is a good example of that). I guess it's important to note that when we actually connect with peers and they give us different information than what was provided we will reassess the connection based on the new information.
- If `NodeA` falsely advertises that `NodeC` is better to connect with than `NodeD` and we get no other information then we will connect with `NodeC` first. If we have enough slots we will also connect with `NodeD`.
- If `NodeB` advertises conflicting information about `NodeD` then they will have conflicting PubKeys and NodeIDs but the same address.  The node would keep both copies of `NodeD` and if it dialed the address it could work out which one was the true `NodeD` 
So as long as a Node is initially connected to one honest peer, I think it will be able to snuff out other peers sending it fake addresses. What do you think?

xla:
>     * So as long as a Node is initially connected to one honest peer, I think it will be able to snuff out other peers sending it fake addresses. What do you think?

Sure, the info provided by the actual node should be the canonical one and never what has been learned from another peer.

xla:
Why is this PR closed?

xla:
Can we reopen this? Quite frustrating to see this just being closed without any closure.

liamsi:
Sorry, accidentally deleted the branch (restored now).

Looks like this was automatically closed for no other reason than inactivity 😢 I think the tendermint team will soon figure out the granularity of its [p2p roadmap](https://github.com/tendermint/tendermint/blob/master/docs/rfc/rfc-000-p2p-roadmap.rst#solutions). If the idea will be to abandon the current PEX soon then it is probably not worth extending it right now. If the idea is to keep it this proposal is still worth discussing and exploring. I'm in favor of not solely relying on a DHT but keeping a lightweight PEX-like protocol for ad-hoc peer/service discovery once the connections were established via e.g. a dht (independent of a potential move to libp2p or not).

xla:
Thanks for the follow-up. While I am aware of the ongoing discussions on the adoption of libp2p, it is an orthogonal concern to what is proposed in this change-set. It appears to be the go-to reasoning to not move forward with incremental changes until that long-ish transition period is over - with uncertain outcome at this point anyway.

Exploring and hashing out general service discovery is not in-scope for this PR and if that is the expectation I'd rather conclude the conversation at this point.

Issue #293, opened 2021-05-07
Light client in Go initialization


josef-widder:
Surfaced from [Informal Systems](https://informal.systems/) Evidence end-to-end audit of https://github.com/tendermint/tendermint/commit/5a39f72aa8288696d7284e56e3f630ab721a9f9a.

The initialization logic of the lightnode in the Go implementation differs significantly from the initialization that is specified in the supervisor specification. Here is a list of features that should be documented in an "implementation-specific" specifications that focus on the features of the Go implementation of the light client

- [Trustoptions](https://github.com/tendermint/tendermint/blob/5a39f72aa8288696d7284e56e3f630ab721a9f9a/light/client.go#L148) are used to set trusting period and to point into the initial lightstore to signal that the user trusts that header.
  
- if initialized with a header hash it does [cross check with witnesses](https://github.com/tendermint/tendermint/blob/5a39f72aa8288696d7284e56e3f630ab721a9f9a/light/client.go#L983). A faulty witness, may prevent the lightnode to initialize. As a result this check was not included in the [specification](https://github.com/tendermint/spec/blob/master/spec/light-client/supervisor/supervisor_001_draft.md#lc-func-init1)

- [latesttrustedblock](https://github.com/tendermint/tendermint/blob/5a39f72aa8288696d7284e56e3f630ab721a9f9a/light/client.go#L123). II guess for performance reasons it is stored explicitly. It needs to be maintained explicitly.
  

- More generally, the [initialization](https://github.com/tendermint/tendermint/blob/5a39f72aa8288696d7284e56e3f630ab721a9f9a/light/client.go#L264) based on `options.Height` a bit unclear. This is capturing a unclear case distinction where on initialization "options" is pointing into the lightstore. We should document initialization conditions. What is the initialization supposed to do with respect to an existing lightstore?

cmwaters:
Hey @josef-widder, I would like to write a section on initialisation - specifically how this works when the light clients trusted store is not empty. Can you point to where I should add a piece on such logic. 

> if initialized with a header hash it does cross check with witnesses. A faulty witness, may prevent the lightnode to initialize. As a result this check was not included in the specification

I think we should make the explicit check that all nodes must initially agree on the same block and that that aligns with the trusted block. If a faulty witness wants to prevent the node to initialize, I feel it's better that the operator sees this immediately and can remove that witness if they choose rather than having to deal with it at some later point. It also provides a liveness check.

josef-widder:
[...]

PR #320, opened 2021-07-16
light: update initialization description

cmwaters:
Closes: #293

### DIFF
--- a/spec/light-client/supervisor/supervisor_001_draft.md
+++ b/spec/light-client/supervisor/supervisor_001_draft.md
@@ -488,7 +488,7 @@ Cross-checking this trusted block with providers upon initialization is helpful
 for ensuring that the node is responsive and correctly configured but does not
 increase trust since proving a conflicting block is a
 [light client attack](https://informal.systems)
-and not just a [bogus](https://informal.systems) block could result in
+and not just a [bogus](https://github.com/tendermint/spec/blob/master/spec/light-client/detection/detection_003_reviewed.md#tmbc-bogus1) block could result in
 performing backwards verification beyond the trusted period, thus a fruitless
 endeavour.
 
--- a/spec/light-client/supervisor/supervisor_001_draft.md
+++ b/spec/light-client/supervisor/supervisor_001_draft.md
@@ -463,67 +463,91 @@ func Sequential-Supervisor (initdata LCInitData) (Error) {
 #### Initialization
 
 The light client is based on subjective initialization. It has to
-trust the initial data given to it by the user. It cannot do any
-detection of attack. So either upon initialization we obtain a
-lightblock and just initialize the lightstore with it. Or in case of a
-genesis file, we download, verify, and cross-check the first block, to
-initialize the lightstore with this first block. The reason is that
-we want to maintain [LCV-INV-TP.1] from the beginning.
+trust the initial data given to it by the user. It cannot perform any
+detection of an attack yet instead requires an initial point of trust.
+There are three forms of initial data:
 
-> If the lightclient is initialized with a lightblock, one might think
-> it may increase trust, when one cross-checks the initial light
-> block. However, if a peer provides a conflicting
-> lightblock, the question is to distinguish the case of a
-> [bogus](https://informal.systems) block (upon which operation should proceed) from a
-> [light client attack](https://informal.systems) (upon which operation should stop). In
-> case of a bogus block, the lightclient might be forced to do
-> backwards verification until the blocks are out of the trusting
-> period, to make sure no previous validator set could have generated
-> the bogus block, which effectively opens up a DoS attack on the lightclient
-> without adding effective robustness.
+- A trusted block from a prior initialization
+- A trusted height and hash
+- A genesis file
+
+The golang light client implementation checks this initial data in that 
+order; first attempting to find a trusted block from the trusted store, 
+then acquiring a light block from the primart at the trusted height and matching
+the hash, or finally checking for a genesis file to verify the initial header.
+
+The light client doesn't need to check if the trusted block is within the
+trusted period because it already trusts it, however, if the light block is 
+outside the trust period, there is a higher chance the light client won't be
+able to verify anything. 
+
+Cross-checking this trusted block with providers upon initialization is helpful
+for ensuring liveness and correct configuration of such nodes but does not 
+increase trust since proving a conflicting block is a 
+[light client attack](https://informal.systems) 
+and not just a [bogus](https://informal.systems) block could result in
+performing backwards verification beyond the trusted period, thus a fruitless
+endeavour.
+
+However, with the notion of it's better to fail earlier than later, the golang
+light client implementation will perform a consistency check on all providers
+and will error if one returns a different header, allowing the user
+the opportunity to reinitialize.
 
 #### **[LC-FUNC-INIT.1]:**
 
 ```go
-func InitLightClient (initData LCInitData) (LightStore, Error) {
+func InitLightClient(initData LCInitData) (LightStore, Error) {
+    var initialBlock LightBlock
 
-    if LCInitData.LightBlock != nil {
-        // we trust the provided initial block.
-        newblock := LCInitData.LightBlock
+    switch {
+    case LCInitData.TrustedBlock != nil:
+        // we trust the block from a prior initialization
+        initialBlock = LCInitData.TrustedBlock
+
+    case LCInitData.TrustedHash != nil:
+        untrustedBlock := FetchLightBlock(PeerList.Primary(), LCInitData.TrustedHeight)
+        
+
+        // verify that the hashes match
+        if untrustedBlock.Hash() != LCInitData.TrustedHash {
+            return nil, Error("Primary returned block with different hash")
+        }
+        // after checking the hash we now trust the block
+        initialBlock = untrustedBlock        
     }
-    else {
-        genesisBlock := makeblock(initData.genesisDoc);
+    case LCInitData.Genesis != nil:
+        untrustedBlock := FetchLightBlock(PeerList.Primary(), LCInitData.Genesis.InitialHeight)
+        
+        // verify that 2/3+ of the validator set signed the untrustedBlock
+        if err := VerifyCommit(untrustedBlock.Commit, LCInitData.Genesis.Validators); err != nil {
+            return nil, err
+        }
 
-        result := NoResult;
-        while result != ResultSuccess {
-            current = FetchLightBlock(PeerList.primary(), genesisBlock.Header.Height + 1)
-            // QUESTION: is the height with "+1" OK?
+        // we can now trust the block
+        initialBlock = untrustedBlock
+    default:
+        return nil, Error("No initial data was provided")
 
-            if CANNOT_VERIFY = ValidAndVerify(genesisBlock, current) {
-                Replace_Primary();
-            }
-            else {
-                result = ResultSuccess
-            }
-        }
-  
-        // cross-check
-  auxLS := new LightStore
-  auxLS.Add(current)
-        Evidences := AttackDetector(genesisBlock, auxLS)
-        if Evidences.Empty {
-            newBlock := current
-        }
-        else {
-            // [LC-SUMBIT-EVIDENCE.1]
-            submitEvidence(Evidences);
-            return(nil, ErrorAttack);
-        }
+    if err := CrossCheck(initialBlock, PeerList.Witnesses()); err != nil {
+        return nil, err
     }
 
+    // initialize light store
     lightStore := new LightStore;
     lightStore.Add(newBlock);
     return (lightStore, OK);
+}
+
+func CrossCheck(lb LightBlock, witnesses []Provider) error {
+    for _, witness := range witnesses {
+        witnessBlock := FetchLightBlock(witness, lb.Height)
+
+        if witnessBlock.Hash() != lb.Hash() {
+            return Error("Witness has different block")
+        }
+    }
+    return OK
 }
 
 ```
--- a/spec/light-client/supervisor/supervisor_001_draft.md
+++ b/spec/light-client/supervisor/supervisor_001_draft.md
@@ -467,26 +467,27 @@ func Sequential-Supervisor (initdata LCInitData) (Error) {
 The light client is based on subjective initialization. It has to
 trust the initial data given to it by the user. It cannot perform any
 detection of an attack yet instead requires an initial point of trust.
-There are three forms of initial data:
+There are three forms of initial data which are used to obtain the 
+first trusted block:
 
 - A trusted block from a prior initialization
 - A trusted height and hash
 - A genesis file
 
-The golang light client implementation checks this initial data in that 
-order; first attempting to find a trusted block from the trusted store, 
-then acquiring a light block from the primart at the trusted height and matching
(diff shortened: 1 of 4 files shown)


josef-widder (changes requested):
There was a quite long discussion about the current initialization that clarified that cross-checking with secondaries does not add security, and therefore it was not included. I thought the idea was to add another section on the golang initialization (that differs from that one) rather than changing the existing spec. We should clarify this. 

In any case, if the algorithm changes, the tag **[LC-FUNC-INIT.1]:** would need to go to a new version number  **[LC-FUNC-INIT.2]:*

spec/light-client/supervisor/supervisor_001_draft.md
```diff
@@ -463,69 +463,93 @@ func Sequential-Supervisor (initdata LCInitData) (Error) {
...
+outside the trust period, there is a higher chance the light client won't be
+able to verify anything. 
+
+Cross-checking this trusted block with providers upon initialization is helpful
+for ensuring liveness and correct configuration of such nodes but does not 
+increase trust since proving a conflicting block is a 
+[light client attack](https://informal.systems) 
+and not just a [bogus](https://informal.systems) block could result in
+performing backwards verification beyond the trusted period, thus a fruitless
+endeavour.
```
cmwaters, line 490: Here I do note that cross-checking does no increase trust. I then go on to say that the `golang light client imlementation` performs a "consistency check" on all providers

spec/light-client/supervisor/supervisor_001_draft.md
```diff
@@ -463,69 +463,93 @@ func Sequential-Supervisor (initdata LCInitData) (Error) {
...
         }
+
+        // we can now trust the block
+        initialBlock = untrustedBlock
+    default:
+        return nil, Error("No initial data was provided")
+
+    if err := CrossCheck(initialBlock, PeerList.Witnesses()); err != nil {
+        return nil, err
     }
```
cmwaters, line 534: I could remove this cross check from the spec pseudocode as it's not really important to the correctness of the spec. I think `[LC-FUNC-INIT.1]` would need to be updated to  `[LC-FUNC-INIT.2]` anyway because there are still changes to the logic

josef-widder: OK. Perhaps we could also add a comment within the pseudo code that that line is optional and done in the golang version. I am a bit under water right now and will do a closer check of the new version in the week of July 26

spec/light-client/supervisor/supervisor_001_draft.md
```diff
@@ -463,69 +463,94 @@ func Sequential-Supervisor (initdata LCInitData) (Error) {
...
-            submitEvidence(Evidences);
-            return(nil, ErrorAttack);
+        // after checking the hash we now trust the block
+        initialBlock = untrustedBlock        
+    }
+    case LCInitData.Genesis != nil:
+        untrustedBlock := FetchLightBlock(PeerList.Primary(), LCInitData.Genesis.InitialHeight)
+        
+        // verify that 2/3+ of the validator set signed the untrustedBlock
+        if err := VerifyCommit(untrustedBlock.Commit, LCInitData.Genesis.Validators); err != nil {
```
josef-widder, line 523: Is the function defined somewhere? Previously we used
[`ValidandVerified`](https://github.com/tendermint/spec/blob/master/spec/light-client/verification/verification_002_draft.md#lcv-func-valid2) here (with a typo). As it takes as inputs two lightblocks we generated a gensisblock.

spec/light-client/supervisor/supervisor_001_draft.md
josef-widder: Sorry, I didn't see that earlier.

I am not sure I understand how it is helpful for liveness, though? Perhaps you could add a sentence that clarifies the scenario you have in mind?

spec/light-client/supervisor/supervisor_001_draft.md
```diff
@@ -463,69 +463,94 @@ func Sequential-Supervisor (initdata LCInitData) (Error) {
...
 
-    if LCInitData.LightBlock != nil {
-        // we trust the provided initial block.
-        newblock := LCInitData.LightBlock
-    }
-    else {
-        genesisBlock := makeblock(initData.genesisDoc);
+    switch {
+    case LCInitData.TrustedBlock != nil:
+        // we trust the block from a prior initialization
```
josef-widder, line 505: Is it necessarily from "prior" initialization?

spec/light-client/supervisor/supervisor_001_draft.md
```diff
@@ -463,69 +463,94 @@ func Sequential-Supervisor (initdata LCInitData) (Error) {
...
+        initialBlock = LCInitData.TrustedBlock
 
-        result := NoResult;
-        while result != ResultSuccess {
-            current = FetchLightBlock(PeerList.primary(), genesisBlock.Header.Height + 1)
-            // QUESTION: is the height with "+1" OK?
+    case LCInitData.TrustedHash != nil:
```
josef-widder, line 508: Please add the Field `TrustedHash` to the data structure definition further above in the code.

spec/light-client/supervisor/supervisor_001_draft.md
cmwaters: By liveness I'm meaning that we were able to establish a connection with the provider and they were responsive. This might actually be a different definition of liveness to the one you have in mind

spec/light-client/supervisor/supervisor_001_draft.md
cmwaters: I guess not. As in it's feasible that someone adds a light block to the store and then passes that in (without it going through a previous initialization process)

spec/light-client/supervisor/supervisor_001_draft.md
cmwaters: It's mentioned here: https://github.com/tendermint/spec/blob/master/spec/light-client/verification/README.md and is the name of a function within the golang codebase. 

AFAICS: `ValidandVerified` does three things
- Check that all formats are consistent and as expected
- Checks that their is a sufficient degree of overlap in the validators between two heights
- Checks that the 2/3+ of the validator set and the untrusted height signed the untrusted block

`VerifyCommit` performs only the last of these three things

spec/light-client/supervisor/supervisor_001_draft.md
cmwaters: Sure. I've added them. One might argue that the `TrustedHash` and `TrustedHeight` combination are just for convenience (instead of passing the entire `TrustedBlock`) and so doesn't belong in the spec. But perhaps it's good to have them here just for documentation sake because it's definitely a prevalent concept in the golang codebase

spec/light-client/supervisor/supervisor_001_draft.md
josef-widder: I didn't have any definition in mind ;-)  
I think if there is a specific goal, it should be written more explicitly.

spec/light-client/supervisor/supervisor_001_draft.md
josef-widder: I see. I think we should add the function definition of `VerifyCommit` to the spec. (Also the linked README.md only contains `VerifyCommitTrusting` and `VerifyCommitFull`)

spec/light-client/supervisor/supervisor_001_draft.md
josef-widder: I just wondered, because the text of the comment was different in the previous version of the spec?

spec/light-client/supervisor/supervisor_001_draft.md
josef-widder: I think it is good to give the options and details here. Thanks for adding the hash option?

spec/light-client/supervisor/supervisor_001_draft.md
cmwaters: Right. The way the light client is initialised, at least in the golang repo, is that we give users the option to either start form a trusted hash and height or from a trusted store which would've come from previously using the light client. I guess I was just translating that notion across

spec/light-client/supervisor/supervisor_001_draft.md
cmwaters: In this case, I'm talking about `VerifyCommitFull`. `VerifyCommitTrusting` is the name of the function that performs the second bullet point

cmwaters: Should I add these functions to the verification document?

josef-widder: Yes, please

spec/light-client/supervisor/supervisor_001_draft.md
josef-widder: It would be great to add this explanation as a comment to the spec.

spec/light-client/supervisor/supervisor_001_draft.md
cmwaters: Added `VerifyCommitLight` here: https://github.com/tendermint/spec/blob/callum/light-client/spec/light-client/verification/verification_001_published.md#lcv-func-verifycommitlight1

cmwaters: It would be nice to have all these verification function well documented in the spec: https://github.com/tendermint/tendermint/blob/master/types/validation.go

cmwaters:
@josef-widder this PR is ready for a second pass whenever you get the time

spec/light-client/supervisor/supervisor_001_draft.md
```diff
@@ -463,69 +465,95 @@ func Sequential-Supervisor (initdata LCInitData) (Error) {
...
+then acquiring a light block from the primary at the trusted height and matching
+the hash, or finally checking for a genesis file to verify the initial header.
+
+The light client doesn't need to check if the trusted block is within the
+trusted period because it already trusts it, however, if the light block is
+outside the trust period, there is a higher chance the light client won't be
+able to verify anything.
+
+Cross-checking this trusted block with providers upon initialization is helpful
+for ensuring that the node is responsive and correctly congfigured but does not
```
goral09, line 488:
```suggestion
for ensuring that the node is responsive and correctly configured but does not
```

spec/light-client/supervisor/supervisor_001_draft.md
```diff
@@ -463,69 +465,95 @@ func Sequential-Supervisor (initdata LCInitData) (Error) {
...
+increase trust since proving a conflicting block is a
+[light client attack](https://informal.systems)
+and not just a [bogus](https://informal.systems) block could result in
```
goral09, line 491: I'm not sure how these links are helping here.

josef-widder, line 491:
```suggestion
and not just a [bogus](https://github.com/tendermint/spec/blob/master/spec/light-client/detection/detection_003_reviewed.md#tmbc-bogus1) block could result in
```

spec/light-client/supervisor/supervisor_001_draft.md
```diff
@@ -463,69 +465,95 @@ func Sequential-Supervisor (initdata LCInitData) (Error) {
...
+
+The light client doesn't need to check if the trusted block is within the
+trusted period because it already trusts it, however, if the light block is
+outside the trust period, there is a higher chance the light client won't be
+able to verify anything.
+
+Cross-checking this trusted block with providers upon initialization is helpful
+for ensuring that the node is responsive and correctly congfigured but does not
+increase trust since proving a conflicting block is a
+[light client attack](https://informal.systems)
```
josef-widder, line 490:
```suggestion
[light client attack](https://github.com/tendermint/spec/blob/master/spec/light-client/detection/detection_003_reviewed.md#tmbc-lc-attack1)
```

josef-widder (commented):
I added suggestions for links to definitions of attacks and bogus blocks

spec/light-client/supervisor/supervisor_001_draft.md
josef-widder: Thanks a lot for spotting the default links!! I have replaced them with meaningful ones.

goral09: Thank you!

spec/light-client/supervisor/supervisor_001_draft.md
```diff
@@ -463,69 +465,95 @@ func Sequential-Supervisor (initdata LCInitData) (Error) {
...
-> period, to make sure no previous validator set could have generated
-> the bogus block, which effectively opens up a DoS attack on the lightclient
-> without adding effective robustness.
+trust the initial data given to it by the user. It cannot perform any
+detection of an attack yet instead requires an initial point of trust.
+There are three forms of initial data which are used to obtain the 
+first trusted block:
+
+- A trusted block from a prior initialization
+- A trusted height and hash
```
goral09, line 474: Is height necessary? Wouldn't `hash` identify a block uniquely?

spec/light-client/supervisor/supervisor_001_draft.md
```diff
@@ -463,69 +465,95 @@ func Sequential-Supervisor (initdata LCInitData) (Error) {
...
+
+The golang light client implementation checks this initial data in that
+order; first attempting to find a trusted block from the trusted store,
+then acquiring a light block from the primary at the trusted height and matching
+the hash, or finally checking for a genesis file to verify the initial header.
+
+The light client doesn't need to check if the trusted block is within the
+trusted period because it already trusts it, however, if the light block is
+outside the trust period, there is a higher chance the light client won't be
+able to verify anything.
```
goral09, line 485: If the trusted block is outside of the "trust period" (i.e. all of its validators have already unbonded), can it be trusted at all? Light client would be subject to a long range attack in that case, correct?

spec/light-client/supervisor/supervisor_001_draft.md
josef-widder: I guess the problem is that currently we can only ask a full node for a block at a given height, rather than a block corresponding to a hash.

goral09: I see, thank you.

cmwaters:
Sorry I haven't followed up with this in a while. I've been busy trying to get the first 0.35 release candidate out the door. I will aim to look at this later this week. AFAICS, the remaining thing left here is to update [LC-FUNC-INIT.1] to [LC-FUNC-INIT.2]. Is there anything else I'm missing?

josef-widder:
> Sorry I haven't followed up with this in a while. I've been busy trying to get the first 0.35 release candidate out the door. I will aim to look at this later this week. AFAICS, the remaining thing left here is to update [LC-FUNC-INIT.1] to [LC-FUNC-INIT.2]. Is there anything else I'm missing?

yep. Renaming the tag, and moving over the changes to new files with new version numbers
for the supervisor and the verification.

I guess for verification the changes should be done from `verification_002_draft.md` and put into a new file `verification_003_draft.md`.

josef-widder (approved):
We should just adopt the tags, and move the changes to new versions of the spec.

cmwaters:
> I guess for verification the changes should be done from verification_002_draft.md and put into a new file verification_003_draft.md.

I still don't understand if `verification_002` is a draft why there need's to be `verification_003`

PR #321, opened 2021-07-16
light: add sequential verification method

cmwaters:
Closes: #292

### DIFF
--- a/spec/light-client/verification/verification_001_published.md
+++ b/spec/light-client/verification/verification_001_published.md
@@ -517,6 +517,16 @@ func (ls LightStore) Get(height Height) (LightBlock, bool)
     - returns a LightBlock at a given height or false in the second argument if
     the LightStore does not contain the specified LightBlock.
 
+#### **[LCV-FUNC-VALID-LIGHT-BLOCK.1]**
+
+```go
+func (lb LightBlock) ValidateBasic() error
+```
+
+- Expected postcondition
+  - *Header*, *Commit* and *ValidatorSet* are [**well-formed**][block]
+  - *Header.ValidatorsHash* == hash(*ValidatorSet*)
+
 #### **[LCV-FUNC-LATEST-VERIF.1]**
 
 ```go
@@ -734,7 +744,11 @@ func VerifyToTarget(primary PeerID, lightStore LightStore,
         }
 
         // Verify
-        verdict = ValidAndVerified(lightStore.LatestVerified, current)
+        if sequential {
+          verdict = VerifySequential(lightStore.LatestVerified, current)
+        } else {
+          verdict = ValidAndVerified(lightStore.LatestVerified, current)
+        }
 
         // Decide whether/how to continue
         if verdict == SUCCESS {
@@ -780,7 +794,7 @@ func ValidAndVerified(trusted LightBlock, untrusted LightBlock) Result
 ```
 
 - Expected precondition:
-    - *untrusted* is valid, that is, satisfies the soundness [checks][block]
+    - *untrusted* and *trusted* are valid, that is, satisfies **[LCV-FUNC-VALID-LIGHT-BLOCK.1]**
     - *untrusted* is **well-formed**, that is,
         - *untrusted.Header.Time < now + clockDrift*
         - *untrusted.Validators = hash(untrusted.Header.Validators)*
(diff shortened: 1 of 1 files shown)


cmwaters:
Just browsing throught the light spec. There are a couple of times where we mention backwards verification but I don't think we explicity include it in the `VerifyToTarget` algorithm. In the golang implementation. The untrusted block can be greater than or less than the trusted block.

josef-widder:
In the specification of the supervisor we have [[LC-FUNC-MAIN-VERIF-DETECT.1]](https://github.com/tendermint/spec/blob/master/spec/light-client/supervisor/supervisor_001_draft.md#lc-func-main-verif-detect1) that details where we need backward verification. `VerifyToTarget` is the forward case here (in the else branch of `if r2 = false`).

cmwaters:
Is `Backwards` already defined somewhere?

josef-widder:
https://github.com/tendermint/spec/blob/master/spec/light-client/verification/verification_002_draft.md#lcv-func-backwards2

cmwaters:
Ah I missed that. I will remove the backwards verification method then

cmwaters:
I've removed the duplicate backwards verification section

spec/light-client/verification/verification_001_published.md
```diff
@@ -517,6 +517,16 @@ func (ls LightStore) Get(height Height) (LightBlock, bool)
     - returns a LightBlock at a given height or false in the second argument if
     the LightStore does not contain the specified LightBlock.
 
+#### **[LCV-FUNC-VALID-LIGHT-BLOCK.1]**
+
+```go
+func (lb LightBlock) ValidateBasic() error
+```
+
+- Expected postcondition
```
josef-widder, line 526: I think, typically this should be expected precondition, and the postcondition is something like "no error in case the precondition is satisfied"

spec/light-client/verification/verification_001_published.md
```diff
@@ -734,7 +744,11 @@ func VerifyToTarget(primary PeerID, lightStore LightStore,
         }
 
         // Verify
-        verdict = ValidAndVerified(lightStore.LatestVerified, current)
+        if sequential {
```
josef-widder, line 747: What is `sequential` here?

spec/light-client/verification/verification_001_published.md
```diff
@@ -829,9 +867,11 @@ func ValidAndVerified(trusted LightBlock, untrusted LightBlock) Result
 func Schedule(lightStore, nextHeight, targetHeight) Height
 ```
 
-- Implementation remark: If picks the next height to be verified.
-  We keep the precise choice of the next header under-specified. It is
-  subject to performance optimizations that do not influence the correctness
+- In the case of sequential verification, `Schedule` will always return the very
```
josef-widder, line 870: That is a very nice way to add sequential verification into the framework!!

spec/light-client/verification/verification_001_published.md
cmwaters: a boolean config variable that is part of the light client. I've just glossed over that part

josef-widder: I see. There is a section `### Variables`. Could you please add `sequential` there and explain the initialization/assumption there.

cmwaters: Done

josef-widder:
I just realized that the changes were done on version 001, while there is a later 002 version. How do we resolve that?

cmwaters:
I don't mind transferring this across to the draft version. I believe I did it originally in the published version because these are changes that are already in place in the implementation whereas the draft copy suggests to me that the new specification hasn't been finalized yet.

On a broader point, how would the research team feel about using the git based versioning system as opposed to having different versioned documents? I feel it's much easier to iterate upon existing pieces than to rewrite a lot of this stuff. We could just add a changelog to the top of these documents for visibility and readers can simply track through the different releases or even commits to work out the state of the spec at different points in time. It also helps give a good relation to what state each document was in with respect to other documents and makes it easier to track ongoing work (through PR's).

josef-widder:
> On a broader point, how would the research team feel about using the git based versioning system as opposed to having different versioned documents? I feel it's much easier to iterate upon existing pieces than to rewrite a lot of this stuff.

I think we care less about the versioned documents than about the versioned tags as we use them as references across artifacts. For instance [here](https://github.com/tendermint/spec/blob/ed7fa806939f8305f8083a744d9813646ec8b3f6/spec/light-client/verification/LCVerificationApi_003_draft.tla#L52) we have parts of the TLA+ specification of [[LCV-FUNC-VALID.1]](https://github.com/tendermint/spec/blob/master/spec/light-client/verification/verification_001_published.md#lcv-func-valid1). This function has changed in version 002, so we know that the TLA+ refers to the old version. It is not obvious how to trace such dependencies just via git. On our side @shonfeder is working on a tracing application for such dependencies.

Also the terms "_published", "_draft", etc. help us to understand at what stage a spec is, that is, in what quality or how well it had been reviewed.

cmwaters:
I see. There are probably merits to both systems. I have a slight preference towards the spec only containing complete documents and thus drafts being tracked as PR's on separate branches. 

Back to the more specific problem, do you want me to port the changes across to v002?

konnov:
> I don't mind transferring this across to the draft version. I believe I did it originally in the published version because these are changes that are already in place in the implementation whereas the draft copy suggests to me that the new specification hasn't been finalized yet.
> 
> On a broader point, how would the research team feel about using the git based versioning system as opposed to having different versioned documents? I feel it's much easier to iterate upon existing pieces than to rewrite a lot of this stuff. We could just add a changelog to the top of these documents for visibility and readers can simply track through the different releases or even commits to work out the state of the spec at different points in time. It also helps give a good relation to what state each document was in with respect to other documents and makes it easier to track ongoing work (through PR's).

Actually, we introduced version tags to the specifications after using git for some time. The problem with the git approach is that specifications are not developed at the same pace as the implementation, so it may happen (and it did happen) that the specification was newer than the implementation and the other way around. It was hard to figure out what specification we were talking about. On top of that, git repositories were renamed, files were moved and sometimes we even had to introduce a copy of the spec, in order to use it in another project.

After experiencing all those issues, we have decided to attach clear version tags to every specification and also label it with a status. So we would know that a draft requires further discussion and verification, whereas a published version is not going to be modified, it can only be converted to the next version spec.

If you check engineering specifications (e.g., Java language specs), they usually co-exist in one place and it is clear which version every document has. I am not sure, whether anybody is comparing them, but I would not be surprised, if somebody wanted to run diff against two TLA+ specs. I agree that we need a changelog as well, and it would be great to have it right in the spec.

josef-widder:
> Back to the more specific problem, do you want me to port the changes across to v002?

Could you please add a new file 003 with the changes starting from 002. Please make sure that all functions whose pseudo code (or definition) has been changed get a new version number. Thanks!

cmwaters:
I'm implying that you want me to remove the changes from the 001 version right?

josef-widder:
> I'm implying that you want me to remove the changes from the 001 version right?

Yes, please. When there is a new version, the older version should not be changed anymore.

spec/light-client/verification/verification_001_published.md
```diff
@@ -821,6 +839,30 @@ func ValidAndVerified(trusted LightBlock, untrusted LightBlock) Result
...
+```
+
+- Expected precondition:
+    - *untrusted* and *trusted* are valid, that is, satisfies **[LCV-FUNC-VALID-LIGHT-BLOCK.1]**
+    - *untrusted.Header.Time < now + clockDrift*
+    - *untrusted.Header.Time > now - trustingPeriod*
+    - blocks are adjacent. That is `trusted.Header.Height` + 1 == `untrusted.Header.Height`
+    - *trusted.Commit* is a commit for the header
+     *trusted.Header*, i.e., it contains
+     the correct hash of the header, and +2/3 of signatures
```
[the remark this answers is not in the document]
josef-widder: Here we check whether the commit matches the header, which requires us to check [[TMBC-SOUND-DISTR-POSS-COMMIT.1]](https://github.com/tendermint/spec/blob/master/spec/light-client/verification/verification_001_published.md#tmbc-sound-distr-poss-commit1).

josef-widder:
I guess the plan was to move over the changes to the new file and merge. So I re-open.

Issue #295, opened 2021-05-07
Evidence reactor specification


josef-widder:
Surfaced from [Informal Systems](https://informal.systems/) Evidence end-to-end audit of https://github.com/tendermint/tendermint/commit/5a39f72aa8288696d7284e56e3f630ab721a9f9a.

The code of the evidence reactor code does not mention any design documents that were used to document the underlying protocol. We thus consider this protocol under-specified. We propose to document the functionality of the evidence reactor. In the following we highlight some points that seemed important to discuss.

### Evidence Pool

General comments on [pool.go](https://github.com/tendermint/tendermint/blob/5a39f72aa8288696d7284e56e3f630ab721a9f9a/evidence/pool.go)

- Relationships between the functions (what calls what, who is responsible) for what checks are unclear.

- the context in which this code runs is unclear. Following [comments in the code](https://github.com/tendermint/tendermint/blob/5a39f72aa8288696d7284e56e3f630ab721a9f9a/evidence/verify.go#L35) it seems that it is possible that the node is not synchronized. In this case several checks become imprecise.

- invariants on data is only implicit

### AddEvidence

The responsibility of the function are unclear. 
- In the [comment before the function](https://github.com/tendermint/tendermint/blob/5a39f72aa8288696d7284e56e3f630ab721a9f9a/evidence/pool.go#L145) it is claimed that "AddEvidence checks the evidence is valid and adds it to the pool",
- in the [comment](https://github.com/tendermint/tendermint/blob/5a39f72aa8288696d7284e56e3f630ab721a9f9a/evidence/pool.go#L149) directly after the start of the function it is claimed that "We have already verified this piece of evidence - no need to do it again",
- several lines later, [verify](https://github.com/tendermint/tendermint/blob/5a39f72aa8288696d7284e56e3f630ab721a9f9a/evidence/pool.go#L164) is called.

At the same time, all these verification checks at this point seem to  be optimizations to shield the evidence pool against evidence  that is obviously invalid, because the final verdict about the  evidence is done when the evidence is put into a block and  executed.

### Key

It is unclear how the [key](https://github.com/tendermint/tendermint/blob/5a39f72aa8288696d7284e56e3f630ab721a9f9a/evidence/pool.go#L692) is constructed. Why is `height` encoded separately? It seems that the used library ensures that the encoding maintains ordering. Such invariants between keys and data should be made explicit.

### Verify

[...]

PR #324, opened 2021-07-16
evidence: add section explaining evidence

cmwaters:
Closes: #295

### DIFF
--- a/spec/consensus/evidence.md
+++ b/spec/consensus/evidence.md
@@ -2,8 +2,8 @@
 
 Evidence is an important component of Tendermint's security model. Whilst the core
 consensus protocol provides correctness gaurantees for state machine replication
-that can tolerate up to 1/3 failures, the evidence system looks to detect and
-gossip byzantine faults of greater than 1/3 in power. It is worth noting that
+that can tolerate less than 1/3 failures, the evidence system looks to detect and
+gossip byzantine faults whose combined power is greater than  or equal to 1/3. It is worth noting that
 the evidence system is designed purely to detect possible attacks, gossip them,
 commit them on chain and inform the application running on top of Tendermint.
 Evidence in itself does not punish "bad actors", this is left to the discretion
@@ -28,7 +28,7 @@ replication of state across all nodes, a validator tries to convince some subset
 of nodes to commit one block whilst convincing another subset to commit a
 different block. This is achieved by double voting (hence
 `DuplicateVoteEvidence`). A successful duplicate vote attack requires greater
-than 1/3 voting power and a network partition between the aforementioned
+than 1/3 voting power and a (temporary) network partition between the aforementioned
 subsets. This is because in consensus, votes are gossiped around. When a node
 observes two conflicting votes from the same peer, it will use the two votes of
 evidence and begin gossiping this evidence to other nodes.
@@ -68,7 +68,7 @@ type LightClientAttackEvidence struct {
 If a node receives evidence, it will first try to verify it, then persist it.
 Evidence of byzantine behavior should only be committed once (uniqueness) and
 should be committed within a certain period from the point that it occurred
-(timely).Timelines is defined by the `EvidenceParams`: `MaxAgeNumBlocks` and
+(timely). Timelines is defined by the `EvidenceParams`: `MaxAgeNumBlocks` and
 `MaxAgeDuration`. In Proof of Stake chains where validators are bonded, evidence
 age should be less than the unbonding period so validators still can be
 punished. Given these two propoerties the following initial checks are made.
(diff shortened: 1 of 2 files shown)


spec/consensus/evidence.md
```diff
@@ -0,0 +1,196 @@
+# Evidence
+
+Evidence is an important component of Tendermint's security model. Whilst the core
+consensus protocol provides correctness gaurantees for state machine replication
+that can tolerate up to 1/3 failures, the evidence system looks to detect and
```
josef-widder, line 5:
```suggestion
that can tolerate less than 1/3 failures, the evidence system looks to detect and
```

spec/consensus/evidence.md
```diff
@@ -0,0 +1,196 @@
...
+gossip byzantine faults of greater than 1/3 in power. It is worth noting that
```
josef-widder, line 6:
```suggestion
gossip byzantine faults whose combined power is greater than  or equal to 1/3. It is worth noting that
```

spec/consensus/evidence.md
```diff
@@ -0,0 +1,196 @@
...
+## Detection
+
+### Equivocation
+
+Equivocation is the most fundamental of byzantine faults. Simply put, to prevent
+replication of state across all nodes, a validator tries to convince some subset
+of nodes to commit one block whilst convincing another subset to commit a
+different block. This is achieved by double voting (hence
+`DuplicateVoteEvidence`). A successful duplicate vote attack requires greater
+than 1/3 voting power and a network partition between the aforementioned
```
josef-widder, line 31:
```suggestion
than 1/3 voting power and a (temporary) network partition between the aforementioned
```

spec/consensus/evidence.md
```diff
@@ -0,0 +1,196 @@
...
+evidence and begin gossiping this evidence to other nodes.
+
+```go
+type DuplicateVoteEvidence struct {
+    VoteA Vote
+    VoteB Vote
+
+    // and abci specific fields
+}
+```
```
josef-widder, line 43: I wonder whether it would make sense to put the verification rules right here. If you prefer to keep them in separate section, it would be great if you could put a link here to section `### DuplicateVoteEvidence`. (The same comment applies to light client evidence below)

spec/consensus/evidence.md
```diff
@@ -0,0 +1,196 @@
...
+    CommonHeight int64
+}
+```
+
+## Verification
+
+If a node receives evidence, it will first try to verify it, then persist it.
+Evidence of byzantine behavior should only be committed once (uniqueness) and
+should be committed within a certain period from the point that it occurred
+(timely).Timelines is defined by the `EvidenceParams`: `MaxAgeNumBlocks` and
```
josef-widder, line 71:
```suggestion
(timely). Timelines is defined by the `EvidenceParams`: `MaxAgeNumBlocks` and
```

josef-widder (commented):
Thanks! Looks great and informative. I have just added some minor comments.

spec/consensus/evidence.md
cmwaters: My aim was to structure the document in terms of lifecycle events i.e. detection, validation, committing and application as opposed to by evidence type

josef-widder: I understand. Then please just add a link to the verification rules. For me, just the data structures without a mention what constitutes valid evidence seems incomplete information, even at this point.

josef-widder (approved):
Again, thanks for adding these explanations!

spec/consensus/evidence.md
```diff
@@ -0,0 +1,199 @@
...
+of the application. A common form of punishment is slashing where the validators
+that were caught violating the protocol have all or a portion of their voting
+power removed. Evidence, given the assumption that 1/3+ of the network is still
+byzantine, is susceptible to censorship and should therefore be considered added
+security on a "best effort" basis.
+
+This document walks through the various forms of evidence, how they are detected,
+gossiped, verified and committed.
+
+> NOTE: Evidence here is internal to tendermint and should not be confused with
```
creachadair, line 19: Given that this ambiguity is important enough to call out, is there a more specific name we could use throughout to make the distinction clear? E.g., "protocol evidence" or "fault evidence", in contrast with "application evidence"?

The trouble with contextually overloaded terms—particularly ones as generic as "evidence"—is that the definition doesn't follow the usage (whereas the name does).

cmwaters: I find that "protocol evidence" or "fault evidence" probably wouldn't help to better distinguish whether it refers to application based evidence or Tendermint based evidence (apps also have protocols).  

Evidence is definitely a broad term. I've seen applications use the term "fraud proofs" to mean incorrect state transitions (in regards to business logic) and If I was an authoritarian dictator I would force them to keep to that term. 

Maybe we could try introduce a distinction by using terms like fault vs fraud to denote the different layers (tm and app)

josef-widder: Thanks for the comment. I think naming is very important. For the scope of this specific document, I feel that replacing "evidence" with "fault evidence" will just clutter the text. At the same time, in this document we do not consider any other kinds of evidence. 

We could rename the title to `# Evidence of Validator Misbehavior`, but then stick with the term "evidence" for the rest of the document.

PR #327, opened 2021-08-02
ABCI++ markdown integration into Tendermint (for spec)

ValarDragon:
This is a continuation of #283. This recombines the changes in an easier to view format, namely:

v0.md contains a transcription of the Tendermint BFT paper into markdown directly.
v1.md contains how ABCI hooks in.
v2.md contains a conversion to single-threaded ABCI++, with no optimization for splitting up verification of the header and block data
v3.md contains the optimization for splitting up verification of the header and block data
v4.md contains the multi-threaded optimization for delaying the end of ProcessProposal.

The structure for reviewing this is to review it commit by commit. Each commit shows the update to one file, so that it may be incrementally reviewed. (As a side effect, it also involves copying the file as well)

The final commit here is resolving the remaining unaddressed comments from #283


josef-widder:
I am going to review this week.

ValarDragon:
Hrmm, can we-reopen / merge this, and then iteratively update the markdown spec per review? Would rather it be in the branch then lost in a PR.

I don't have perms to re-open the PR tho

spec/abci++/v0.md
```diff
@@ -0,0 +1,155 @@
...
+This translates the latex code for Tendermint consensus from the Tendermint paper into markdown.
+
+### Initialization
+
+```go
+h_p ← 0
+round_p ← 0
+step_p is one of {propose, prevote, precommit}
+decision_p ← Vector()
+lockedValue_p ← nil
```
milosevic, line 12:
```suggestion
lockedRound_p ← -1
lockedValue_p ← nil
```

spec/abci++/v0.md
```diff
@@ -0,0 +1,155 @@
...
+    if valid(v) ∧ (lockedRound_p = −1 ∨ lockedValue_p = v) {
+        broadcast ⟨PREVOTE, h_p, round_p, id(v)⟩ 
+    } else {
+        broadcast ⟨PREVOTE, h_p, round_p, nil⟩ 
+    }
+    step_p ← prevote
+}
+```
+
+In the case where the node is locked on a round, the following is ran:
```
milosevic, line 51: In this case, the same as above, local node can be locked or not, but the importance is on the fact that proposed value need to be supported by the quorum of `PREVOTE` messages as `validRound >= 0`.

spec/abci++/v0.md
```diff
@@ -0,0 +1,155 @@
...
+    step_p ← prevote
+}
+```
+
+### Prevote timeout
+
+Upon receiving 2f + 1 prevotes, setup a timeout.
+
+```go
+upon 2f + 1 ⟨PREVOTE, h_p, vr, -1⟩ with step_p = prevote for the first time, do {
```
milosevic, line 71:
```suggestion
upon 2f + 1 ⟨PREVOTE, h_p, vr, *⟩ with step_p = prevote for the first time, do {
```

milosevic: `*` is wildcard in this case. We can read this as "upon any set of prevote messages for the current round, trigger `timeoutPrevote`.

spec/abci++/v0.md
```diff
@@ -0,0 +1,155 @@
...
+    step_p ← propose
+    if proposer(h_p, round_p) = p {
+        if validValue_p != nil {
+            proposal ← validValue_p
+        } else {
+            proposal ← getValue()
+        }
+        broadcast ⟨PROPOSAL, h_p, round_p, proposal, validRound_p⟩
+    } else {
+        schedule OnTimeoutPropose(h_p,round_p) to be executed after timeoutPropose(round_p)
```
milosevic, line 31: We are lacking logic for `OnTimeoutPropose`.

spec/abci++/v0.md
```diff
@@ -0,0 +1,155 @@
...
+```
+
+### Receiving enough prevotes to precommit
+
+The following code is ran upon receiving 2f + 1 prevotes for the same block
+
+```go
+upon ⟨PROPOSAL, h_p, round_p, v, *⟩
+  from proposer(h_p, round_p)
+  AND 2f + 1 ⟨PREVOTE, h_p, vr, id(v)⟩
```
milosevic, line 94:
```suggestion
  AND 2f + 1 ⟨PREVOTE, h_p, round_p, id(v)⟩ 
```

spec/abci++/v1.md
```diff
@@ -0,0 +1,162 @@
...
+}
+```
+
+### ReceiveProposal
+
+In the case where the local node is not locked on any round, the following is ran:
+
+```go
+upon ⟨PROPOSAL, h_p, round_p, v, −1) from proposer(h_p, round_p) while step_p = propose do {
+    if valid(v) ∧ (lockedRound_p = −1 ∨ lockedValue_p = v) {
```
milosevic, line 44: We should clarify the semantic of `valid(v)` function, i.e., is it done purely at the Tendermint level (basically block header verification), or we assume it requires also application level checks.

spec/abci++/v2.md
```diff
@@ -0,0 +1,180 @@
...
+    step_p ← propose
+    if proposer(h_p, round_p) = p {
+        if validValue_p != nil {
+            proposal ← validValue_p
+        } else {
+            txdata ← mempool.GetBlock()
+            // getUnpreparedBlockProposal takes tx data, and fills in the unprepared header data
+            unpreparedProposal ← getUnpreparedBlockProposal(txdata)
+            // ABCI++: the proposer may reorder/update transactions in `unpreparedProposal`
+            proposal ← ABCI.PrepareProposal(unpreparedProposal)
```
milosevic, line 32: Might be good idea adding note that `PrepareProposal` is adding latency on the critical path, and therefore, `timeoutPropose` would need to be updated accordingly. One idea would be allowing application to provide some hint to the Tendermint regarding duration of `PrepareProposal` so the `timeoutPropose` could be estimated more correctly.

spec/abci++/v2.md
```diff
@@ -0,0 +1,180 @@
...
+function startRound(round) {
+    round_p ← round
+    step_p ← propose
+    if proposer(h_p, round_p) = p {
+        if validValue_p != nil {
+            proposal ← validValue_p
+        } else {
+            txdata ← mempool.GetBlock()
+            // getUnpreparedBlockProposal takes tx data, and fills in the unprepared header data
+            unpreparedProposal ← getUnpreparedBlockProposal(txdata)
```
milosevic, line 30: I guess this is the same as `getBlockProposal(txdata)` in v0. Why we use different function name here?

spec/abci++/v2.md
```diff
@@ -0,0 +1,180 @@
...
+}
+```
+
+### ReceiveProposal
+
+In the case where the local node is not locked on any round, the following is ran:
+
+```go
+upon ⟨PROPOSAL, h_p, round_p, v, −1) from proposer(h_p, round_p) while step_p = propose do {
+    if valid(v) ∧ ABCI.ProcessProposal(h_p, v).accept ∧ (lockedRound_p = −1 ∨ lockedValue_p = v) {
```
milosevic, line 47: I would suggest storing the result of `ABCI.ProcessProposal` in a local variable and using it at line 52, as this presentation might suggest two ABCI calls.

spec/abci++/v2.md
```diff
@@ -0,0 +1,180 @@
...
+        broadcast ⟨PREVOTE, h_p, round_p, id(v)⟩ 
+    } else {
+        broadcast ⟨PREVOTE, h_p, round_p, nil⟩ 
+        // Include any slashing evidence that may be sent in the process proposal response
+        for evidence in ABCI.ProcessProposal(h_p, v).evidence_list {
```
milosevic, line 52: I am not sure to see benefit of having evidence handling at this level. I guess we want to signal that evidences will be provided by the application earlier to Tendermint (compared to the current version where it is after block is committed), right?

spec/abci++/v2.md
```diff
@@ -0,0 +1,180 @@
...
+```
+
+### Upon Receiving 2f + 1 precommits
+
+The following code is ran upon receiving 2f + 1 precommits for the same block
+
+```go
+upon ⟨PROPOSAL, h_p, r, v, *⟩
+  from proposer(h_p, r)
+  AND 2f + 1 ⟨ PRECOMMIT, h_p, r, id(v)⟩
```
milosevic, line 168: I guess we need also precommit_extension field as part of PRECOMMIT message?

spec/abci++/v3.md
```diff
@@ -0,0 +1,201 @@
...
+    }
+}
+```
+
+### ReceiveProposal
+
+In the case where the local node is not locked on any round, the following is ran:
+
+```go
+upon ⟨PROPOSAL, h_p, round_p, v_header, −1) from proposer(h_p, round_p) while step_p = propose do {
```
milosevic, line 46: From semantic perspective this presentation is not ideal as upon rules are atomic, so blocking within upon rule prevents any other rule being triggered. If we want to separate sending/receiving proposal_header from the proposal_data, we should probably introduce two message types.

spec/abci++/v3.md
```diff
@@ -0,0 +1,201 @@
...
+    }
+}
+```
+
+### Receiving enough prevotes to precommit
+
+The following code is ran upon receiving 2f + 1 prevotes for the same block
+
+```go
+upon ⟨PROPOSAL, h_p, round_p, v, *⟩
```
milosevic, line 129: PROPOSAL message contains full block proposal in this rule, while above it is only header. As mentioned above, we can make it clear by having in addition to PROPOSAL, also PROPOSAL_HEADER message and then have a separate rule for PROPOSAL_HEADER message.

spec/abci++/v4.md
```diff
@@ -0,0 +1,199 @@
...
+
+### ReceiveProposal
+
+In the case where the local node is not locked on any round, the following is ran:
+
+```go
+upon ⟨PROPOSAL, h_p, round_p, v, −1) from proposer(h_p, round_p) while step_p = propose do {
+    if valid(v) ∧ ABCI.VerifyHeader(h_p, v.header) ∧ (lockedRound_p = −1 ∨ lockedValue_p = v) {
+        // We fork process proposal into a parallel process
+        Fork ABCI.ProcessProposal(h_p, v)
```
milosevic, line 48: I would prefer using slightly different semantics to signal it is asynchronous computation: we can distinct between synchronous ABCI calls where we are synchronously waiting for result, or asynchronous where we are essentially sending a request and will receive reply later (eventually). I don't have strong preference how we can capture this, maybe using Future or Promise as a return value.  Then instead of Join we can have upon rule contains condition based on response. The benefit is that there is no blocking wait inside the upon rule as it semantically prevents other upon rules being triggered.

spec/abci++/v4.md
```diff
@@ -0,0 +1,199 @@
...
+
+```go
+upon ⟨PROPOSAL, h_p, round_p, v, *⟩
+  from proposer(h_p, round_p)
+  AND 2f + 1 ⟨PREVOTE, h_p, vr, id(v)⟩ 
+while valid(v) ∧ step_p >= prevote for the first time do {
+    if (step_p = prevote) {
+        lockedValue_p ← v
+        lockedRound_p ← round_p
+        processProposalOutput ← Join ABCI.ProcessProposal(h_p, v)
```
milosevic, line 116: I guess the invariant here is that we should not commit a block proposal before being processed successfully, right?

milosevic (approved):
I have left few comments but they don't necessarily need to be addressed as part of this PR. I would suggest to merge this PR and continue working in smaller PRs. Thanks a lot @ValarDragon for amazing work here!

PR #328, opened 2021-08-06
statesync: new messages for gossiping consensus params

cmwaters:
To be able to run state sync purely through p2p, the state provider, which has the responsibility of building and verifying tendermint state requires the ability to fetch consensus params. This PR introduces the messages needed.

### DIFF
--- a/proto/statesync/types.proto
+++ b/proto/statesync/types.proto
@@ -13,6 +13,8 @@ message Message {
     ChunkResponse      chunk_response       = 4;
     LightBlockRequest  light_block_request  = 5;
     LightBlockResponse light_block_response = 6;
+    ParamsRequest      params_request       = 7;
+    ParamsResponse     params_response      = 8;
   }
 }
 
@@ -46,4 +48,13 @@ message LightBlockRequest {
 
 message LightBlockResponse {
   tendermint.types.LightBlock light_block = 1;
+}
+
+message ParamsRequest {
+  uint64 height = 1;
+}
+
+message ParamsResponse {
+  uint64 height = 1;
+  tendermint.types.ConsensusParams consensus_params = 2;
 }
--- a/spec/p2p/messages/state-sync.md
+++ b/spec/p2p/messages/state-sync.md
@@ -119,7 +119,7 @@ A reciever to the request will use the state store to fetch the consensus params
 
 ### Message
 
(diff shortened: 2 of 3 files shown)


spec/p2p/messages/state-sync.md
```diff
@@ -13,6 +13,7 @@ State sync has three distinct channels. The channel identifiers are listed below
 | SnapshotChannel   | 96     |
 | ChunkChannel      | 97     |
 | LightBlockChannel | 98     |
+| ParamsChannel     | 99     |
```
tychoish, line 16: does adding this channel break the protocol?

cmwaters: No. If I'm running the new version and you're running the old version without this channel, then I simply don't send messages to you. That being said, if everyone is on the old version then I won't ever receive any response. In which case I'd have to revert back to using RPC state provider

marbar3778:
Does state sync still require the user to get a hash, height and genesis?

cmwaters:
Yeah a hash and height is still needed. Initialising a light client using the genesis hasn't been implemented yet.

spec/p2p/messages/state-sync.md
tychoish: If the statesync provider is configurable at runtime, and both are supported, then people can switch between them as needed, so that should be fine.

tychoish (approved):

spec/p2p/messages/state-sync.md
williambanfield: The text above this diff says 'state sync has three distinct channels'. Should now say four.

proto/statesync/types.proto
```diff
@@ -13,6 +13,8 @@ message Message {
     ChunkResponse      chunk_response       = 4;
     LightBlockRequest  light_block_request  = 5;
     LightBlockResponse light_block_response = 6;
+    ParamsRequest      params_request       = 7;
+    ParamsResponse     params_response      = 8;
```
marbar3778, line 17: I think this is fine for now, but I believe these messages and the channels should live in light client. This would allow the light client to be used in a general way within the p2p layer.

marbar3778: talked with @cmwaters over slack about approaches for this. He mentioned this could be done in a non breaking way after 0.35. resolving this

marbar3778 (approved):

spec/p2p/messages/state-sync.md
cmwaters: Good spot 👍

PR #333, opened 2021-08-18
Update supervisor_001_draft.md

goral09:
If the only node in the *FullNodes* set is the primary, that was just deemed faulty, we can't find honest primary.

### DIFF
--- a/spec/light-client/supervisor/supervisor_001_draft.md
+++ b/spec/light-client/supervisor/supervisor_001_draft.md
@@ -237,7 +237,7 @@ Replace_Primary(root-of-trust LightBlock)
         - pick a new secondary *nsec* while ensuring [LC-INV-ROOT-AGREED.1]
         - that is, we need to ensure that root-of-trust = FetchLightBlock(nsec, root-of-trust.Header.Height)
 - Expected precondition
-    - *FullNodes* is nonempty
+    - *FullNodes \ primary* is nonempty
 - Expected postcondition
     - *primary* is moved to *FaultyNodes*
     - a secondary *s* is moved from *Secondaries* to primary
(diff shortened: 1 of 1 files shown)


spec/light-client/supervisor/supervisor_001_draft.md
```diff
@@ -237,7 +237,7 @@ Replace_Primary(root-of-trust LightBlock)
         - pick a new secondary *nsec* while ensuring [LC-INV-ROOT-AGREED.1]
         - that is, we need to ensure that root-of-trust = FetchLightBlock(nsec, root-of-trust.Header.Height)
 - Expected precondition
-    - *FullNodes* is nonempty
+    - *FullNodes \ primary* is nonempty
```
josef-widder, line 240: [LC-DATA-PEERS.1] says that the set of nodes is "partitioned", that is, that primary is not part of the set FullNodes. I think it is currently used consistently throughout the spec. I think we may clarify that in [LC-DATA-PEERS.1] rather than doing this change. What do you think?

josef-widder (changes requested):
We should clarify the relation of primary to the set FullNodes

spec/light-client/supervisor/supervisor_001_draft.md
goral09: I see - so the *FullNodes* _doesn't_ include the primary OR the secondaries?

josef-widder: yes. it is basically a pool of "unused" full nodes.

goral09: Now, when you said it, I can see that [LC-DATA-PEERS.1] could be interpreted this way. I will clarify that rule then instead. Thanks.

goral09: @josef-widder , changed in https://github.com/tendermint/spec/pull/333/commits/e3f5a503fbf24fb4cb06365170665d465d7041cd.

josef-widder (approved):
Thanks for the clarification!

PR #335, opened 2021-08-18
Update supervisor_001_draft.md

goral09:
We can't move a node to *Secondaries* if it's already there.

### DIFF
--- a/spec/light-client/supervisor/supervisor_001_draft.md
+++ b/spec/light-client/supervisor/supervisor_001_draft.md
@@ -254,7 +254,7 @@ Replace_Secondary(addr Address, root-of-trust LightBlock)
     - maintain [LCD-INV-ROOT-AGREED.1], that is,
     ensure root-of-trust = FetchLightBlock(nsec, root-of-trust.Header.Height)
 - Expected precondition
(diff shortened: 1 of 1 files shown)


spec/light-client/supervisor/supervisor_001_draft.md
```diff
@@ -254,7 +254,7 @@ Replace_Secondary(addr Address, root-of-trust LightBlock)
     - maintain [LCD-INV-ROOT-AGREED.1], that is,
     ensure root-of-trust = FetchLightBlock(nsec, root-of-trust.Header.Height)
 - Expected precondition
-    - *FullNodes* is nonempty
+    - *FullNodes \ Secondaries* is nonempty
```
josef-widder, line 257: As in the other PR, [LC-DATA-PEERS.1] defines that FullNodes and Secondaries are disjoint sets. It is better to clarify this than to do this change, I believe.

josef-widder (changes requested):
I think it is better to clarify the definition of the set FullNodes

goral09:
Closing in favor of #333 which will clarify the [LC-DATA-PEERS.1] rule.

Issue #316, opened 2021-07-14
v0.1.0 todolist


marbar3778:
## Summary

This is a todo list of thing to do prior to versioning the spc

- [ ] informal read through
- [ ] IG read through
- [ ] Link fixing
- [ ] proto files update
- [ ] changelog update

marbar3778:
From the IG side, the breakdown for reviewing the spec is:

- core by callum, sam and I
- rpc: marko
- light-client: callum
- evidence: callum
- abci: will and sam
- p2p: sam and bez
- consensus: will and sam

cmwaters:
For the p2p/pex section, we should look to close this issue as well: https://github.com/tendermint/spec/issues/287

tessr:
> I'm not sure if adding "if using the legacy stack" is really a good edit to make

If only we _already_ had spec versioning, we could just make the edits cleanly and refer people back to an older version if they wanted to know about the legacy stack

marbar3778:
the p2p spec is the least implementation agnostic part of the spec. It might make sense to call it good now, and open an issue to make it agnostic.

tychoish:
I think that makes sense as a strategy. there's definitely a lot of work to do in the p2p area of the spec for the next release, but I don't think it's bad to release in its current form, and I think we'll do net less work if don't revise this part of the spec in this release cycle. 

Also I read through the core sections, and didn't see anything that jumped out at me though I did clean up a couple of little things: https://github.com/tendermint/spec/pull/332.

I gave a look through the consensus. I think I could probably normalize the style (some) or do an edit for clarity (similar to the paragraph I rewrote in the core section,) but I don't know if I have context from the code that would be useful in describing where the spec is out of sync. I suspect that @williambanfield might have more useful context here.

PR #336, opened 2021-08-19
abci: clarify what abci stands for

williambanfield:
[Rendered](https://github.com/tendermint/spec/blob/wb/abci-readme/spec/abci/README.md)
Part of: #316

### DIFF
--- a/spec/abci/README.md
+++ b/spec/abci/README.md
@@ -14,7 +14,7 @@ _methods_, each with a corresponding `Request` and `Response`message type.
 To perform state-machine replication, Tendermint calls the ABCI methods on the 
 ABCI application by sending the `Request*` messages and receiving the `Response*` messages in return.
 
-All ABCI messages and methods are defined in protocol buffers. 
+All ABCI messages and methods are defined in [protocol buffers](https://github.com/tendermint/spec/blob/master/proto/abci/types.proto). 
 This allows Tendermint to run with applications written in many programming languages.
 
 This specification is split as follows:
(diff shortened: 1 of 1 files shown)


spec/abci/README.md
```diff
@@ -7,14 +7,15 @@ parent:
...
 ABCI is the interface between Tendermint (a state-machine replication engine)
 and your application (the actual state machine). It consists of a set of
-_methods_, where each method has a corresponding `Request` and `Response`
-message type. Tendermint calls the ABCI methods on the ABCI application by sending the `Request*`
-messages and receiving the `Response*` messages in return.
+_methods_, each with a corresponding `Request` and `Response`message type. 
+To perform state-machine replication, Tendermint calls the ABCI methods on the 
+ABCI application by sending the `Request*` messages and receiving the `Response*` messages in return.
 
-All message types are defined in a [protobuf file](https://github.com/tendermint/tendermint/blob/master/proto/tendermint/abci/types.proto).
```
williambanfield, line 16: Removed in favor of not linking against master of a different repository.

creachadair: This makes sense, but it would be nice if there were some indication to the reader of where to go for the definitions. Do we have a place on the docs site we could publish the release version of the proto file? 

Or if that isn't practical, maybe it's still worth saying something like

> All message types are defined in `abci/types.proto` in the [tendermint repository](https://github.com/tendermint/tendermint).

or words to that effect?

williambanfield: Realized that they're also in this repo! linked to them here.

creachadair (approved):

PR #337, opened 2021-08-19
abci: clarify connection use in-process

williambanfield:
[rendered](https://github.com/tendermint/spec/blob/wb/abci-connections/spec/abci/abci.md)

For: #316

### DIFF
--- a/spec/abci/abci.md
+++ b/spec/abci/abci.md
@@ -17,18 +17,18 @@ opens four connections to the application for ABCI methods. The connections each
 handle a subset of the ABCI method calls. These subsets are defined as follows:
 
 #### **Consensus** connection
+* Driven by a consensus protocol and is responsible for block execution.
 * Handles the `InitChain`, `BeginBlock`, `DeliverTx`, `EndBlock`, and `Commit` method
 calls.
-* Driven by a consensus protocol and is responsible for block execution.
 #### **Mempool** connection
-* Handles the `CheckTx` calls.
 * For validating new transactions, before they're shared or included in a block.
+* Handles the `CheckTx` calls.
 #### **Info** connection
-* Handles the `Info` and `Query` calls.
 * For initialization and for queries from the user.
+* Handles the `Info` and `Query` calls.
 #### **Snapshot** connection
-* Handles the `ListSnapshots`, `LoadSnapshotChunk`, `OfferSnapshot`, and `ApplySnapshotChunk` calls.
 * For serving and restoring [state sync snapshots](apps.md#state-sync).
+* Handles the `ListSnapshots`, `LoadSnapshotChunk`, `OfferSnapshot`, and `ApplySnapshotChunk` calls.
 
 Additionally, there is a `Flush` method that is called on every connection,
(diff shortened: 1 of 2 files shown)


spec/abci/abci.md
```diff
@@ -5,27 +5,32 @@ title: Method and Types
...
-The mempool connection is for validating new transactions, before they're
-shared or included in a block.
-
-The info connection is for initialization and for queries from the user.
-
-The snapshot connection is for serving and restoring [state sync snapshots](apps.md#state-sync).
+## Connections
+
+ABCI applications can run either within the same _process_ as the Tendermint 
+state-machine replication engine _or_ as a separate process from the state-machine
```
creachadair, line 11: Minor nit: It feels like we're emphasizing different things here.

```suggestion
ABCI applications can run either within the _same_ process as the Tendermint 
state-machine replication engine, or as a _separate_ process from the state-machine
```

spec/abci/abci.md
```diff
@@ -5,27 +5,32 @@ title: Method and Types
...
+replication engine. When run within the same process, Tendermint will call the ABCI
+application methods directly as typical Go method calls. When Tendermint and the
+ABCI application are run as separate processes, Tendermint opens a set of connections
+to the application.
+
+When Tendermint and the application are run as separate processes, Tendermint opens
+four connections for ABCI methods. The connections each handle a subset of the ABCI
+method calls. They are defined as follows:
```
creachadair, line 19: There is some duplication between these two paragraphs that I'd suggest collapsing.
```suggestion
application methods directly as Go method calls.

When Tendermint and the ABCI application are run as separate processes, Tendermint
opens four connections to the application for ABCI methods. The connections each 
handle a subset of the ABCI method calls. These subsets are defined as follows:
```

creachadair: Not directly related to this change, but it seems odd to have all of these on separate connections. Is there some rationale for that, which we could summarize? (Made-up hypothetical: "This allows the application to more easily handle different parts of the consensus workflow concurrently" or something like that?)

spec/abci/abci.md
```diff
@@ -5,27 +5,32 @@ title: Method and Types
...
+
+#### **Consensus** connection
+* Handles the `InitChain`, `BeginBlock`, `DeliverTx`, `EndBlock`, and `Commit` method
```
creachadair, line 22: I'd suggest putting the "summary" explanation first, and the list of methods second, in each case.

spec/abci/abci.md
williambanfield: reread this just now and noticed the same thing.

spec/abci/abci.md
williambanfield: done.

spec/abci/abci.md
williambanfield: I'm not 100% sure I know the answer. this document is how I learned about how this works. I'd recommend asking in the tendermint slack channel.

tac0turtle (approved):
looks good, thanks for the additions

Issue #338, opened 2021-08-24
Generate the protocol message descriptions from protobuf to markdown


williambanfield:
We currently check in the protocol buffer messages that define the defined the tendermint wire protocol to this repository. For example, the ABCI messages are defined in [spec/proto/abci/types.proto](https://github.com/tendermint/spec/blob/master/proto/abci/types.proto#L188). 
The protocol is also defined in markdown in the `spec` folder. For ABCI, this is defined in [spec/abci/abci.md](https://github.com/tendermint/spec/blob/master/spec/abci/abci.md#checktx). 

One of these two descriptions of the protocol should be the source of truth. I propose here, making the `types.proto` file into the source of truth. We can then use this file to generate the markdown specification using the protocol buffer definition. This will prevent errors when updating the prose version of the spec as it will track directly with the machine version of the spec.

A tool exists here that could be used or taken as inspiration for our implementation: https://github.com/pseudomuto/protoc-gen-doc

williambanfield:
# Quick Prototype

## Overview
Put in some work to determine what this may actually look like. 
I used the [protoc-gen-doc](https://github.com/pseudomuto/protoc-gen-doc) plugin to generate documentation.
The result is in the [wb/proto-doc-gen](https://github.com/tendermint/spec/tree/wb/proto-doc-gen) branch. 
This mechanism seems to work well enough, has a few shortcomings and some room for improvement.
To interact with, clone the branch and run `make markdown-proto-gen-abci`. 

## What was done

Copied the `InitChain` method documentation from the abci.md file into [abci/types.proto](https://github.com/tendermint/spec/blob/966330879f5ef767743755c881c39efdcfd46bb4/proto/tendermint/abci/types.proto#L362) as well as the `RequestInitChain` and `ResponseInitChain` documentation to the respective methods.

Added  a small template file under [spec/abci/templates/abci.tmpl](https://github.com/tendermint/spec/blob/966330879f5ef767743755c881c39efdcfd46bb4/spec/abci/templates/abci.tmpl#L1). 

Added a [Makefile](https://github.com/tendermint/spec/blob/966330879f5ef767743755c881c39efdcfd46bb4/Makefile#L1) that calls the `protoc-doc-gen` plugin and generates the documentation.

Checked in the resulting documentation under the [spec/abci/abci_generated.md](https://github.com/tendermint/spec/blob/966330879f5ef767743755c881c39efdcfd46bb4/spec/abci/abci_generated.md#L59) file.

## Discussion

[...]

PR #344, opened 2021-08-31
proto: move proto files under the correct directory related to their package name

williambanfield:
The protos defined in the `/proto` directory all declare themselves to be under a `tendermint.` package. Additionally, the protos import each other under a `/tendermint/` folder. E.g. [`/tendermint/types/block.proto`](https://github.com/tendermint/spec/blob/master/proto/blockchain/types.proto#L6). This change moves the protos to be under a `/tendermint` folder to align their directory with the import names.


creachadair (approved):

williambanfield:
> 1. The proto files define `go_package` names rooted in the `tendermint/tendermint` repository, not the `spec` repo.

I noticed this as well while preparing this [quick prototype](https://github.com/tendermint/spec/issues/338#issuecomment-909612229) for the markdown generator. 
What I think is the right call is to remove the `go_package` option. We can pass a go package name to the `protoc` tool when we generate the `.go` files.  This would make the protos more generic and, in my view, clarify for users that they are not strictly go related. (I'm not sure if other proto tools choke on these field when, say, generating rust code?).

creachadair:
> What I think is the right call is to remove the `go_package` option. We can pass a go package name to the `protoc` tool when we generate the `.go` files. 

That's worth trying. In the past, however, I have had problems with errors from the Go proto compiler plugin if the source didn't define `go_package`. Older versions of the plugin didn't check—then at some point it started checking and I had to re-do my build workflow. Possibly gogo doesn't have this requirement—but if we are still interested in moving off gogo for the implementation repo, we should probably at least make sure both _could_ work.

> This would make the protos more generic and, in my view, clarify for users that they are not strictly go related. (I'm not sure if other proto tools choke on these field when, say, generating rust code?).

Other language plugins to care about the `go_package` option. Having to specify the package manually in the build isn't too bad, but it might be something we should document and/or provide scripts to do (or perhaps go:generate rules).

marbar3778 (approved):

williambanfield:
> That's worth trying. In the past, however, I have had problems with errors from the Go proto compiler plugin if the source didn't define go_package. Older versions of the plugin didn't check—then at some point it started checking and I had to re-do my build workflow. Possibly gogo doesn't have this requirement—but if we are still interested in moving off gogo for the implementation repo, we should probably at least make sure both could work.


That's good to know. I just tried this in a new branch ([wb/protogen-no-option](https://github.com/tendermint/spec/tree/wb/protogen-no-option)) to check. Created this [generator script to verify](https://github.com/tendermint/spec/blob/wb/protogen-no-option/proto/protoc_gen_script.sh) and stripped out the `option` from the protos. It seems to work without the option as long as you specify the mapping manually. Perhaps this is a bit of a hassle but checking in a simple script or using a more advanced build tool could make this not miserable. Certainly possible there are issues I'm not yet seeing with this approach.

PR #348, opened 2021-09-23
proto: abci++ changes

marbar3778:
## Description

Add proto files to reflect changes being made for abci++


marbar3778:
marking as R4R to get reviews in, but there will be further doc changes

proto/tendermint/abci/types.proto
```diff
@@ -21,20 +21,24 @@ import "gogoproto/gogo.proto";
...
+    RequestEcho                echo                  = 1;
+    RequestFlush               flush                 = 2;
+    RequestInfo                info                  = 3;
+    RequestInitChain           init_chain            = 4;
+    RequestQuery               query                 = 5;
+    RequestBeginBlock          begin_block           = 6 [deprecated = true];
+    RequestCheckTx             check_tx              = 7;
+    RequestDeliverTx           deliver_tx            = 8 [deprecated = true];
+    RequestEndBlock            end_block             = 9 [deprecated = true];
+    RequestCommit              commit                = 10;
```
cmwaters, line 33: Isn't this also being deprecated?

proto/tendermint/abci/types.proto
```diff
@@ -258,6 +297,32 @@ message ResponseApplySnapshotChunk {
...
+  tendermint.types.VoteExtension vote_extension = 1;
+}
+
+message ResponseVerifyVoteExtension {
+  Result result = 1;
+
+  enum Result {
+    UNKNOWN = 0;  // Unknown result, treat as ACCEPT by default
+    ACCEPT  = 1;  // Vote extension verified, include the vote
+    SLASH   = 2;  // Vote extension verification aborted, continue but slash validator
```
cmwaters, line 310: What does this mean for tendermint? i.e. what does tendermint do when it gets a `SLASH` response?

proto/tendermint/abci/types.proto
```diff
@@ -258,6 +297,32 @@ message ResponseApplySnapshotChunk {
...
+    REJECT  = 3;  // Vote extension invalidated
+  }
+}
+
+message ResponsePrepareProposal {
+  repeated bytes block_data = 1;
```
cmwaters, line 316: don't we also include the max block size in here?

cmwaters (commented):
Looking good so far. I assume there are more changes to come but I have a few comments in the meantime

proto/tendermint/abci/types.proto
marbar3778: This stays

cmwaters: What's the flow here? something like this:
1. Receive 2/3+ precommits
2. Execute `FinalizeBlock`
3. Persist the block and state changes
4. Execute `Commit`

proto/tendermint/abci/types.proto
marbar3778: Tendermint already knows the max size. we double check the response size when its returned.

cmwaters: https://github.com/tendermint/tendermint/blob/f408849112ab06ecafd6f4a41ed908e3f5b6907c/proto/tendermint/abci/types.proto#L133-L140

marbar3778: we send the size to the app. We dont need it back. What is the question?

cmwaters: lol my bad. Got the request and response mixed up. Ignore this

proto/tendermint/abci/types.proto
```diff
@@ -21,20 +21,24 @@ import "gogoproto/gogo.proto";
...
+    RequestBeginBlock          begin_block           = 6 [deprecated = true];
+    RequestCheckTx             check_tx              = 7;
+    RequestDeliverTx           deliver_tx            = 8 [deprecated = true];
+    RequestEndBlock            end_block             = 9 [deprecated = true];
+    RequestCommit              commit                = 10;
+    RequestListSnapshots       list_snapshots        = 11;
+    RequestOfferSnapshot       offer_snapshot        = 12;
+    RequestLoadSnapshotChunk   load_snapshot_chunk   = 13;
+    RequestApplySnapshotChunk  apply_snapshot_chunk  = 14;
+    RequestPrepareProposal     prepare_proposal      = 15;
```
creachadair, line 38: Tag 15 was removed in 82e4693c, we should probably not re-use it.

Was:

    RequestApplySnapshotChunk apply_snapshot_chunk = 15;

We can use a [reserved clause](https://developers.google.com/protocol-buffers/docs/proto3#reserved) to avoid old tags in the future. 

I'm not sure if this is the only one, but it's the only one I found offhand.

marbar3778: changing the number format in major releases, while not preferable, should be fine. Apps should refetch all the proto files when updating otherwise they could encounter other errors.

cmwaters: Yeah it seems like we actually removed `set_option`:
```
RequestSetOption          set_option           = 4;
```
and pushed everything up so really we should reserve 4 but I think we can get away with this (we already did this in 0.34).

cmwaters (approved):
LGTM 🥇

PR #349, opened 2021-09-23
proto: add tendermint go changes

marbar3778:
### Description

Changes in go tendermint no reflected in the spec


proto/tendermint/abci/types.proto
```diff
@@ -196,6 +196,10 @@ message ResponseCheckTx {
   string         codespace  = 8;
   string         sender     = 9;
   int64          priority   = 10;
+
+  // mempool_error is set by Tendermint.
+  // ABCI applictions creating a ResponseCheckTX should not set mempool_error.
```
cmwaters, line 201:
```suggestion
  // ABCI applications creating a ResponseCheckTX should not set mempool_error.
```

cmwaters: Also what's the context for adding this field if it isn't set by the application?

cmwaters (approved):

proto/tendermint/abci/types.proto
marbar3778: https://github.com/tendermint/tendermint/pull/6740

PR #358, opened 2021-11-09
update proto generation and testing pipelines

williambanfield:
This pull request aims to make it possible to generate, format, and lint the protos within this repo.
To accomplish that end, the Dockerfile containing common tools for building the tendermint protos has been moved into this repository. A related github workflow for building that dockerfile has been added to this repo (and removed from the tendermint repository). Additionally, a `buf.yaml` and `buf.gen.yaml` have been added for working with the `buf` tool that we use for compiling our protos. A 
`./third_party` directory with the gogoproto proto file has been added to the root of the repo so that the proto compiler can find this dependency.

The `go_package` variable has been removed from all of the protos. This variable was set to `github.com/tendermint/tendermint/{PACKAGE}` which does not make any sense since this is not the tendermint repository. This variable is now set when the protos are built within the tendermint repo. This does present the downside that users who wish to build this protos in go will also need to manually add this option. The community of users doing that seems likely to be small, and we can provide support to them with the scripts that are added in: https://github.com/tendermint/tendermint/pull/7269.


williambanfield:
clang formatting seems worse now?

proto/Dockerfile
```diff
@@ -1,24 +1,24 @@
...
-
-
-FROM alpine:3.14
-
-RUN echo 'http://dl-cdn.alpinelinux.org/alpine/edge/testing' >> /etc/apk/repositories && \
-  apk add --update --no-cache build-base clang-extra-tools curl git && \
-  rm -rf /var/cache/apk/*
-
-WORKDIR /work
+  mv $(go env | grep GOPATH | sed 's/GOPATH="\(.*\)"/\1/g')/bin/* /usr/local/bin/
```
creachadair, line 21:
```suggestion
  mv "$(go env GOPATH)"/bin/* /usr/local/bin/ 
```

creachadair (approved):
One simplification suggestion, otherwise LGTM.

proto/tendermint/abci/types.proto
```diff
@@ -1,8 +1,6 @@
 syntax = "proto3";
 package tendermint.abci;
 
-option go_package = "github.com/tendermint/tendermint/abci/types";
```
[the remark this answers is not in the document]
marbar3778: when this is imported into Tendermint, will it not cause package issues?

`package tendermint_abci`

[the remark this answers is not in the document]
marbar3778: whoops, never mind, I read the other PR now. 

The only downside of doing it this way is, users of our photo files will need to append option go_package as well, this could cause confusion. Before it was plug and play.

proto/Dockerfile
williambanfield: TIL `go env` can take an argument!

proto/tendermint/abci/types.proto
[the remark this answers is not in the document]
williambanfield: Unfortunately, this is really a downside of hosting the protos in the spec repo in general. If users were to try to generate these protos themselves using the old `option`, they would have created protos with import paths that do not match their local module name. As a result of this, they would have two options, both of which are not ideal.

Option 1. Rewrite the import paths locally, either before or after proto generation so that the paths match the local module. This is effectively what we are changing Tendermint to do. 
Option 2. Shove the protos into the `vendor` directory, where protos are allowed to have a name that does not match the module name, but is also a directory that go programmers frequently cede to the `go mod` tooling to manage.

Neither option is great, and I think we can all agree that using the module name `github.com/tendermint/tendermint` in this package is strictly incorrect. If this solution causes serious headache for users who are attempting to vendor these protos into their local `vendor` directory, we can reconsider, but it also seems odd to me that a user would do that and not simply import the generated protos that exist within the Tendermint repository instead since they are effectively equivalent.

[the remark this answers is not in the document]
marbar3778: The issue is that when the user imports the proto file into there proto file and generate their files the go import will be set to something else that isn't actually importing a file. This is how it is currently done for all go projects. 

It's fine if we break this, wanted to make note of it mainly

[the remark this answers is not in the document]
williambanfield: Makes sense, do you know of people doing this?

.github/workflows/proto-check.yml
```diff
@@ -0,0 +1,23 @@
+name: Proto Check
+# Protobuf runs buf (https://buf.build/) lint and check-breakage
+# This workflow is only run when a .proto file has been modified
```
creachadair, line 3:
```suggestion
# This workflow is only run when a file in the proto directory
# has been modified.
```

.github/workflows/proto-dockerfile.yml
```diff
@@ -0,0 +1,51 @@
+name: Build & Push TM Proto Builder
```
creachadair, line 1:
```suggestion
# This workflow (re)builds and pushes a Docker image containing the
# protobuf build tools used by the other workflows.
#
# When making changes that require updates to the builder image, you
# should merge the updates first and wait for this workflow to complete,
# so that the changes will be available for the dependent workflows.
#
# TODO(#7272): Update the target location of the builder image.

name: Build & Push TM Proto Builder
```

buf.gen.yaml
```diff
@@ -0,0 +1,13 @@
+# The version of the generation template.
+# Required.
```
creachadair, line 2:
```suggestion
# The version of the generation template (required).
```

buf.gen.yaml
```diff
@@ -0,0 +1,13 @@
...
+version: v1beta1
+
+# The plugins to run.
+plugins:
+  # The name of the plugin.
+  - name: gogofaster
+    # The the relative output directory.
+    out: proto
+    # Any options to provide to the plugin.
+    opt: Mgoogle/protobuf/timestamp.proto=github.com/gogo/protobuf/types,Mgoogle/protobuf/duration.proto=github.com/golang/protobuf/ptypes/duration,plugins=grpc,paths=source_relative
```
creachadair, line 13: Some suggestions to make the comments more explanatory:

```suggestion
plugins:
  - name: gogofaster
    # The output directory relative to the source files.
    out: proto
    # Set options to assign import paths to the well-known types
    # and to enable service generation.
    opt: Mgoogle/protobuf/timestamp.proto=github.com/gogo/protobuf/types,Mgoogle/protobuf/duration.proto=github.com/golang/protobuf/ptypes/duration,plugins=grpc,paths=source_relative
```

I'm actually not sure if the `out` option is relative to the source files (as my suggestion says) or to the repository root. Whichever it is, I think we should say. (The proto plugin is set to be `source_relative`, but I don't know if this is an option for the compiler or the plugin)

proto/Dockerfile
```diff
@@ -0,0 +1,24 @@
+# This Dockerfile defines a toolbox for use in linting, formatting, and compiling the Tendermint protos.
+
```
creachadair, line 2:
```suggestion
# This Dockerfile defines an image containing tools for linting, formatting,
# and compiling the Tendermint protos.

```

proto/Dockerfile
```diff
@@ -0,0 +1,24 @@
...
+FROM bufbuild/buf:latest as buf
+
+FROM alpine:3.14
+
+# Install a commonly used set of programs for use with our protos.
+# clang-extra-tools is included here because it provides clang-format, the program used for formatting the protos.
```
creachadair, line 8:
```suggestion
# clang-extra-tools is included here because it provides clang-format,
# used to format the .proto files.
```

proto/Dockerfile
```diff
@@ -0,0 +1,24 @@
...
+RUN echo 'http://dl-cdn.alpinelinux.org/alpine/edge/testing' >> /etc/apk/repositories && \
```
creachadair, line 9: I think if we base on `golang:1.17-alpine` we could avoid having to install Go here. More importantly, that would make it clear which version of Go we are getting.

(And with 1.16 or 1.17, we can assume `GO111MODULE=on` by default)

third_party/proto/gogoproto/gogo.proto
```diff
@@ -0,0 +1,147 @@
...
+	optional string casttype = 65007;
+	optional string castkey = 65008;
+	optional string castvalue = 65009;
+
+	optional bool stdtime = 65010;
+	optional bool stdduration = 65011;
+	optional bool wktpointer = 65012;
+
+	optional string castrepeated = 65013;
+}
```
creachadair, line 147: Unless this was already in the origin file, can we please have a newline at EOF?

buf.gen.yaml
williambanfield: `out` is relative to where the generate command is run. noted

PR #361, opened 2021-11-18
Update build image workflow.

creachadair:
Updates https://github.com/tendermint/tendermint/issues/7272.

The main effect of this change is to push the builder image to ghcr.io instead of dockerhub. For this change to take effect we will also need to update the Makefile, but we cannot do that until the image has been pushed since the CI workflow uses the existing tag.


creachadair:
The dead Markdown links are pre-existing. We should either fix those or disable the check. There are several dozen broken links, so this appears to have been a long-standing issue.

williambanfield (approved):
Very happy to ahve this improving

Issue #363, opened 2021-11-19
FIx markdown lint errors


creachadair:
The spec repo currently has a Markdown linter workflow that runs on every PR. There are several dozen broken links flagged by this linter.

We should either disable the linter, or fix up the broken links so that the baseline is clean.

PR #364, opened 2021-11-19
build: Update the builder image location.

creachadair:
A follow-up to #361. Updates https://github.com/tendermint/tendermint/issues/7272.

Point Makefile to ghcr.io instead of DockerHub.


creachadair:
The Markdown failure is (again) pre-existing. I filed #363 for this and will deal with it separately.

williambanfield (approved):

PR #365, opened 2021-11-19
RFC006: Semantic Versioning

cmwaters:
This RFC is the first in a set of items relating to Tendermint's upgrading story. It tries to capture prior internal dicussions held on versioning in Tendermint.

[Rendered](https://github.com/tendermint/spec/blob/callum/semver-rfc/rfc/006-semantic-versioning.md)


marbar3778 (commented):
high level question: there are multiple versions in Tendermint, will all of these converge on a single version or all have their own rules?

rfc/006-semantic-versioning.md
```diff
@@ -0,0 +1,76 @@
...
+
+In a more broader sense, these four users are inextricably linked and their concerns reflect that of the end user of the applications. The major conclusion that this RFC makes is that **the ability for chains to provide continual service is more important than the actual upgrade burden put on the developers of these chains**. What that means is that although it may be burdensome for application developers to upgrade Tendermint versions, this is far outweighed by the pain node operators experience in upgrading, by the pain that external clients have in accommodating changes and ultimately the pain of end users who are benefiting from the application. Developers have all the time in world to make these changes whilst for other user groups, upgrading can be considered part of this real-time critical path.
```
adizere, line 36: Very reasonable conclusion 💯

rfc/006-semantic-versioning.md
```diff
@@ -0,0 +1,76 @@
...
+2. Node Operators, those running node infrastructure, are predominantly concerned with downtime, complexity and frequency of upgrading and loss of data. They may be also concerned about changes that may break the scripts and tooling they use to supervise their nodes.
+3. External clients, for example, wallets and block explorers are those that perform any of the following:
+     - consume the RPC endpoints of nodes like `/block`
+     - subscribe to events that are streamed via websockets
+     - make queries to the indexer
+  
+    This set are concerned with chain upgrades which will impact their ability to query state and block data as well as broadcast transactions.
+  
+4. There is also a fourth emerging user: IBC module and relayers (although in many ways relayers can also be considered as external clients). The developers of IBC and consumers of their software are concerned about changes that may affect a chains ability to send arbitrary messages to another chain. Specifically, this can happen with a breaking change to the light client verification algorithm.
```
adizere, line 34: Thank you for acknowledging IBC mod & relayers as a class of users that are affected by changes in tm-go! It seems that this category of users is in some sense a superset of (3). It's not clear to me, however, what part of tm-go affects the light client verification algorithm: is it any changes to block headers, for example?

cmwaters: Exactly. Part of the feature set for 0.36 includes modifications to the `CommitSig` and `Commit`. The questions I have been asking the ibc-go team is that with such changes, could a chain that doesn't upgrade their IBC module still be able to verify packets (using the light client algorithm) with a chain that had upgraded. If not, then this Tendermint change effectively breaks IBC and chains across different versions won't be able to communicate. Changes to `ValidatorSet`, `Commit` or `Header` could all potentially break IBC.

cmwaters:
> high level question: there are multiple versions in Tendermint, will all of these converge on a single version or all have their own rules?

This is a good question. This RFC, in a way, ties in how the different sub versions within Tendermint i.e. (ABCI, Block, & P2P) relate to the overall version. There's probably still value in keeping these sub versions as a means of mapping to the different  user groups although I do like the idea of having one simple version to represent the state of Tendermint.

rfc/006-semantic-versioning.md
```diff
@@ -0,0 +1,76 @@
...
+- An external client should not need to modify their code to interact with a node's RPC endpoints. New endpoints can be added but existing ones can not be altered or removed.
+- Migrations should be automatic. Upgrading of one node can happen asynchronously with respect to other nodes (although agreement of a network-wide upgrade will happen synchronously via consensus).
+
+A minor version in Tendermint guarantees that for the entire cycle of that minor version:
+
+- All publicly available Go API's, for example in `node` or `abci` packages will not change in a breaking way that would require any consumer (not just application developers) to modify their code.
+- No breaking changes to the block protocol. This means that all block related data structures should not change in a way that breaks any of the hashes, the consensus engine or light client verification.
+- Upgrades should not result in any down-time (i.e. no migrations), nor should changes to the config files be required to continue with the existing behavior. Upgrading should simply be stopping the existing process, swapping the binary, and beginning the new process.
```
creachadair, line 53:
```suggestion
- Upgrades between minor versions may not result in any downtime (i.e., no migrations are required), nor require any changes to the config files to continue with the existing behavior. A minor version upgrade will require only stopping the existing process, swapping the binary, and starting the new process.
```

rfc/006-semantic-versioning.md
```diff
@@ -0,0 +1,76 @@
...
+
+These guarantees come into effect post 1.0.
```
creachadair, line 55:
```suggestion
These guarantees will come into effect at release 1.0.
```

creachadair (commented):
This all seems good to me, and although I have a few suggestions for things we could expand, I don't see any problems.

rfc/006-semantic-versioning.md
williambanfield: I think the distinguishing feature of this set of clients is cross-chain interoperability. Perhaps we can find a better name, but other clients do not need this type of interoperability, they just need to be able to parse out the relevant data structures. Clients performing cross-chain operations are in a special category of potentially requiring both chains to match completely, or nearly completely, which isn't a requirement of other applications like wallets etc.

rfc/006-semantic-versioning.md
```diff
@@ -0,0 +1,76 @@
...
+
+- 2021-11-19: Initial Draft
+
+## Author(s)
+
+- Callum Waters @cmwaters
+
+## Context
+
+We use versioning as an instrument to hold a set of promises to users and signal when such a set changes and how. In the conventional sense of a Go library, major versions signal that the public Go API’s have changed in a breaking way and thus require the users of such libraries to change their usage accordingly. Tendermint is a bit different in that there are multiple users: application developers (both in-process and out-of-process), node operators, and external clients. More importantly, both how these users interact with Tendermint and what's important to these users differs from how users interact and what they find important in a more conventional library.
```
williambanfield, line 13: To clarify, is this document solely about the Go project? If so, should it be in the Go repo? or is it about spec-specific changes? If we're talking about cross-chain data/ relayers etc, I would say this belongs here in the spec repo, but I wanted to make sure I was understanding correctly.

rfc/006-semantic-versioning.md
williambanfield: > Developers have all the time in world to make these changes whilst for other user groups, upgrading can be considered part of this real-time critical path.

I'm not totally sure I understand what this means. I think we're considering imposing costs in different areas of the ecosystem and it seems like the notion here is that having developers do extra legwork keeps overall costs of a change much lower by shielding external clients from too much additional work. 

I would add that developers don't exactly have _all the time in the world_ since they are trying to ship features and fix bugs, and costly upgrades keep them from satisfying user needs in other ways. I think what this is getting at is that operators, and clients to an extent, experience upgrades as _downtime_, whereas painful upgrades performed by developers create longer release cycles. One pain is a bit more acute than the other.

rfc/006-semantic-versioning.md
```diff
@@ -0,0 +1,76 @@
...
+
+## Proposal
+
+Tendermint will follow a same formatting structure of [SemVer](https://semver.org/) with a major, minor and patch version but with slight adjustments to what they encompass:
+
+A major version in Tendermint guarantees that for the entire cycle of that major version:
+
+- All blocks and state data in a blockchain can be queried. All headers can be verified even across minor version changes. Nodes can both block sync and state sync from genesis to the head of the chain.
```
williambanfield, line 44: Not sure I understand what 'can be queried' means here. Queried and parsed by clients released within the same major version?

Additionally, does `genesis` here mean genesis of the entire chain, or genesis of the sub-chain piece that is produced as a result of the most recent upgrade?

rfc/006-semantic-versioning.md
```diff
@@ -0,0 +1,76 @@
...
+- An external client should not need to modify their code to interact with a node's RPC endpoints. New endpoints can be added but existing ones can not be altered or removed.
+- Migrations should be automatic. Upgrading of one node can happen asynchronously with respect to other nodes (although agreement of a network-wide upgrade will happen synchronously via consensus).
+
+A minor version in Tendermint guarantees that for the entire cycle of that minor version:
+
+- All publicly available Go API's, for example in `node` or `abci` packages will not change in a breaking way that would require any consumer (not just application developers) to modify their code.
+- No breaking changes to the block protocol. This means that all block related data structures should not change in a way that breaks any of the hashes, the consensus engine or light client verification.
```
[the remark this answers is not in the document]
williambanfield: `hash-breaking` -> `verification-breaking`?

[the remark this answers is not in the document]
creachadair: > `hash-breaking` ⇒ `verification-breaking`?

That seems fine too—but I'm not sure it is any clearer? Isn't verification essentially an extension of hash equality?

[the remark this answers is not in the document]
williambanfield: For the most part, although if we change the signature scheme, say to aggregated signatures, we haven't really broken the hash verification but we have produced a 'verification' breaking change since the old code can no longer verify the new block structure.

rfc/006-semantic-versioning.md
creachadair: Hmm. So maybe "multi-chain external clients"?

cmwaters: I was talking about something similar to this with @creachadair yesterday. My interpretations is that there are two types of clients to a blockchain: one that queries information from a trusted source and the other that wants to validate information. I'd ideally like to map this distinction to our two IPC layers: P2P for verification and RPC for information hence the reason for wanting to bring the light client into the p2p stack - that way clients that want verification run their light client in the browser and query it for information via RPC. The only tricky situation in all of this is the IBC module which uses the RPC layer and is outside of what we'd consider the network boundary yet requires data for verification purposes. I will write some notes up.

rfc/006-semantic-versioning.md
```diff
@@ -0,0 +1,76 @@
...
+
+In a more broader sense, these four users are inextricably linked and their concerns reflect that of the end user of the applications. The major conclusion that this RFC makes is that **the ability for chains to provide continual service is more important than the actual upgrade burden put on the developers of these chains**. What that means is that although it may be burdensome for application developers to upgrade Tendermint versions, this is far outweighed by the pain node operators experience in upgrading, by the pain that external clients have in accommodating changes and ultimately the pain of end users who are benefiting from the application. Developers have all the time in world to make these changes whilst for other user groups, upgrading can be considered part of this real-time critical path.  
+
+## Proposal
+
+Tendermint will follow a same formatting structure of [SemVer](https://semver.org/) with a major, minor and patch version but with slight adjustments to what they encompass:
+
+A major version in Tendermint guarantees that for the entire cycle of that major version:
```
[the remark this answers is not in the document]
cmwaters: Yeah that makes sense. I had sort of dodged talking about any guarantees between major releases because there isn't really and the way we plan to support prior major versions is something I wanted to address in the soft upgrades ADR

rfc/006-semantic-versioning.md
```diff
@@ -0,0 +1,76 @@
...
+- Nodes in a network are able to communicate and perform BFT state machine replication so long as the agreed network version is the lowest of all nodes in a network. For example, nodes using version 1.5.x and 1.2.x can operate together so long as the network version is 1.2 or lower (but still within the 1.x range). This rule essentially captures the concept of network backwards compatibility.
+- An external client should not need to modify their code to interact with a node's RPC endpoints. New endpoints can be added but existing ones can not be altered or removed.
+- Migrations should be automatic. Upgrading of one node can happen asynchronously with respect to other nodes (although agreement of a network-wide upgrade will happen synchronously via consensus).
+
+A minor version in Tendermint guarantees that for the entire cycle of that minor version:
+
+- All publicly available Go API's, for example in `node` or `abci` packages will not change in a breaking way that would require any consumer (not just application developers) to modify their code.
```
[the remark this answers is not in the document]
cmwaters: > Can we either commit to moving those libraries, or define which ones are explicitly exempted from this promise?

I would like to commit to moving the libraries. I'm not sure what packages are outstanding (https://github.com/tendermint/tendermint/issues/4451)

rfc/006-semantic-versioning.md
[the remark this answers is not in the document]
cmwaters: This is true. I think verification breaking is more accurate although I'll keep the examples of changing the contents in a hash or changes in signature verification

rfc/006-semantic-versioning.md
cmwaters: I wanted the semantic versioning to be something that all Tendermint implementations adopt hence why I decided to have this in the spec repo (although I also thought about writing this in the tendermint go repo).

rfc/006-semantic-versioning.md
cmwaters: Genesis, here, means of the entire chain. Prior you'd see multiple versions of a chain depicted in their `chainID` (i.e. `cosmoshub-1`, `cosmoshub-2`... `cosmoshub-4`). In this case you could argue that the genesis file is actually that of the respective `chainID` but I'm implying that we don't want this anymore and genesis should refer to the very beginning of a network.

cmwaters: > Not sure I understand what 'can be queried' means here. Queried and parsed by clients released within the same major version?

Just that it should be possible to retrieve all information corresponding to a chain regardless of how the versioning of that chain evolves

rfc/006-semantic-versioning.md
cmwaters: @williambanfield, yeah I agree _all the time in the world_ is an exaggeration and I should be a little more precise in my wording. I think your description more accurately conveys what I was trying to say.

rfc/006-semantic-versioning.md
```diff
@@ -0,0 +1,76 @@
...
+
+A major version in Tendermint guarantees that for the entire cycle of that major version:
+
+- All blocks and state data in a blockchain can be queried. All headers can be verified even across minor version changes. Nodes can both block sync and state sync from genesis to the head of the chain.
+- Nodes in a network are able to communicate and perform BFT state machine replication so long as the agreed network version is the lowest of all nodes in a network. For example, nodes using version 1.5.x and 1.2.x can operate together so long as the network version is 1.2 or lower (but still within the 1.x range). This rule essentially captures the concept of network backwards compatibility.
+- An external client should not need to modify their code to interact with a node's RPC endpoints. New endpoints can be added but existing ones can not be altered or removed.
+- Migrations should be automatic. Upgrading of one node can happen asynchronously with respect to other nodes (although agreement of a network-wide upgrade will happen synchronously via consensus).
```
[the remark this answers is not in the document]
cmwaters: I think we want migrations for both coordinated and rolling upgrades to be automatic. I think the only clause for this is if we find the migration to be something optional (an example of which is migrating the peer store which in the end we decided not to do).

rfc/006-semantic-versioning.md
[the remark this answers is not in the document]
creachadair: I agree we should avoid overcommitting about what we'll guarantee across releases. I was thinking more about clarifying the scope of maintenance of older releases, and setting some ground rules about when a user has to upgrade to get the new features vs. getting backports.

rfc/006-semantic-versioning.md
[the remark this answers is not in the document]
creachadair: I see. The first sentence equates "automatic" with "asynchronous", while the second describes "network-wide" upgrades as having to be synchronous, so I thought they were meant to be an exception.

Are there two distinctions we need to make here, e.g., "synchronous vs. asynchronous" and "automatic vs. hands-on" or something like that?

rfc/006-semantic-versioning.md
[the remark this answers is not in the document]
creachadair: I see, so the issue is not just changes to the block _structure_, but also changes to the protocol rules (e.g., your example of the hash algorithm). That makes sense.

creachadair:
If you update your branch, I think the lint errors should go away. 🙂

rfc/006-semantic-versioning.md
williambanfield: I think that this makes sense. We keep definitions for the public APIs in this repo so it makes sense to have a document describing how we version them live alongside them. It means tho, that we should probably come up with a more disciplined process for 'releasing' this repo.

cmwaters: Yeah I totally agree

creachadair:
> If you update your branch, I think the lint errors should go away. 🙂

Well, they _were_ all fixed. But since then more folks have checked in broken links 😞 

We should either make that check required, or remove it.

williambanfield:
> > If you update your branch, I think the lint errors should go away. slightly_smiling_face
> 
> Well, they _were_ all fixed. But since then more folks have checked in broken links disappointed
> 
> We should either make that check required, or remove it.

I vote make required, it's a simple check and should prevent some reasonably embarrassing issues.

cmwaters:
I also agree and have updated it accordingly

rfc/006-semantic-versioning.md
```diff
@@ -0,0 +1,76 @@
...
+
+## Proposal
+
+Tendermint will follow a same formatting structure of [SemVer](https://semver.org/) with a major, minor and patch version but with slight adjustments to what they encompass:
+
+A major version in Tendermint guarantees that for the entire cycle of that major version:
+
+- All blocks and state data in a blockchain can be queried. All headers can be verified even across minor version changes. Nodes can both block sync and state sync from genesis to the head of the chain.
+- Nodes in a network are able to communicate and perform BFT state machine replication so long as the agreed network version is the lowest of all nodes in a network. For example, nodes using version 1.5.x and 1.2.x can operate together so long as the network version is 1.2 or lower (but still within the 1.x range). This rule essentially captures the concept of network backwards compatibility.
```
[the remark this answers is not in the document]
cmwaters: > Is "network version" different from "Tendermint version"? It seems like maybe it's different but related?

Network version is a subset of the Tendermint version that pertains to changes made to the p2p layer that would affect communication over the wire.

What I am trying to get at is that we don't provide forward compatibility of both the wire format and data structures, thus if a network agrees to bump the version, those that haven't upgraded will stop. Does that answer your question?

rfc/006-semantic-versioning.md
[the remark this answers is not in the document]
cmwaters: Right. I was associating automatic with not requiring manual intervention from the node operators although I'm also not sure if this is a gaurantee that we can really make.

rfc/006-semantic-versioning.md
```diff
@@ -0,0 +1,84 @@
...
+
+Tendermint has two primary mediums with which it can communicate to other processes: RPC and P2P. The division marks the boundary between the internal and external components of the network. The P2P layer is used in all cases that nodes (validator, full, seed or light) need to communicate with one another whereas the RPC is for any outside process that wants to communicate with a node. The assumption made here is that all communication via the RPC is to a trusted source and thus the ability to interpret information is more important than being able to verify the information. P2P is therefore primary medium for verification. As an example, an in-browser light client would consist of verifying headers (and perhaps application state) via the p2p layer and passing the information on to the client via the RPC layer (or potentially directly via an API).
+
+The exception to this is the IBC module and relayers which are external but require verifiable data. Breaking changes to the light client verification path mean that all neighbouring chains that are connected will no longer be able to verify state transitions and thus pass messages back and forward.
```
cmwaters, line 42: I'm still uncertain as to what guarantees we should give this user group? Should we not break light client verification across the entire major version? 

The most problematic aspect about this is that individual nodes are easier to upgrade than entire networks. It may take 6 months for a network to update their IBC module and include it in the next release. I've been pondering on this for a while and I can't think of any tricks to get around this. ABCI++ will add vote extensions which will change the bytes that get signed over such that previous clients won't be able to verify new signatures. A couple ideas I've had to mitigate the damage. 
- Allow wasm-like clients that can be upgraded separately to the entire chain
- If there are signature changes, have validators sign over the current and previous data types ? Or have ways we can parse the old data formats from the new.

Other than that the best I can think of is to release the light client stuff before everything else so we can give the IBC team and their users as much time as possible to upgrade

 Thoughts? @creachadair @colin-axner

colin-axner: > Should we not break light client verification across the entire major version?

Yes I think so (at least from the perspective of IBC and at least in the short term). Breaking the light client verification means that all counterparties must upgrade their code in order to prevent their IBC connection from breaking. 

As you say, using Wasm to upgrade clients as opposed to chain code upgrades can ease this burden, but I don't think we can assume every chain using IBC is running Wasm. 

> If there are signature changes, have validators sign over the current and previous data types ? Or have ways we can parse the old data formats from the new.

I think exploration of this idea is useful, but this would be a non-breaking solution (since it allows for backwards compatibility)

---

Does there exist light client breaking changes which would not normally cause a major version bump?

In general I'm still fairly concerned about upgrading IBC clients. I think the practical timeline of expecting counterparties to upgrade to support the new IBC client is minimum 1 year. Furthermore, even if most chains use Wasm, the current clients that have been initialized do not currently have a way to migrate to become Wasm clients

I don't see a way around this though since breaking the light client algorithm means new code needs to be added to handle the breakage. The only things that come to mind are trying to minimize the surface area of changes which require this upgrade path, either with backwards compatible fixes, wasm clients or hub routing (hub routing would only be beneficial if the hub is running a tm chain that will never break the light client algorithm or it has a wasm client and most connected chains have a wasm client).

cmwaters: It's an interesting game-theory-like ploy to use breaking IBC as a forcing function to get networks to upgrade Tendermint versions 😄.

Anyway, I'am also concerned about this. I think even if it was to be a month of incompatibility it would still be disastrous for the ecosystem. Osmosis' entire value proposition rests on a functioning IBC. 

> Does there exist light client breaking changes which would not normally cause a major version bump?

Prior to this discussion most light client breaking changes would not result in a major version bump. A major version bump really means you need to start a new chain; your existing chain is incompatible with the new software. We want to reduce that as much as possible so all nodes (including light clients) will be able to verify data for all minor versions that are less than equal to their minor version. This means that all nodes need to upgrade ahead of when the entire network itself decides to change versions (In fact, validators won't vote to upgrade the network version if they don't have that software so you'll need 2/3+). I think this works fine within the boundaries of a single network which releases the same software for everyone but falls apart across multiple networks where each ecosystem has their own release schedules. 

Having wasm clients or some other method for dynamically upgrading clients would be a great thing to have in the future. We don't want to have too many light client breaking changes but they will probably still happen so putting something in place that can minimize the downtime of IBC when it happens is definitely worth investing in.

creachadair: I don't have a strong opinion here, but I guess for me the question that matters is this: Is there some reason we should _not_ guarantee against light-client breaking changes within a major release? 

As with the other user groups, I'm receptive to an escape hatch for security issues. Is there some other obvious technical reason why we might want to break light client verification specifically, without breaking other users of the protocol?

If we needed to do that anyway, then an argument to the incentive might be a reasonable justification for it—but if we don't have a clear need my intuition is we should include these users in the version guarantees.

rfc/006-semantic-versioning.md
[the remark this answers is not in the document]
creachadair: This seems like a choice we can make. Some options that occur to me (there may be more):

- Option 1: "We promise that any changes within a major version do not require manual upgrade of nodes in standard configuration."[1] Meaning: We will defer changes that _do_ require manual intervention to the next major version.
- Option 2: "We promise that there will be at most 1 change within a major version that requires manual upgrade of nodes in standard configuration."[1] Meaning: We will allow a budget of non-breaking manual upgrades within a major version.
- Option 3: "We will try really really hard to ensure that all changes within a major version can be upgraded without manual intervention for the majority of nodes." Meaning: Best-effort and we will try to provide tools, but we won't stop the world if we can't get everybody.

Given that we do not have a reliable customer list, (3) may be the only thing we can practically commit to. But from the user's perspective it isn't really a commitment they can rely on, since it allows us to throw up our hands at any time. To the extent that version promises are a means of building trust, we should choose cautiously.

[1] For some definition of "standard". Basically leaving an escape-hatch for weird and/or highly-customized nodes.

rfc/006-semantic-versioning.md
cmwaters: > Is there some other obvious technical reason why we might want to break light client verification specifically

This will quite often be a by-product of adding some other feature. Like PBTS could break the light client although we've chosen not to for 0.36. More direct forms of breaking the verification algorithm would be to change the way we do signatures, change what's in those signatures or change what's in the header.

creachadair: Understood. But wouldn’t those example changes also break other users as well?

cmwaters: Yes but for node operators and external clients we find such breakage acceptable. Node operators need to upgrade their binary but this can be done automatically and networks should only upgrade when all/most of the validators have the next version running. For external clients we care more for readability than verifiability

rfc/006-semantic-versioning.md
```diff
@@ -0,0 +1,84 @@
...
+
+### Modes of Interprocess Communication
+
+Tendermint has two primary mediums with which it can communicate to other processes: RPC and P2P. The division marks the boundary between the internal and external components of the network. The P2P layer is used in all cases that nodes (validator, full, seed or light) need to communicate with one another whereas the RPC is for any outside process that wants to communicate with a node. The assumption made here is that all communication via the RPC is to a trusted source and thus the ability to interpret information is more important than being able to verify the information. P2P is therefore primary medium for verification. As an example, an in-browser light client would consist of verifying headers (and perhaps application state) via the p2p layer and passing the information on to the client via the RPC layer (or potentially directly via an API).
```
creachadair, line 40:
```suggestion
Tendermint has two primary mechanisms to communicate with other processes: RPC and P2P. The division marks the boundary between the internal and external components of the network:

- The P2P layer is used in all cases that nodes (of any type) need to communicate with one another.
- The RPC interface is for any outside process that wants to communicate with a node.

The design principle here is that **communication via RPC is to a trusted source** and thus the RPC service prioritizes inspection rather than verification. The P2P interface is the primary medium for verification.

As an example, an in-browser light client would verify headers (and perhaps application state) via the p2p layer, and then pass along information on to the client via RPC (or potentially directly via a separate API).
```

rfc/006-semantic-versioning.md
creachadair:
```suggestion
The main exceptions to this are the IBC module and relayers, which are external to the node but also require verifiable data. Breaking changes to the light client verification path mean that all neighbouring chains that are connected will no longer be able to verify state transitions and thus pass messages back and forward.
```

creachadair: If we agree that the IBC use case is high priority, perhaps we should start with a conservative guarantee for those users specifically.

rfc/006-semantic-versioning.md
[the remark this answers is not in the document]
creachadair: I think it does answer my question. I'm trying to think about how to word this promise in a way that would avert my original confusion. Perhaps:

> All nodes in a network operating at or above the agreed network version are able to communicate and perform BFT state replication.

Trying to get at the idea that the whole network doesn't stop just because one node is recalcitrant -- only the laggards get stuck. Maybe there's a better way to put that.

creachadair (approved):
Thanks, I really like the improvements you've made. I have some suggestions, most or all of which should be seen as optional. I also added a few questions, but nothing that I would block over.

rfc/006-semantic-versioning.md
```diff
@@ -0,0 +1,84 @@
...
+- Nodes in a network are able to communicate and perform BFT state machine replication so long as the agreed network version is the lowest of all nodes in a network. For example, nodes using version 1.5.x and 1.2.x can operate together so long as the network version is 1.2 or lower (but still within the 1.x range). This rule essentially captures the concept of network backwards compatibility.
+- Node RPC endpoints will remain compatible with existing external clients:
+    - New endpoints may be added, but old endpoints may not be removed.
+    - Old endpoints may be extended to add new request and response fields, but requests not using those fields must function as before the change.
+- Migrations should be automatic. Upgrading of one node can happen asynchronously with respect to other nodes (although agreement of a network-wide upgrade must still occur synchronously via consensus).
+
+For the entire cycle of a **minor version** in Tendermint:
+
+- Public Go API's, for example in `node` or `abci` packages will not change in a way that requires any consumer (not just application developers) to modify their code.
```
williambanfield, line 59: The mild difficulty that I see this presenting is that Go's tooling will consider two versions 'compatible' within the same major version. Using the tooling to automatically grab the latest will now grab a version of the software that breaks compile. I don't have a huge objection to this, but it is a bit out of line with the way most go projects work.

rfc/006-semantic-versioning.md
```diff
@@ -0,0 +1,84 @@
...
+
+The exception to this is the IBC module and relayers which are external but require verifiable data. Breaking changes to the light client verification path mean that all neighbouring chains that are connected will no longer be able to verify state transitions and thus pass messages back and forward.
+
+## Proposal
+
+Tendermint version labels will follow the syntax of [Semantic Versions 2.0.0](https://semver.org/) with a major, minor and patch version. The version components will be interpreted according to these rules:
```
williambanfield, line 46: Should we clarify that patch releases are for bug fixes and security patches only and that new features will always accompany bumps to the minor version?

williambanfield (approved):

rfc/006-semantic-versioning.md
creachadair: Agreed, but note below:
> These guarantees will come into effect at release 1.0.

so I think we are safe. Possibly we should highlight that line at the top of the doc, though.

Rose2161 (approved):

rfc/006-semantic-versioning.md
```diff
@@ -0,0 +1,84 @@
...
+
+It's worth to further acknowledge that the strictness of our versioning guarantees may mean nothing if we don't also address the frequency that the team intends on making version changes. We could make the strictest guarantees in the world yet simply break them with every release.
+
+Finally I would like to remark that this RFC only addresses the what, as in what are the rules for versioning. The how, or how does Tendermint plan to follow the versioning logic, will be addressed in a later RFC on Soft Upgrades.
+
+## Discussion
+
+We first begin with a round up of the various users and a set of assumptions on what these users expect from Tendermint in regards to versioning:
+
+1. **Application Developers**, those that use the ABCI to build applications on top of Tendermint, are chiefly concerned with that API. Breaking changes will force developers to modify large portions of their codebase to accommodate for the changes. Some ABCI changes such as introducing priority for the mempool don't require any effort and can be lazily adopted whilst changes like ABCI++ may force applications to redesign their entire execution system.
```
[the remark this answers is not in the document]
cmwaters: Where I was coming from when I wrote this statement is developers that want to upgrade to a new version. When you upgrade to v0.36, you'll get ABCI++ as part of it. You can not opt out of it and are "forced" to make changes to your application

rfc/006-semantic-versioning.md
colin-axner: This [ADR](https://github.com/cosmos/ibc-go/blob/main/docs/architecture/adr-027-ibc-wasm.md) might help. Wasm allows light clients to be upgraded by submitting a message to a chain. Existing light clients require the entire chain to upgrade to upgrade the light client

rfc/006-semantic-versioning.md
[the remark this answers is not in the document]
cmwaters: I think your distinction here in your earlier comment captures something I hadn't thought too much of. Which is that the surface application developers have when writing in Go is quite different to when writing in other languages and that we shouldn't look at just the ABCI interface but other things like the RPC which for example the CosmosSDK uses.

rfc/006-semantic-versioning.md
cmwaters: Yeah that might make sense to add a short note about

rfc/006-semantic-versioning.md
```diff
@@ -0,0 +1,84 @@
...
+    - New endpoints may be added, but old endpoints may not be removed.
+    - Old endpoints may be extended to add new request and response fields, but requests not using those fields must function as before the change.
+- Migrations should be automatic. Upgrading of one node can happen asynchronously with respect to other nodes (although agreement of a network-wide upgrade must still occur synchronously via consensus).
+
+For the entire cycle of a **minor version** in Tendermint:
+
+- Public Go API's, for example in `node` or `abci` packages will not change in a way that requires any consumer (not just application developers) to modify their code.
+- No breaking changes to the block protocol. This means that all block related data structures should not change in a way that breaks any of the hashes, the consensus engine or light client verification.
+- Upgrades between minor versions may not result in any downtime (i.e., no migrations are required), nor require any changes to the config files to continue with the existing behavior. A minor version upgrade will require only stopping the existing process, swapping the binary, and starting the new process.
+
```
williambanfield, line 62:
```suggestion
A new **patch version** of Tendermint will only contain bug fixes and updates that impact the security and stability of Tendermint. 
```

rfc/006-semantic-versioning.md
```diff
@@ -0,0 +1,93 @@
...
+For a versioning policy to make sense, we must also address the intended frequency of breaking changes. The strictest guarantees in the world will not help users if we plan to break them with every release.
+
+Finally I would like to remark that this RFC only addresses the "what", as in what are the rules for versioning. The "how" of Tendermint implementing the versioning rules we choose, will be addressed in a later RFC on Soft Upgrades.
+
+## Discussion
+
+We first begin with a round up of the various users and a set of assumptions on what these users expect from Tendermint in regards to versioning:
+
+1. **Application Developers**, those that use the ABCI to build applications on top of Tendermint, are chiefly concerned with that API. Breaking changes will force developers to modify large portions of their codebase to accommodate for the changes. Some ABCI changes such as introducing priority for the mempool don't require any effort and can be lazily adopted whilst changes like ABCI++ may force applications to redesign their entire execution system.
```
cmwaters, line 25:
```suggestion
1. **Application Developers**, those that use the ABCI to build applications on top of Tendermint, are chiefly concerned with that API. Breaking changes will force developers to modify large portions of their codebase to accommodate for the changes. Some ABCI changes such as introducing priority for the mempool don't require any effort and can be lazily adopted whilst changes like ABCI++ may force applications to redesign their entire execution system. It's also worth considering that the API's for go developers differ to developers of other languages. The former here can use the entire Tendermint library, most notably the local RPC methods, and so the team must be wary of creating changes there as well. 
```

PR #374, opened 2021-11-30
rpc: add header and header_by_hash methods

cmwaters:
Adds two new methods: `header` and `header_by_hash`

Reference: https://github.com/tendermint/tendermint/pull/7270


creachadair (approved):
This makes me realize we will also need to add these to our OpenAPI definition, as well.

PR #375, opened 2021-12-03
PBTS: new system model and problem statement

cason:
It seems that some of the changes listed here were already addressed by the previous PR.

This PR essentially adds `pbts-sysmodel_002_draft.md`, a new version of `pbts-sysmodel_001_draft.md`.

Not all the properties and references to the TLA spec are included, and this PR will probably require more commits and some discussions.

### DIFF
--- a/spec/consensus/proposer-based-timestamp/pbts-algorithm_002_draft.md
+++ b/spec/consensus/proposer-based-timestamp/pbts-algorithm_002_draft.md
@@ -12,16 +12,16 @@ time values that differ more than `PRECISION`, which is a system parameter.
 
 ### Proposal times are definitive
 
-When a value `v` is produced by a process, it also assigns the associated time `v.time`.
+When a value `v` is produced by a process, it also assigns the associated proposal time `v.time`.
 If the same value `v` is then re-proposed in a subsequent round of consensus,
 it retains its original time, assigned by its original proposer.
 
-A value `v` should re-proposed when it becomes valid, i.e., when it receives `2f + 1 PREVOTES` in a round `r` of consensus.
+A value `v` should re-proposed when it becomes locked by the network, i.e., when it receives `2f + 1 PREVOTES` in a round `r` of consensus.
 This means that processes with `2f + 1`-equivalent voting power accepted, in round `r`, both `v` and its associated time `v.time`.
 Since the originally proposed value and its associated time were considered valid, there is no reason for reassigning `v.time`.
 
 In the [first version][v1] of this specification, proposals were defined as pairs `(v, time)`.
-In addition, the same value could be proposed, in different rounds, associated to distinct times.
+In addition, the same value `v` could be proposed, in different rounds, but would be associated to distinct times each time it was reproposed.
 Since this possibility does not exist in this second specification, the proposal time became part of the proposed value.
 With this simplification, several small changes to the [arXiv][arXiv] algorithm are no longer required.
 
@@ -47,7 +47,7 @@ A value rejected by the `valid()` implementation is not accepted by any correct 
 ## Timely Proposals
 
 PBTS introduces a new requirement for a process to accept a proposal: the proposal must be `timely`.
-It is a temporal requirement, associated to a couple of synchrony (that is, timing)
+It is a temporal requirement, associated with the following synchrony (that is, timing)
 [assumptions][model] regarding the behavior of processes and the network:
 
 - Synchronized clocks: the values simultaneously read from clocks of any two correct processes differ by at most `PRECISION`;
--- a/spec/consensus/proposer-based-timestamp/pbts-algorithm_002_draft.md
+++ b/spec/consensus/proposer-based-timestamp/pbts-algorithm_002_draft.md
@@ -47,15 +47,10 @@ A value rejected by the `valid()` implementation is not accepted by any correct 
 ## Timely Proposals
 
 PBTS introduces a new requirement for a process to accept a proposal: the proposal must be `timely`.
-It is a temporal requirement, associated to a couple of synchrony (that is, timing) assumptions regarding the behavior of processes and the network.
+It is a temporal requirement, associated to a couple of synchrony (that is, timing)
+[assumptions][model] regarding the behavior of processes and the network:
 
-The evaluation of the `timely` requirement requires comparing the proposal's sending time with the proposal's receiving time.
-As these two time values can be read from different clocks, at different processes, we need to assume that processes' clocks are synchronized.
-As these two times refer to two distinct events, we need to assume a minimum and a maximum real time interval between the occurrence of the two events.
-
-The two synchronous assumptions adopted to evaluate the `timely` predicate are:
-
-- Synchronized clocks: the values read from clocks of any two correct processes at the same instant of real time differ by at most `PRECISION`;
+- Synchronized clocks: the values simultaneously read from clocks of any two correct processes differ by at most `PRECISION`;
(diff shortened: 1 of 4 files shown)


spec/consensus/proposer-based-timestamp/pbts-sysmodel_002_draft.md
```diff
@@ -0,0 +1,222 @@
...
+This parameter was removed in this version for two reasons:
+
+- Clock accuracy is hard, if possible at all, to assess in distributed systems
+- The adoption of an `ACCURACY` parameter renders the `PRECISION` parameter redundant:
+if the accuracy of processes clocks is bound by `ACCURACY`,
+then the clocks precision must be bound by `2 * ACCURACY`.
+
+The adoption of an `ACCURACY` was intended to formalize the relation between a block time and the real time.
+We observe, however, that clients will compare the block time to their local time,
+or to the time they retrieve from a trusted source of time, not to a physical clock.
```
williambanfield, line 52: > not to a physical clock

I'm not totally sure I understand what this phrase means since their local time is generated from a physical clock. I think this sentence is attempting to describe the fact that processes local clocks are not perfectly accurate, is that roughly correct?

spec/consensus/proposer-based-timestamp/pbts-sysmodel_002_draft.md
```diff
@@ -0,0 +1,222 @@
...
+
+- If `p` and `q` are equipped with synchronized clocks, then `|C_p(t) - C_q(t)| < PRECISION`
+
+`PRECISION` thus bounds the difference on the times simultaneously read by processes
+from their local clocks, so that their clocks can be considered synchronized.
+
+#### Accuracy
+
+The [previous system model][v1] included a second clock-related parameter, `ACCURACY`,
+that relates the values read by processes from their synchronized clocks with the real time.
```
williambanfield, line 41: I think that this implies a relationship between the synchronized clock (ntp in this case) and the 'real' time. Should we make that somewhat explicit in this document?

spec/consensus/proposer-based-timestamp/pbts-algorithm_002_draft.md
```diff
@@ -0,0 +1,145 @@
...
+The proposal time is read from the clock of the process that proposes a value for the first time, its original proposer.
+
+With PBTS, therefore, we assume that processes have access to **synchronized clocks**.
+The proper definition of what it means can be found in the [system model][model],
+but essentially we assume that two correct processes do not simultaneous read from their clocks
+time values that differ more than `PRECISION`, which is a system parameter.
+
+### Proposal times are definitive
+
+When a value `v` is produced by a process, it also assigns the associated time `v.time`.
```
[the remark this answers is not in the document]
cason:
```suggestion
When a value `v` is produced by a process, it also assigns the associated proposal time `v.time`.
```

spec/consensus/proposer-based-timestamp/pbts-sysmodel_002_draft.md
cason: Accuracy is between clock readings and real time. We try to improve accuracy of the local clock by synchronizing it with a more accurate clock. By synchronizing here we mean trying to reduce (so, to improve) at most the precision between the local clock and the reference clock. As a result, our local clock accuracy will be the accuracy of the reference clock plus (so, reduced by) the precision achieved in the synchronization process.

spec/consensus/proposer-based-timestamp/pbts-sysmodel_002_draft.md
cason: I don't like this description either. I will try to fix it.

spec/consensus/proposer-based-timestamp/pbts-sysmodel_002_draft.md
```diff
@@ -0,0 +1,222 @@
...
+- Clock accuracy is hard, if possible at all, to assess in distributed systems
+- The adoption of an `ACCURACY` parameter renders the `PRECISION` parameter redundant:
+if the accuracy of processes clocks is bound by `ACCURACY`,
+then the clocks precision must be bound by `2 * ACCURACY`.
+
+The adoption of an `ACCURACY` was intended to formalize the relation between a block time and the real time.
+We observe, however, that clients will compare the block time to their local time,
+or to the time they retrieve from a trusted source of time, not to a physical clock.
+The restriction imposed by the `PRECISION` parameter can then be used, indirectly,
+to bound the difference block times and real time.
```
josef-widder, line 54: It would be good to be more explicit about "can then be used, indirectly".

spec/consensus/proposer-based-timestamp/pbts-sysmodel_002_draft.md
```diff
@@ -0,0 +1,222 @@
...
+In this section we define the properties of Tendermint consensus
+(cf. the [arXiv paper][arXiv]) in this new system model.
+
+#### **[PBTS-PROPOSE.0]**
+
+A proposer proposes a consensus value `v` with an associated proposal time `v.time`.
+
+#### **[PBTS-INV-AGREEMENT.0]**
+
+[Agreement] No two correct processes decide on different values `v` or proposal times `v.time`.
```
josef-widder, line 95:
```suggestion
[Agreement] No two correct processes decide on different values `v`. (This implies that no two correct processes decide on different proposal times `v.time`.)
```

spec/consensus/proposer-based-timestamp/pbts-sysmodel_002_draft.md
```diff
@@ -0,0 +1,222 @@
...
+
+The `timely` predicate is evaluated when a process receives a proposal.
+Let `now_p` be time a process `p` reads from its local clock when `p` receives a proposal.
+Let `v` be the proposed value and `v.time` the proposal time.
+The proposal is considered `timely` by `p` if:
+
+1. `now_p >= v.time - PRECISION` and
+1. `now_p <= v.time + MSGDELAY + PRECISION`
+
+### Timely Proof-of-Locks
```
josef-widder, line 119: If I understand the structure of the document correctly, the following text is about analysis / correctness arguments. If this is right, perhaps we should add a header making this more explicit.

josef-widder (approved):
Looks great. Thanks! There are a couple a small issues raised by me and @williambanfield. Once they are fixed we should move forward and merge.

spec/consensus/proposer-based-timestamp/pbts-sysmodel_002_draft.md
cason: A TLA+-related header?

spec/consensus/proposer-based-timestamp/pbts-sysmodel_002_draft.md
cason: Please take a look on the new version.

spec/consensus/proposer-based-timestamp/pbts-sysmodel_002_draft.md
josef-widder: Not sure that it is TLA+ related. I guess the point is just that the following text is more interesting when one wants to understand correctness (rather than protocol details). Perhaps we can discuss this synchronously.

cason: I've added some headers to the latest version.

josef-widder:
We merge the current state. @cason will add a README that explains the status of the files that is
- open issues
- TODOs
- etc.

cason:
I think this branch is still not mergeable with master, needing a rebase. Can you confirm that? I don't have the option to merge the branch, @williambanfield can you help me (again) with this :D

PR #384, opened 2022-01-10
proto: rename timing params to synchrony params

williambanfield:
This pull request renames the PBTS parameters from `TimingParams` to `SynchronyParams`. This name more specifically reflects what the parameters are for. Additionally, another set of consensus parameters are planned to be added called `TimeoutParams` as part of the work for [ADR-74](https://github.com/tendermint/tendermint/pull/7503). The similarity of these names will definitely cause confusion so changing this name seems like a good fix.

### DIFF
--- a/proto/tendermint/types/params.proto
+++ b/proto/tendermint/types/params.proto
@@ -13,7 +13,7 @@ message ConsensusParams {
   EvidenceParams  evidence  = 2;
   ValidatorParams validator = 3;
   VersionParams   version   = 4;
-  SynchronyParams timing    = 5;
+  SynchronyParams synchrony = 5;
 }
 
 // BlockParams contains limits on the block size.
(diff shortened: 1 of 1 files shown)


proto/tendermint/types/params.proto
```diff
@@ -13,7 +13,7 @@ message ConsensusParams {
   EvidenceParams  evidence  = 2;
   ValidatorParams validator = 3;
   VersionParams   version   = 4;
-  TimingParams    timing    = 5;
+  SynchronyParams timing    = 5;
```
creachadair, line 16: Should we also update the field name as well?

Both are breaking changes to generated code, but neither is wire-breaking.

williambanfield: Ah, thanks for the catch.
The break is in unreleased code, so the change shouldn't truly be breaking at all 🤷

creachadair (approved):

proto/tendermint/types/params.proto
creachadair: Yeah, but it will become a migration guide item later. 🙂

cmwaters (approved):

cason (approved):

PR #385, opened 2022-01-13
Fixes linter 

SweeXordious:
Bumps linter version to v4 and makes checkout fetch deep.

### DIFF
--- a/.github/workflows/action.yml
+++ b/.github/workflows/action.yml
@@ -8,3 +8,5 @@ jobs:
     steps:
     - uses: actions/checkout@master
     - uses: gaurav-nelson/github-action-markdown-link-check@1.0.13
+      with:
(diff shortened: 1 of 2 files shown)


creachadair:
I believe markdown-link-check is "stuck" because no markdown files are touched in this PR. I need to figure out how to convince it to "succeed" quietly in that case, but until then if you add an empty .md file to the PR that should trigger the CI.

SweeXordious:
@creachadair I just added the `check-modified-files-only: 'yes'` option as it might fix the issue.

creachadair:
> @creachadair I just added the `check-modified-files-only: 'yes'` option as it might fix the issue.

I think I figured out the problem. I incorporated your changes along with my fix in #387, since it was a small change anyway. Thanks for the PR!

PR #391, opened 2022-01-25
Add a releases document

cmwaters:
This PR adds documentation around the releasing process of tendermint specifications.


RELEASES.md
```diff
@@ -0,0 +1,42 @@
+# Releases
+
+The Tendermint specification follows [semantic versioning](https://semver.org/). Implementations of the Tendermint protocol are encouraged to either mirror versions or have some form of mapping for clear readibility of the state of the implementation. The `master` branch is used for active protocol development. Developers should be wary that the specifications are subject to change and should instead rely on the latest release. In the event of a security vulnerability, fixes may be backported to prior major releases but generally only the last major release is supported. The following documentation is provided to describe the release process.
```
creachadair, line 3:
```suggestion
The Tendermint specification follows [semantic versioning](https://semver.org/). Implementations of the Tendermint protocol are encouraged to either mirror versions or have some form of mapping for clear readibility of the state of the implementation. The `master` branch is used for active protocol development. Developers should be aware that the specifications are subject to change and should refer to tagged releases rather than head. In the event of a security vulnerability, fixes may be backported to prior major releases, but generally only the last major release is supported. The following documentation is provided to describe the release process.
```

RELEASES.md
```diff
@@ -0,0 +1,42 @@
...
+
+Releases should capture a complete snapshot of the specification at a point in time. The protocols described must be coherent and capable of functioning together. There should be no drafts or outstanding/unanswered comments. Some specifications contain references to a versioning schema that is used for formal verification. There should be no ambiguity around what version is being used. Uncompleted or proposed but not yet finalized work can remain in master but it is preferable that large protocol designs use a separate long living branch.
```
creachadair, line 5: Suggestions, mainly to be more prescriptive ("must" rather than "should"):

```suggestion
Releases capture a complete snapshot of the specification at a point in time. The protocols described in a release must be coherent and capable of functioning together. A release must include no drafts or outstanding/unanswered comments. For specifications that contain references to versions used for formal verification, there must be no ambiguity around what version is being used. Uncompleted or proposed but not yet finalized work can remain in master but it is preferable that large protocol designs use a separate long-lived development branch.
```

RELEASES.md
```diff
@@ -0,0 +1,42 @@
...
+
+## Major Releases
+
+Before any major release, it is preferable to create release candidates (RC) for Tendermint implementations as well as applications to test out and validate. This is done using git tags which are built off major release branches i.e. `v0.35.x`.
```
creachadair, line 9: Do we really want to completely mirror the release-candidate strategy for the spec? It seems like a lot of bookkeeping, and introduces another potential synchronization problem between the spec versions and the implementation versions.

I could see it making sense if we require the spec and implementation to _match_ on these versions, e.g., "we tag -rc1 to coordinate with -rc1 of the core" or words to that effect. But it seems like we should have some rule that doesn't permit the spec to go off on a release tangent.

cmwaters: I think the spec should be released ahead of the implementation (to prove that it actually guides the implementation), yet there might be things that are uncovered that require us to make changes to the spec later on. It's kind of like a two-phase commit so I feel like there also needs to be release-candidates for the spec as well

creachadair: All that makes sense, but I feel like we should have a rule to the effect of "you mayn't release a new spec version without that you also release a corresponding implementation version" and possibly vice versa. Otherwise if we need to change the spec in some way that doesn't affect the implementation, they'll get out of sync.

This would imply we could have "artificial" releases (e.g., "TM 3.2.9 is the same as 3.2.8 but for the spec bump"). I don't think that's bad, necessarily, but we have to remember to do it.

cmwaters: Can you think of a situation where we might be forced to be out of sync? Maybe it's difficult to manage the minor versions because the implementation might want to release something with a bunch of non-breaking changes that the spec doesn't really care about but I'd imagine that at least for all major releases it's not a hard ask to guarantee that they align

creachadair: Re: Situation. Suppose we discover a typo in the spec. Where it says "the node MUST delete the fnrfblxty" it should have read "the node MUST NOT delete the fnrfblxty". We edit the spec to fix it, but now the released version of the spec needs updated. It turns out the implementation _doesn't_ delete frnfblxties, so we don't have to change it, but we need a point release on the spec.

creachadair: Taking a step back, I think the point I'm trying to make is that the value of version numbers on the spec is to help the reader understand which versions of the spec and the implementation are compatible.

If we're saying "v29 of the spec and all v29.x.y versions of the implementation must match", then we don't really need three degrees of variation for the spec version. We could, for example, make a rule like "the tip of the v29 spec branch MUST always be mutually compatible with all v29.x.y versions of TM core" or words to that effect. 

If we change the spec in a meaningful way—one requiring TM to change—we should likely do a release of both.

cmwaters: I agree that the three degrees of variation is a bit overkill. Yes the goal is so that someone goes "I'm using tm v0.29.0, what is the version of the spec I should be looking at" and can easily find it. This is easiest if they are like: "I'm using v0.29.0 so I'm going to look at the spec of the same version number". 

I like your suggestion about having the tip of the v29 spec needing to be mutually compatible with all v29.x.y versions of TM Core. So that way a person goes "I'm running v0.29.3 and the latest v0.29.x version in the spec is v0.29.1 so that's where I need to go to look at the spec"

creachadair: Agreed—I'd be fine with making sure they are always in sync, regardless how many decimal points we use. (Of course, there'd be brief periods where "spec > implementation", e.g,. when we've released a change in the spec but not done the implementation bump yet, but that's fine as long as we do them "close in time").

cmwaters: On minor versions there would actually be a tendency for the implementation to get ahead of the spec. Take for instance bumping to v0.35.1. There are no changes that need to be made to the spec so it would remain at v0.35.0. For major versions we would release the spec first but as you say the implementation should closely follow

creachadair: Yeah—that's why I was suggesting we not have all three dots on the spec (whether we do, say v0.35 vs. v0.35.0)

RELEASES.md
```diff
@@ -0,0 +1,42 @@
...
+
+Tags for RCs should follow the "standard" release naming conventions, with `-rcX` at the end (for example, `v0.35.0-rc0`) and incrementing from `0`.
+
+If this is the first RC, then you'll need to create a new branch for the major release:
+
+1. Start on `master`
+2. Create and push the new branch: `git checkout -b v0.35.x && git push origin v0.35.x`
+
+Now, to prepare the new RC:
```
williambanfield, line 18: I'm guessing this was copy+pasted from Tendermint? I'm not really sure we need to actually do release candidates of the spec. RCs AFAICT exist to get the software into users hands and indicate a certain level of stability so that would-be early adopters can begin integrating. I don't see an analog for the spec so I'm not sure we need an analogous mechanism.

RELEASES.md
williambanfield: @creachadair Are you suggesting that all major versions of "Tendermint the code base" have only 1 version of "Tendermint the specification"? Like, "v1.4.7" has "spec v1"? Or would _minor_ releases still have varied versions of the spec? 

To 2 vs 3 dots: I personally don't see a massive need for the patch version since no one imports the spec, they just reference it. We don't have a tremendous need to signal that we've made minor 'bug fixes' to the spec. We will occasionally need to fix typos and minor mistakes etc, but as long as they are mostly cosmetic or just mistakes, they don't have much bearing on what "Tendermint the code base" should do and in general I can't imagine users largely caring.

creachadair: > @creachadair Are you suggesting that all major versions of "Tendermint the code base" have only 1 version of "Tendermint the specification"? Like, "v1.4.7" has "spec v1"? Or would _minor_ releases still have varied versions of the spec?

That's one option, yes. Or maybe major+minor shared. The point is I don't think we benefit much from having the spec have separate patch releases, release candidates, betas, etc. And ideally the spec should not be allowed to get more than 1 minor release ahead of TM core, since that would imply we made spec changes we didn't implement.

cmwaters:
Putting this PR on hold until we have clarity around the future of the spec repo.

Issue #386, opened 2022-01-14
abci: update proto's to reflect changes in ABCI++ 


cmwaters:
## Summary

The changes to the application interface described in https://github.com/tendermint/spec/pull/366 needs to be ported across to the proto definitions.

NOTE: we should first tag a snapshot of the protobuf definitions of v0.35 and create a long lived branch in the event that a minor release updates them.

____

#### For Admin Use

- [ ] Not duplicate issue
- [ ] Appropriate labels applied
- [ ] Appropriate contributors tagged
- [ ] Contributor assigned/self-assigned

sergio-mena:
Addressed via #392

PR #392, opened 2022-01-28
Adapt abci protobuf files to ABCI++ spec (and vice versa)

sergio-mena:
After this PR the "methods" section of ABCI++ are in sync with the types in proto/tendermint/abci/

This is addressing issue #386 

I am seeking to merge to sergio/abci++ (newly created branch) as merging with master would break anyone running "make proto-gen" from Tendermint Go repo.

In order to test the changes, I've adapted  "scripts/protocgen.sh" so that "make proto-gen" picks up my PR branch.

### DIFF
--- a/proto/tendermint/abci/types.proto
+++ b/proto/tendermint/abci/types.proto
@@ -37,10 +37,10 @@ message Request {
     RequestApplySnapshotChunk  apply_snapshot_chunk  = 14;
     RequestPrepareProposal     prepare_proposal      = 15;
     RequestProcessProposal     process_proposal      = 16;
-    RequestExtendVote          extend_vote           = 17;
-    RequestVerifyVoteExtension verify_vote_extension = 18;
     RequestFinalizeBlock       finalize_block        = 19;
   }
+  reserved 17; // Placeholder for RequestExtendVote in v0.37
+  reserved 18; // Placeholder for RequestVerifyVoteExtension in v0.37
 }
 
 message RequestEcho {
@@ -142,21 +142,6 @@ message RequestProcessProposal {
   repeated Evidence       byzantine_validators = 5 [(gogoproto.nullable) = false];
 }
 
-// Extends a vote with application-side injection
-message RequestExtendVote {
-  bytes hash   = 1;
-  int64 height = 2;
-}
-
-// Verify the vote extension
-message RequestVerifyVoteExtension {
-  bytes app_signed        = 1;
-  bytes tendermint_signed = 2;
-  bytes hash              = 3;
-  bytes validator_address = 4;
-  int64 height            = 5;
-}
-
 message RequestFinalizeBlock {
   bytes                   hash                 = 1;
   tendermint.types.Header header               = 2 [(gogoproto.nullable) = false];
@@ -187,10 +172,10 @@ message Response {
     ResponseApplySnapshotChunk  apply_snapshot_chunk  = 15;
     ResponsePrepareProposal     prepare_proposal      = 16;
     ResponseProcessProposal     process_proposal      = 17;
-    ResponseExtendVote          extend_vote           = 18;
-    ResponseVerifyVoteExtension verify_vote_extension = 19;
     ResponseFinalizeBlock       finalize_block        = 20;
   }
+  reserved 18; // Placeholder for ResponseExtendVote in v0.37
+  reserved 19; // Placeholder for ResponseVerifyVoteExtension in v0.37
 }
 
 // nondeterministic
@@ -323,7 +308,7 @@ message ResponsePrepareProposal {
   repeated ExecTxResult            tx_results              = 4;
   repeated ValidatorUpdate         validator_updates       = 5;
   tendermint.types.ConsensusParams consensus_param_updates = 6;
-  repeated bytes                   app_signed_updates      = 7;
+  reserved 7; // Placeholder for app_signed_updates in v0.37
 }
 
 message ResponseProcessProposal {
@@ -332,15 +317,6 @@ message ResponseProcessProposal {
   repeated ExecTxResult            tx_results              = 3;
   repeated ValidatorUpdate         validator_updates       = 4;
   tendermint.types.ConsensusParams consensus_param_updates = 5;
-}
-
-message ResponseExtendVote {
-  bytes app_signed        = 1;
-  bytes tendermint_signed = 2;
-}
-
-message ResponseVerifyVoteExtension {
-  bool accept = 1;
 }
 
 message ResponseFinalizeBlock {
@@ -436,8 +412,8 @@ message ValidatorUpdate {
 message VoteInfo {
   Validator validator                   = 1 [(gogoproto.nullable) = false];
   bool      signed_last_block           = 2;
-  bytes     tendermint_signed_extension = 3;
-  bytes     app_signed_extension        = 4;
+  reserved 3; // Placeholder for tendermint_signed_extension in v0.37
+  reserved 4; // Placeholder for app_signed_extension in v0.37
 }
 
 enum EvidenceType {
@@ -488,7 +464,5 @@ service ABCIApplication {
   rpc ApplySnapshotChunk(RequestApplySnapshotChunk) returns (ResponseApplySnapshotChunk);
   rpc PrepareProposal(RequestPrepareProposal) returns (ResponsePrepareProposal);
   rpc ProcessProposal(RequestProcessProposal) returns (ResponseProcessProposal);
-  rpc ExtendVote(RequestExtendVote) returns (ResponseExtendVote);
-  rpc VerifyVoteExtension(RequestVerifyVoteExtension) returns (ResponseVerifyVoteExtension);
   rpc FinalizeBlock(RequestFinalizeBlock) returns (ResponseFinalizeBlock);
 }
--- a/proto/tendermint/abci/types.proto
+++ b/proto/tendermint/abci/types.proto
@@ -21,20 +21,24 @@ import "gogoproto/gogo.proto";
 
 message Request {
   oneof value {
-    RequestEcho               echo                 = 1;
-    RequestFlush              flush                = 2;
-    RequestInfo               info                 = 3;
-    RequestInitChain          init_chain           = 4;
-    RequestQuery              query                = 5;
-    RequestBeginBlock         begin_block          = 6;
-    RequestCheckTx            check_tx             = 7;
-    RequestDeliverTx          deliver_tx           = 8;
-    RequestEndBlock           end_block            = 9;
-    RequestCommit             commit               = 10;
-    RequestListSnapshots      list_snapshots       = 11;
-    RequestOfferSnapshot      offer_snapshot       = 12;
-    RequestLoadSnapshotChunk  load_snapshot_chunk  = 13;
-    RequestApplySnapshotChunk apply_snapshot_chunk = 14;
+    RequestEcho                echo                  = 1;
+    RequestFlush               flush                 = 2;
+    RequestInfo                info                  = 3;
+    RequestInitChain           init_chain            = 4;
+    RequestQuery               query                 = 5;
+    RequestBeginBlock          begin_block           = 6 [deprecated = true];
+    RequestCheckTx             check_tx              = 7;
+    RequestDeliverTx           deliver_tx            = 8 [deprecated = true];
+    RequestEndBlock            end_block             = 9 [deprecated = true];
+    RequestCommit              commit                = 10;
+    RequestListSnapshots       list_snapshots        = 11;
+    RequestOfferSnapshot       offer_snapshot        = 12;
+    RequestLoadSnapshotChunk   load_snapshot_chunk   = 13;
+    RequestApplySnapshotChunk  apply_snapshot_chunk  = 14;
+    RequestPrepareProposal     prepare_proposal      = 15;
+    RequestExtendVote          extend_vote           = 16;
+    RequestVerifyVoteExtension verify_vote_extension = 17;
+    RequestFinalizeBlock       finalize_block        = 18;
   }
 }
 
@@ -117,26 +121,57 @@ message RequestApplySnapshotChunk {
   string sender = 3;
 }
 
+// Extends a vote with application-side injection
+message RequestExtendVote {
+  types.Vote vote = 1;
+}
+
+// Verify the vote extension
+message RequestVerifyVoteExtension {
+  types.Vote vote = 1;
+}
+
+message RequestPrepareProposal {
+  // block_data is an array of transactions that will be included in a block,
+  // sent to the app for possible modifications.
+  // applications can not exceed the size of the data passed to it.
+  repeated bytes block_data = 1;
+  // If an application decides to populate block_data with extra information, they can not exceed this value.
+  int64 block_data_size = 2;
+}
+
+message RequestFinalizeBlock {
+  repeated bytes          txs                  = 1;
+  bytes                   hash                 = 2;
+  tendermint.types.Header header               = 3 [(gogoproto.nullable) = false];
+  LastCommitInfo          last_commit_info     = 4 [(gogoproto.nullable) = false];
+  repeated Evidence       byzantine_validators = 5 [(gogoproto.nullable) = false];
+}
+
 //----------------------------------------
 // Response types
 
 message Response {
   oneof value {
-    ResponseException          exception            = 1;
-    ResponseEcho               echo                 = 2;
-    ResponseFlush              flush                = 3;
-    ResponseInfo               info                 = 4;
-    ResponseInitChain          init_chain           = 5;
-    ResponseQuery              query                = 6;
-    ResponseBeginBlock         begin_block          = 7;
-    ResponseCheckTx            check_tx             = 8;
-    ResponseDeliverTx          deliver_tx           = 9;
-    ResponseEndBlock           end_block            = 10;
-    ResponseCommit             commit               = 11;
-    ResponseListSnapshots      list_snapshots       = 12;
-    ResponseOfferSnapshot      offer_snapshot       = 13;
-    ResponseLoadSnapshotChunk  load_snapshot_chunk  = 14;
-    ResponseApplySnapshotChunk apply_snapshot_chunk = 15;
+    ResponseException           exception             = 1;
+    ResponseEcho                echo                  = 2;
+    ResponseFlush               flush                 = 3;
+    ResponseInfo                info                  = 4;
+    ResponseInitChain           init_chain            = 5;
+    ResponseQuery               query                 = 6;
+    ResponseBeginBlock          begin_block           = 7 [deprecated = true];
+    ResponseCheckTx             check_tx              = 8;
+    ResponseDeliverTx           deliver_tx            = 9 [deprecated = true];
+    ResponseEndBlock            end_block             = 10 [deprecated = true];
+    ResponseCommit              commit                = 11;
+    ResponseListSnapshots       list_snapshots        = 12;
+    ResponseOfferSnapshot       offer_snapshot        = 13;
+    ResponseLoadSnapshotChunk   load_snapshot_chunk   = 14;
+    ResponseApplySnapshotChunk  apply_snapshot_chunk  = 15;
+    ResponsePrepareProposal     prepare_proposal      = 16;
+    ResponseExtendVote          extend_vote           = 17;
+    ResponseVerifyVoteExtension verify_vote_extension = 18;
+    ResponseFinalizeBlock       finalize_block        = 19;
   }
 }
 
@@ -198,6 +233,7 @@ message ResponseCheckTx {
   int64          priority   = 10;
 
   // mempool_error is set by Tendermint.
+
   // ABCI applications creating a ResponseCheckTX should not set mempool_error.
   string mempool_error = 11;
 }
@@ -260,6 +296,32 @@ message ResponseApplySnapshotChunk {
     RETRY_SNAPSHOT  = 4;  // Retry snapshot (combine with refetch and reject)
     REJECT_SNAPSHOT = 5;  // Reject this snapshot, try others
   }
+}
+
+message ResponseExtendVote {
+  tendermint.types.VoteExtension vote_extension = 1;
+}
+
+message ResponseVerifyVoteExtension {
+  Result result = 1;
+
+  enum Result {
+    UNKNOWN = 0;  // Unknown result, treat as ACCEPT by default
+    ACCEPT  = 1;  // Vote extension verified, include the vote
+    SLASH   = 2;  // Vote extension verification aborted, continue but slash validator
+    REJECT  = 3;  // Vote extension invalidated
+  }
+}
+
+message ResponsePrepareProposal {
+  repeated bytes block_data = 1;
+}
+
+message ResponseFinalizeBlock {
+  repeated ResponseDeliverTx       txs                     = 1;
+  repeated ValidatorUpdate         validator_updates       = 2 [(gogoproto.nullable) = false];
+  tendermint.types.ConsensusParams consensus_param_updates = 3;
+  repeated Event                   events                  = 4 [(gogoproto.nullable) = false, (gogoproto.jsontag) = "events,omitempty"];
 }
 
 //----------------------------------------
@@ -355,15 +417,15 @@ service ABCIApplication {
   rpc Echo(RequestEcho) returns (ResponseEcho);
   rpc Flush(RequestFlush) returns (ResponseFlush);
   rpc Info(RequestInfo) returns (ResponseInfo);
-  rpc DeliverTx(RequestDeliverTx) returns (ResponseDeliverTx);
   rpc CheckTx(RequestCheckTx) returns (ResponseCheckTx);
   rpc Query(RequestQuery) returns (ResponseQuery);
   rpc Commit(RequestCommit) returns (ResponseCommit);
   rpc InitChain(RequestInitChain) returns (ResponseInitChain);
-  rpc BeginBlock(RequestBeginBlock) returns (ResponseBeginBlock);
-  rpc EndBlock(RequestEndBlock) returns (ResponseEndBlock);
   rpc ListSnapshots(RequestListSnapshots) returns (ResponseListSnapshots);
   rpc OfferSnapshot(RequestOfferSnapshot) returns (ResponseOfferSnapshot);
   rpc LoadSnapshotChunk(RequestLoadSnapshotChunk) returns (ResponseLoadSnapshotChunk);
   rpc ApplySnapshotChunk(RequestApplySnapshotChunk) returns (ResponseApplySnapshotChunk);
+  rpc ExtendVote(RequestExtendVote) returns (ResponseExtendVote);
+  rpc VerifyVoteExtension(RequestVerifyVoteExtension) returns (ResponseVerifyVoteExtension);
+  rpc PrepareProposal(RequestPrepareProposal) returns (ResponsePrepareProposal);
 }
--- a/proto/tendermint/abci/types.proto
+++ b/proto/tendermint/abci/types.proto
@@ -36,9 +36,10 @@ message Request {
     RequestLoadSnapshotChunk   load_snapshot_chunk   = 13;
     RequestApplySnapshotChunk  apply_snapshot_chunk  = 14;
     RequestPrepareProposal     prepare_proposal      = 15;
-    RequestExtendVote          extend_vote           = 16;
-    RequestVerifyVoteExtension verify_vote_extension = 17;
-    RequestFinalizeBlock       finalize_block        = 18;
+    RequestProcessProposal     process_proposal      = 16;
+    RequestExtendVote          extend_vote           = 17;
+    RequestVerifyVoteExtension verify_vote_extension = 18;
+    RequestFinalizeBlock       finalize_block        = 19;
   }
 }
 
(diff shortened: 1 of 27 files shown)


sergio-mena:
Pushing a further commit removing the parts related to vote extensions. When the time comes, it will be easy to bring them back by reverting this commit

proto/tendermint/abci/types.proto
```diff
@@ -21,21 +21,25 @@ import "gogoproto/gogo.proto";
...
+    RequestCommit              commit                = 10;
+    RequestListSnapshots       list_snapshots        = 11;
+    RequestOfferSnapshot       offer_snapshot        = 12;
+    RequestLoadSnapshotChunk   load_snapshot_chunk   = 13;
+    RequestApplySnapshotChunk  apply_snapshot_chunk  = 14;
+    RequestPrepareProposal     prepare_proposal      = 15;
+    RequestProcessProposal     process_proposal      = 16;
+    RequestFinalizeBlock       finalize_block        = 19;
   }
+  reserved 17, 18; // Future vote extensions
```
creachadair, line 42: Is there some reason to reserve these now rather than simply adding them to the end later? Mechanically it's fine, but I don't see a clear reason to insist on tag ordering.

proto/tendermint/abci/types.proto
```diff
@@ -262,6 +299,33 @@ message ResponseApplySnapshotChunk {
   }
 }
 
+message ResponsePrepareProposal {
+  bool                             modified_tx             = 1;
+  repeated TransactionRecord       txs                     = 2;
+  bytes                            app_hash                = 3;
+  repeated ExecTxResult            tx_resultsq             = 4;
+  repeated ValidatorUpdate         validator_updates       = 5 [(gogoproto.nullable) = false];
```
creachadair, line 307: Is the gogo annotation meant to imply this field cannot be empty? Or does it merely force the caller to set it empty explicitly instead of to null in the generated code?

proto/tendermint/abci/types.proto
```diff
@@ -285,6 +349,21 @@ message EventAttribute {
...
 
+// ExecTxResult contains results of executing one individual transaction.
+//
+// * Its structure is equivalent to #ResponseDeliverTx which will be deprecated/deleted
+message ExecTxResult {
+  uint32         code       = 1;
+  bytes          data       = 2;
+  string         log        = 3;  // nondeterministic
+  string         info       = 4;  // nondeterministic
+  int64          gas_wanted = 5 [json_name = "gas_wanted"];
```
creachadair, line 360: I think this is the default JSON name anyway, unless you explicitly request mixed case.

proto/tendermint/abci/types.proto
```diff
@@ -295,6 +374,21 @@ message TxResult {
   ResponseDeliverTx result = 4 [(gogoproto.nullable) = false];
 }
 
+// TxAction contains App-provided information on what to do with a transaction that is part of a raw proposal
+enum TxAction {
```
creachadair, line 378: Another option to consider would be to nest this enum inside `TransactionRecord`, and then the `ACTION_` prefix wouldn't be necessary. YMMV.

creachadair (commented):
Drive-by…

proto/tendermint/abci/types.proto
sergio-mena: I structured the two commits in this PR like this:

* The initial commit contains the whole ABCI++ interface (as it appears in the spec)
* The second commit removes what we are likely to _not_ ship in v0.36 (i.e., vote extension related APIs)

The advantage I see in structuring the commits like this is that, by "git revert"ing the second commit, we can easily get vote extensions back whenever we need them (likely, post v0.36).

As a result, the "reserved" was my way of "removing" vote extensions in the second commit, but of course I'm open to alternative ways of doing this.

proto/tendermint/abci/types.proto
sergio-mena: Yep, I'll remove the JSON annotation here. I will also go through the whole PR again to see if this comment also applies elsewhere

proto/tendermint/abci/types.proto
cmwaters: I think it's not necessary to include the `reserved` tags. Maybe we can just add a comment linking to the vote extensions planned for v0.37

proto/tendermint/abci/types.proto
sergio-mena: I had this question when writing this, but I read [here](https://earthly.dev/blog/backward-and-forward-compatibility/) that "grouping messages" is not good for forward compatibility.

After reading your comment, I went back to that blog and I realize I had misunderstood, confusing "nesting" and "grouping".

So, I guess that, if nesting is safe w.r.t. compatibility, it's probably cleaner to do as you suggest and move the enum inside the message, and remove the "ACTION_" prefixes.

Do you agree with this?

sergio-mena:
> Drive-by…

Thanks for the review!

proto/tendermint/abci/types.proto
sergio-mena: OK, I'll turn them into pure comments

creachadair: I see, that makes sense. If we are confident the design will not need to change in such a way that affects these proto messages, I think it's fine to leave reservations, but I agree with @cmwaters that a comment is worthwhile.

(That said: I would _also_ keep the `reserved` notation)

```proto
   reserved 17;  // placeholder for BlahBlah1 in v0.37.
   reserved 18;  // placeholder for BlahBlah2 in v0.37.
```

proto/tendermint/abci/types.proto
creachadair: I don't object to including the annotation, it just stood out because it was the only one among others that could, at least in theory, have it.

I would prefer we try to avoid annotations we don't need, since they increase the semantic burden of reading the file. I also generally feel that nullability is not something that should be specified in the proto file—that's a language-specific concern, not an encoding one.

Having said that, I realize we already have a bunch of these elsewhere, and I think it affects the way gogo generates code (😢) so if we need to do it for consistency that's fine.

proto/tendermint/abci/types.proto
creachadair: Yeah, nesting is fine. It's not crucial, the main thing is that it avoids stutter in the generated code. The `ACTION_UNKNOWN` enumerator at top-level would look like `abcipb.TxAction_ACTION_UNKNOWN` in Go or `abci::TxAction::ACTION_UNKNOWN` in C++.

Whether or not we nest it, it would be nice to drop the `ACTION_` prefixes. And if the enum is _only_ used within `TransactionRecord`, that would also be a helpful hint to the reader. YMMV.

proto/tendermint/abci/types.proto
sergio-mena: @cmwaters what's your take here.
Quite honestly, I tend to agree with @creachadair that "repeated Blah" should not have the "nullable = false" tag. I'm inclined to:
* Remove it from `ResponsePrepareProposal`,  `ResponseProcessProposal`, and `ResponseFinalizeBlock`
* Leave it in `ResponseEndBlock` ... anyway we'll nuke that one soon enough
Let me know if you have concerns with this approach.

spec/abci++/abci++_methods_002_draft.md
```diff
@@ -303,24 +304,26 @@ From the App's perspective, they'll probably skip ProcessProposal
...
+    * The first five parameters of `RequestPrepareProposal` are the same as `RequestProcessProposal`
       and `RequestFinalizeBlock`.
     * The header contains the height, timestamp, and more - it exactly matches the
       Tendermint block header.
     * The Application can modify the transactions received in `RequestPrepareProposal` before sending
       them in `ResponsePrepareProposal`. In that case, `ResponsePrepareProposal.modified_tx` is set to true.
     * If `ResponsePrepareProposal.modified_tx` is false, then Tendermint will ignore the contents of
-      `ResponsePrepareProposal.tx`.
+      `ResponsePrepareProposal.txs`.
+    * If the Application modifies the transactions, the modified block MUST NOT exceed the configured maximum size,
```
williambanfield, line 323: Can we give more specific guidance for what 'modified block' means here? Is it that `sizeof(txs)` in bytes cannot be greater than the `MaxBytes`? Or that all of the fields that the application returns cannot be greater than that? I think that passing the `max_bytes` along to the application makes sense, but I'm not yet sure it's obvious what the application should do with the `MaxBytes` if it is going to modify the block.

spec/abci++/abci++_methods_002_draft.md
```diff
@@ -530,12 +536,13 @@ In the cases when _p_'s Tendermint is to broadcast `precommit nil` messages (eit
...
-    | height    | int64 | Height of the block  (for sanity check).                                                 | 4            |
+    | Name              | Type  | Description                                                                              | Field Number |
+    |-------------------|-------|------------------------------------------------------------------------------------------|--------------|
+    | app_signed        | bytes | Optional information signed by the Application (not by Tendermint).                      | 1            |
+    | tendermint_signed | bytes | Optional information signed by Tendermint.                                               | 2            |
+    | hash              | bytes | The header hash of the propsed block that the vote extension refers to.                  | 3            |
+    | validator_address | bytes | [Address](../core/data_structures.md#address) of the validator that signed the extension | 4            |
+    | height            | int64 | Height of the block  (for sanity check).                                                 | 5            |
```
williambanfield, line 545: Perhaps a different PR, but should we remove this as well given that we're not planning to implement it this release cycle?

spec/abci++/abci++_methods_002_draft.md
sergio-mena: Very good point William. The complication here is that, once we introduce vote extensions (v0.37), we may also give the App the chance to modify *app_signed* data.

After a bit of thinking, this is what I'm gonna do:

* I'll rename this field to *max_tx_bytes*
* We will add another field in v0.37: *max_extension_bytes*

Tendermint will calculate what those fields should be just before calling `RequestPrepareProposal`.

Let me know if you have concerns with this approach.

spec/abci++/abci++_methods_002_draft.md
sergio-mena: I'd prefer to keep the spec complete, so everyone can see what we are planning to do and comment on that.

Then, when the v0.36 (throttle) branch is created, we can remove the vote extension information from that version of the spec.

Adding and issue so we don't forget

sergio-mena (commented):
Thanks William for taking a look

proto/tendermint/abci/types.proto
```diff
@@ -117,27 +121,59 @@ message RequestApplySnapshotChunk {
...
+  LastCommitInfo          last_commit_info     = 4 [(gogoproto.nullable) = false];
+  repeated Evidence       byzantine_validators = 5 [(gogoproto.nullable) = false];
+}
+
+message RequestFinalizeBlock {
+  bytes                   hash                 = 1;
+  tendermint.types.Header header               = 2 [(gogoproto.nullable) = false];
+  repeated bytes          txs                  = 3;
+  LastCommitInfo          last_commit_info     = 4 [(gogoproto.nullable) = false];
+  repeated Evidence       byzantine_validators = 5 [(gogoproto.nullable) = false];
```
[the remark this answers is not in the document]
cmwaters: The `Evidence` struct here isn't actually the `Evidence` struct in `types` but just the name of a validator, their voting power and what they did wrong: either `DuplicateVote` or `LightClientAttack` atm

proto/tendermint/abci/types.proto
```diff
@@ -262,6 +301,34 @@ message ResponseApplySnapshotChunk {
...
 }
 
+message ResponsePrepareProposal {
+  bool                             modified_tx             = 1;
+  repeated TransactionRecord       txs                     = 2;
+  bytes                            app_hash                = 3;
+  repeated ExecTxResult            tx_results              = 4;
+  repeated ValidatorUpdate         validator_updates       = 5;
+  tendermint.types.ConsensusParams consensus_param_updates = 6;
+  reserved 7; // Placeholder for app_signed_updates in v0.37
```
cmwaters, line 311: I don't think it makes sense to add this because:
a) we don't have a similar reserved field for giving the app the data in `RequestPrepareProposal`
b) I don't think we're 100% sure yet that we need to have `app_signed_updates` returned (and in a block)

proto/tendermint/abci/types.proto
```diff
@@ -295,6 +377,21 @@ message TxResult {
   ResponseDeliverTx result = 4 [(gogoproto.nullable) = false];
 }
 
+message TransactionRecord {
+  TxAction action           = 1;
+  bytes tx                  = 2;
+  repeated bytes new_hashes = 3;
```
cmwaters, line 383: Perhaps not here but somewhere in the spec we need to be explicit that the hashing algorithm matches the one that tendermint is using else this doesn't work

sergio-mena: I'll add a bullet with this info in the spec.

proto/tendermint/abci/types.proto
sergio-mena: Regarding a), the data is coming in `LastCommitInfo` (see changes to VoteInfo below in lines 415-416)
As for b), when can always "remove" a reserved field later if it turns out not to be necessary/wanted.

Having said this, I'm open to removing this line completely. But then we'd need to be consistent and also remove lines 415-416.

cmwaters: I guess we can leave it for now - I just don't think we're that sure of it nor that there's a strong need to have future fields in the proto files

cmwaters (approved):

proto/tendermint/abci/types.proto
```diff
@@ -117,27 +122,60 @@ message RequestApplySnapshotChunk {
   string sender = 3;
 }
 
+message RequestPrepareProposal {
+  bytes                   hash                 = 1;
+  tendermint.types.Header header               = 2 [(gogoproto.nullable) = false];
+  // txs is an array of transactions that will be included in a block,
+  // sent to the app for possible modifications.
+  repeated bytes          txs                  = 3;
```
creachadair, line 130: This is a meta-question, not a change request for this PR:

I'm concerned about the size of the data in a repeated transaction field. Do we have any language in the spec around how the implementation is permitted to handle corner cases, report errors, etc.? 

I expect the intent is that not all the txn for a block would be prepared at the same time, similarly to how we CheckTx one-by-one today. Given the tendency of app developers to rely on any corner case not otherwise defined, however, I wonder if we can't make some provision in the spec (e.g., "the implementation is permitted to limit the number/size of txn in this field to <some reasonable rule> provided <minimum standard>" or words to that effect.

Relatedly (and also not necessarily for this PR): We might want to pre-emptively pull this out into a message, e.g.,

```proto
message TransactionBatch {
   repeated bytes txs = 1;
}
```

That makes it easier for us to inject a layer of indirection if we later need to handle really big sets, without breaking the wire format of the enclosing message.

proto/tendermint/abci/types.proto
```diff
@@ -117,27 +122,60 @@ message RequestApplySnapshotChunk {
...
+  LastCommitInfo          last_commit_info     = 4 [(gogoproto.nullable) = false];
+  repeated Evidence       byzantine_validators = 5 [(gogoproto.nullable) = false];
+  // the modified transactions cannot exceed this size.
+  int64                   max_tx_bytes         = 6;
+}
+
+message RequestProcessProposal {
+  bytes                   hash                 = 1;
+  tendermint.types.Header header               = 2 [(gogoproto.nullable) = false];
```
creachadair, line 139: (further thoughts, non-blocking)

Doesn't `tendermint.types` also depend on `tendermint.abci`? I realize the code generator hides the cyclic dependency from the build tool, but I feel like we should try to make the dependencies go in only one direction if we can.

(The one case I know about specifically is `Event`, and maybe the easy solution there would be to move that over to `tendermint.types` and reverse the dependency)

proto/tendermint/abci/types.proto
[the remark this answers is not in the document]
creachadair: As an aside it really is the same type: `types.Evidence` in TM is not a struct, it's an interface, whose implementations wrap the aforementioned duplicate-vote and light-client evidence structures. It gets unpacked from a message shaped like this, though we don't use the generated type directly.

proto/tendermint/abci/types.proto
```diff
@@ -190,14 +228,15 @@ message ResponseCheckTx {
   bytes          data       = 2;
   string         log        = 3;  // nondeterministic
   string         info       = 4;  // nondeterministic
-  int64          gas_wanted = 5 [json_name = "gas_wanted"];
-  int64          gas_used   = 6 [json_name = "gas_used"];
+  int64          gas_wanted = 5;
```
creachadair, line 231: Is gas configuration likely to grow/change in the future? Might we want a message type here (e.g., `GasOptions` or something)? 

I mention this because I see we duplicate these fields in a couple of places, and when there are multiple related bits that need to move together I think "message".

proto/tendermint/abci/types.proto
creachadair: I think it's fine to leave it, but it's also not really necessary since it's not as if there is a gap in the numbering that we need to protect from being filled in. YMMV.

The main value of it IMO is the documentation. 🙂

creachadair (approved):
I made a few more general comments, none of which are blocking.

proto/tendermint/abci/types.proto
sergio-mena: OK. Let's leave it for the moment then

proto/tendermint/abci/types.proto
sergio-mena: The short answer is that *max_tx_bytes* below is the limit (in bytes) of the whole `txs` field. But I understand your concern as, currently, ABCI never sends/receives transactions in a batch... so this will be something we'll need to test extensively.

As you may have already figured out, I have a limited experience with protocol buffers, so if you think that wrapping a potentially huge parameter in its own message offers an extra layer of safety/security/robustness/peace of mind... then we can do definitely do that...

... and if we do it, the moment is now :-)

proto/tendermint/abci/types.proto
sergio-mena: That's an excellent point. I hadn't realized about the dependency cycle. Even if the build tool is able to cope with it, I agree it's not (conceptually) clean.

The way I see it, `tendermint.abci` should depend on `tendermint.types`, and not the contrary, as `tendermint.types` seems to be defining the most basic types. But I might have got it wrong...

I'll open an issue on this, but won't include it in the ABCI++ project. I did a quick search on the file and there are preexisting references to (e.g. RequestInitChain, RequestBeginBlock)

proto/tendermint/abci/types.proto
sergio-mena: AFAICT, the discussions I've been part of in the context of ABCI++ the idea was to keep the concept of *gas* unmodified, albeit, providing parameters in the new APIs if that makes sense.

I would personally keep it as it is now because:

a) as I said above *gas* is not likely to change
b) as `ResponseCheckTx` defined those parameters like this long ago, App developers are used to this

sergio-mena (commented):
Thanks @creachadair !

proto/tendermint/abci/types.proto
```diff
@@ -262,6 +299,33 @@ message ResponseApplySnapshotChunk {
...
+message ResponseProcessProposal {
+  bool                             accept                  = 1;
+  bytes                            app_hash                = 2;
+  repeated ExecTxResult            tx_results              = 3;
+  repeated ValidatorUpdate         validator_updates       = 4 [(gogoproto.nullable) = false];
+  tendermint.types.ConsensusParams consensus_param_updates = 5;
+}
+
+message ResponseFinalizeBlock {
+  repeated Event                   block_events            = 1
```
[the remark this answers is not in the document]
williambanfield: :facepalm:

proto/tendermint/abci/types.proto
creachadair: Adding a message doesn't really mitigate the problem at all—it's more a question of future-proofing. 

If, for example, we wanted to include additional information in the request like "there are more transactions coming" or "I truncated this message" or whatever, that's easier to do if there's a message. 

That said: I'm not advocating this as a best practice, just thinking about how we'd adapt if we run into a situation where there is "too much" tx data. Do we drop the request on the floor? Do we truncate it silently? If so, which data do we drop? Does the application have any way to discover we did this?  Those kinds of questions.

I don't have a strong opinion here.

sergio-mena: I see your point. I think it's hard to say at this point. Let's leave it as it is for now, but let's also keep this in mind as we progress with the implementation.

proto/tendermint/abci/types.proto
creachadair: > The way I see it, `tendermint.abci` should depend on `tendermint.types`, and not the contrary, as `tendermint.types` seems to be defining the most basic types.

I tend to agree. That said: I think it's mainly a matter of understandability rather than correctness.

If we don't want to do it as part of ABCI++ (which is a very reasonable choice), it might not be worth bothering. Once the message type tags get buried in databases and APIs, it'll be hard to change them without breaking stuff, and I'd probably not ship a breaking change just to fix it.

proto/tendermint/abci/types.proto
```diff
@@ -262,6 +301,34 @@ message ResponseApplySnapshotChunk {
   }
 }
 
+message ResponsePrepareProposal {
+  bool                             modified_tx             = 1;
+  repeated TransactionRecord       txs                     = 2;
```
williambanfield, line 306: nit: we use Tx in most places to mean Transaction, I think it's worth keeping that pattern up here and also naming the field `tx_records`.

sergio-mena: Totally agree! Changed. Please double check

proto/tendermint/abci/types.proto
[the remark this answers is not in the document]
thanethomson: I have to say that something about just calling this field `evidence` is appealing to me. Probably due to the fact that we're listing specific evidence of malfeasance (e.g. the height at which the malfeasance took place), and we're not just listing byzantine validators' information.

If we were just capturing byzantine validators' info, then the field name would make more sense to me. But then, of course, it wouldn't be evidence :)

Also applies to the similarly named field in `RequestProcessProposal`.

proto/tendermint/abci/types.proto
thanethomson: I'm torn here because I really like the idea of an encapsulated, easily extensible `gas_options` field. It's not related to ABCI++ though, so perhaps we should consider breaking this out into a separate issue that we can address at another time (perhaps when someone asks for an additional gas-related field?).

proto/tendermint/abci/types.proto
```diff
@@ -285,6 +352,21 @@ message EventAttribute {
...
+  string         log        = 3;  // nondeterministic
+  string         info       = 4;  // nondeterministic
+  int64          gas_wanted = 5;
+  int64          gas_used   = 6;
+  repeated Event tx_events  = 7
+      [(gogoproto.nullable) = false, (gogoproto.jsontag) = "events,omitempty"];  // nondeterministic
+  string          codespace = 8;
+}
+
 // TxResult contains results of executing the transaction.
```
thanethomson, line 370: Is this `TxResult` message used anywhere any more? Its `result` field is of type `ResponseDeliverTx`, which is deprecated right?

proto/tendermint/abci/types.proto
```diff
@@ -21,21 +21,26 @@ import "gogoproto/gogo.proto";
...
-    RequestListSnapshots      list_snapshots       = 11;
-    RequestOfferSnapshot      offer_snapshot       = 12;
-    RequestLoadSnapshotChunk  load_snapshot_chunk  = 13;
-    RequestApplySnapshotChunk apply_snapshot_chunk = 14;
+    RequestEcho                echo                  = 1;
+    RequestFlush               flush                 = 2;
+    RequestInfo                info                  = 3;
+    RequestInitChain           init_chain            = 4;
+    RequestQuery               query                 = 5;
+    RequestBeginBlock          begin_block           = 6 [deprecated = true];
```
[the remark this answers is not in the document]
creachadair: I don't know what the plan is, but there are a couple of considerations:

1. During the transition, keeping the fields allows us to make the software see that they're set and report an appropriate diagnostic ("your software is out of date!" rather than "you omitted this required input!")
2. If any of these messages land in block data, we _might_ want to keep those around longer, so that we could handle soft migrations.
3. But if not (2), then we should probably plan to eventually drop them entirely, perhaps in the following release.

proto/tendermint/abci/types.proto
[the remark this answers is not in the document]
sergio-mena: @thanethomson if you have a strong opinion on this we can change it. The main reason it's not clear to me that we should is that the same parameter with the same name has been around in ABCI for years (in `RequestBeginBlock`)... but since these changes are breaking compatibility anyway I guess renaming it is also a good option.

Let me know what option you prefer

proto/tendermint/abci/types.proto
sergio-mena: Thanks @thanethomson  for looking into this. May I suggest you to create an issue with this dilemma? If you do create it, you might consider to include it in the ABCI++ board (toward the end) and then when ABCI++ implementation is move advanced we can pick it up and see when/how we address this

proto/tendermint/abci/types.proto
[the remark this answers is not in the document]
sergio-mena: I think Michael's option 3 is what we will end up doing, as those are not used in the block structure as such. But for the time being, until we change de logic that uses then to use the new APIs, let's leave them like that so that we don't break the build.
Regarding 17 and 18, if you take a look at the third commit of this PR in isolation (f9b42ecc) you'll see the point for those.

proto/tendermint/abci/types.proto
sergio-mena: Yeah! Well spotted :-). This message is not used elsewhere in the ABCI protos (neither ABCI not ABCI++), but it is used in some places in the code.

My impression is that we will need to through the whole codebase and try to remove all references to `TxResult`; then we can remove/deprecate the `TxResult` message here. But that was too much work for this PR. Hence I named the new message `ExecTxResult` to avoid conflicts.

If we manage to get rid of `TxResult` before v0.36, we might consider renaming `ExecTxResult` to `TxResult`. But this is not crucial.

proto/tendermint/abci/types.proto
cmwaters: My idea was that if the application sent too much evidence we would log an error and the proposer wouldn't propose anything for that round. Given that Tendermint is the client and the application the server, it's hard for Tendermint to actually report errors through ABCI. I can't recall it actually happening anywhere. 

The other option which would be a lot more aggressive is that Tendermint panics when it gets back a tx set which exceeds the max bytes provided. We do a similar thing in the case of validator updates. If the application returns a validator set voting power that is greater than maxInt64 / 8, Tendermint will panic.

proto/tendermint/abci/types.proto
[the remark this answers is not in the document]
cmwaters: What I was meaning in my above comment was that the term `Evidence` for me should contain some data that proves that a validator did something wrong (like `types.DuplicateVoteEvidence` does). `abci.Evidence` doesn't show any proof, rather it's telling that app this is what happened i.e. it's a record of misbehavior (in fact if you look at the SDK they call this `Misbehavior`. So perhaps we could rename the struct to `abci.Misbehavior` and call the field `misbehavior`

proto/tendermint/abci/types.proto
cmwaters: My personal view on this is that `gas` is one of the unique cases were the abstraction between application and consensus engine became blurred. "Gas" is an application defined concept which tries to encapsulate the collective cost to the machines running that network to execute and store a transaction. It's part of a systems economics not part of state machine replication. I think this was initially added because people submitting a tx have to go through Tendermint to the application and back, however if a tx doesn't have enough "gas" this doesn't need to be it's own fields rather it should use `codespace`, `data` and `info`. As we are modifying the methods I think it's as good a time as any to take a closer look at the structs these methods return

proto/tendermint/abci/types.proto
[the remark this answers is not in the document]
sergio-mena: OK, if the SDK is calling it misbehavior, then I think it's a good idea to align with that. I'll rename it today. Please voice any concerns you may have with renaming to "misbehavior"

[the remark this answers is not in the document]
sergio-mena: Actually, I started the renaming but I see that (preexisting) `RequestBeginBlock` is also affected, so I would like to defer any action on this until we have completed the `FinalizeBlock` logic in the code

proto/tendermint/abci/types.proto
sergio-mena: Quite frankly, I would go for the second option. We are providing the App with `max_tx_bytes` in this message, so if the App returns a transaction set that is bigger than that, it's an obvious bug in the App. I don't think it's wise to try to carry on with such a bug in the App. But let's decide what to do when implementing the logic inside Tendermint.

proto/tendermint/abci/types.proto
sergio-mena: I'd propose to open an issue whose goal is deciding what to do here, and implement the decision in a PR. Then, we can add the issue to the board, and prioritize it.

spec/abci++/abci++_methods_002_draft.md
williambanfield: I think that this seems very reasonable.

PR #395, opened 2022-02-01
Updated Apalache type annotations

Kukovec:
The specifications were using the old format of Apalache type annotations. This PR migrates them to the new format (but does not extend the specifications otherwise).


konnov (approved):
Looks great! Thanks a lot!

PR #398, opened 2022-02-02
PBT draft 002: Apalache annotations

Kukovec:
First version of draft 002, with Apalache type annotations. See #395.


josef-widder (approved):
lgtm

PR #400, opened 2022-02-02
Add the newly defined timeout parameters to the consensus parameters

williambanfield:
This pull request adds the `TimeoutParams` that were defined in [ADR-74](https://github.com/tendermint/tendermint/blob/master/docs/architecture/adr-074-timeout-params.md) to the consensus parameters.

The main value of having these parameters in the consensus parameters is to enable the application to more reliably manage their values. The enforcement of their values across all validators is much less important than their equality and configurability by applications.

Closes: https://github.com/tendermint/tendermint/issues/7759

### DIFF
--- a/proto/tendermint/types/params.proto
+++ b/proto/tendermint/types/params.proto
@@ -130,7 +130,7 @@ message TimeoutParams {
   // enable_commit_timeout_bypass configures the node to proceed immediately to
   // the next height once the node has received all precommits for a block, forgoing
   // the remaining commit timeout.
-  // Setting enable_commit_timeout_bypass = false configures Tendermint to wait
+  // Setting enable_commit_timeout_bypass false (the default) causes Tendermint to wait
   // for the full commit.
   bool enable_commit_timeout_bypass = 6;
 }
--- a/proto/tendermint/types/params.proto
+++ b/proto/tendermint/types/params.proto
@@ -86,12 +86,11 @@ message SynchronyParams {
 // TimeoutParams configure the timeouts for the steps of the Tendermint consensus algorithm.
 message TimeoutParams {
 
-  // propose along with propose_delta configure the timeout for the propose step of
-  // the Tendermint consensus algorithm.
-  //
-  // propose configures the initial timeout and propose_delta configures by how much the timeout
-  // grows each subsequent round. For the first round, this propose timeout is used
-  // and for every subsequent round, the timeout grows by propose_delta.
+  // These fields configure the timeouts for the propose step of the Tendermint
+  // consensus algorithm: propose is the initial timeout and propose_delta
+  // determines how much the timeout grows in subsequent rounds.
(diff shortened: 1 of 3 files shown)


proto/tendermint/types/params.proto
```diff
@@ -81,3 +82,55 @@ message SynchronyParams {
...
+
+// TimeoutParams configure the timeouts for the steps of the Tendermint consensus algorithm.
+message TimeoutParams {
+
+  // propose along with propose_delta configure the timeout for the propose step of
+  // the Tendermint consensus algorithm.
+  //
+  // propose configures the initial timeout and propose_delta configures by how much the timeout
+  // grows each subsequent round. For the first round, this propose timeout is used
+  // and for every subsequent round, the timeout grows by propose_delta.
```
creachadair, line 94: Optional suggestions mainly around reducing repetition:

```suggestion
  // These fields configure the timeouts for the propose step of the Tendermint
  // consensus algorithm: propose is the initial timeout and propose_delta
  // determines how much the timeout grows in subsequent rounds.
  // For the first round, this propose timeout is used and for every subsequent
  // round, the timeout grows by propose_delta.
```

proto/tendermint/types/params.proto
```diff
@@ -81,3 +82,55 @@ message SynchronyParams {
...
+  // These parameters influence the vote step timeouts in the the same way that
+  // the propose and propose_delta parameters do to the proposal step.
+  //
+  // The vote timeout does not begin until a quorum of votes has been received. Once
+  // a quorum of votes has been seen and this timeout elapses, Tendermint will
+  // procced to the next step of the consensus algorithm. If Tendermint receives
+  // all of the remaining votes before the end of the timeout, it will proceed
+  // to the next step immediately.
+  google.protobuf.Duration vote = 3
+      [(gogoproto.stdduration) = true];
```
creachadair, line 120: I imagine these would fit on the same line OK.

proto/tendermint/types/params.proto
```diff
@@ -81,3 +82,55 @@ message SynchronyParams {
...
+  // commit configures how long Tendermint will wait after receiving a quorum of
+  // precommits before beginning consensus for the next height. This can be
+  // used to allow slow precommits to arrive for inclusion in the next height before progressing.
+  google.protobuf.Duration commit = 5
+      [(gogoproto.stdduration) = true];
+
+  // enable_commit_timeout_bypass configures the node to proceed immediately to
+  // the next height once the node has received all precommits for a block, forgoing
+  // the remaining commit timeout.
+  // Setting enable_commit_timeout_bypass = false configures Tendermint to wait
```
creachadair, line 133:
```suggestion
  // Setting enable_commit_timeout_bypass false (the default) causes Tendermint to wait
```

creachadair (approved):

PR #402, opened 2022-02-04
make consensus parameters non nullable

williambanfield:
Allow these parameters to be pointers in the Go code doesn't actually buy us much: these parameters must be set and will be set on application startup, so allowing them to be nil only leaves a possible panic without much gain.


creachadair (commented):
I do not like delegating questions about field validity to the code generator. Whether or not the field is represented as a pointer is orthogonal to its presence in the payload: The node has to check either way. 

The more of these gogo annotations we rely on, the harder it will be to swap out to the standard code generator later. The auto-generated helpers can avoid the need for a lot of explicit nil-checks regardless how we annotate the field.

So—I think we shouldn't do this, but given that we've already done this a bunch elsewhere, it wouldn't be the end of the world either.

williambanfield:
I'm find forgoing them, I don't think it's that big of a risk and am certainly somewhat guilty of cargo-culting this into place.

PR #404, opened 2022-02-22
add archive notice

cmwaters:
This PR adds an archive notice to the main README.md file. 

I'm unsure if we want to add a similar redirection notice to every file in the spec﻿


creachadair (approved):

cmwaters:
I haven't added the notice to all files but most of the main ones

